PR #44, opened 2014-12-04
Changes for Vehicle Data classes

### LINKED ISSUE #12: Vehicle Data classes are not scalable
There are currently 25 supported vehicle data types with demand to add even more in the future.

However, current vehicle data RPC classes contain a distinct setter/getter method for each individual vehicle data type - which means each RPC related to vehicle data needs to define 50 separate methods.  As we add more vehicle data types, 2 additional methods (a setter & getter) will need to be added to each RPC.

A better way would be to define all vehicle data types in an enum and provide a generic hash map that maps the vehicle data enum to it's associated object (depends on the specific RPC in question).  In this design, adding a new type of vehicle data only requires that it be added to the enum.

mikeburke106:
This pull request fixes #12.  See discussion in the thread for that issue.

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/proxy/RPCRequestFactory.java
+++ b/sdl_android_lib/src/com/smartdevicelink/proxy/RPCRequestFactory.java
@@ -764,6 +764,7 @@ public class RPCRequestFactory {
 		return msg;
 	}	
 	
+	@Deprecated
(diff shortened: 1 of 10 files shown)


justinjdickow:
Before this feature makes complete sense we need to implement interface to execute callbacks based on correlation ID of request (so the developer does not have to figure out which vehicle data they received)

joeygrover:
Closing as stale.

Issue #10, opened 2014-10-09
RPC messages should sanitize user inputs


mikeburke106:
RPC messages should sanitize user inputs before storing them in the underlying hashtable.

For example, in the AddCommand RPC message, the documentation claims that CmdID can only be between 0 and 2,000,000,000.  However, the code doesn't check that the user input is within this range.

The RPC code should check the user input and throw an InvalidParameterException if it does not meet the specified criteria.

PR #50, opened 2015-01-06
[WIP] Updated Java Docs for SDL

HansenKirk:


mikeburke106:
Kirk,

Could you please set something up like [Joel's example pull request](https://github.com/smartdevicelink/SmartDeviceLink-iOS/pull/45)?  Particularly, the TODO section, which just shows the progress of your pull request and a checklist of things you are working on that you check off as you complete.

From a high level, I would like to discuss the point of documenting the min/max values in the comments if we aren't validating these values within the code and if the min/max values could potentially be different per OEM or SDL adopter.  We've had a similar discussion in #10, where we were told the mobile libraries don't validate user inputs against the requirements in order to keep the libs lightweight and because the values may be different per OEM.  If that is the case, why are we documenting min/max values in the comments?  I think we need to **both** document and validate min/max values or **neither** document nor validate min/max values.  Doing one but not the other doesn't make much sense IMO.

I also think it would be useful to have a "required" or "optional" tag on all the setter methods so the developer can easily determine which variables they MUST set and which variables they can set in addition to the minimum mandatory variables.

Hopefully you're still working on this, but here are some issues I found upon review:

  - AddCommand
    - [MenuParams type is ButtonName](https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AddCommand.java#L58-59)
  - AddCommandResponse
    - [Missing constructor documentation](https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AddCommandResponse.java#L15-21)
  - AddSubMenu
    - ["SubMenu" misspelled "SunMenu" and missing a space](https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AddSubMenu.java#L11)
    - [Strange characters need to be removed](https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AddSubMenu.java#L40)
    - ["greater than or equal to"](https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AddSubMenu.java#L130)
  - AddSubMenuResponse
    - [Missing constructor documentation](https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AddSubMenuResponse.java#L20-22)
  - Airbag Status
    - [Referenced classes don't exist] (https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AirbagStatus.java#L81-82)
    - [Missing constructor documentation] (https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AirbagStatus.java#L96-107)
    - [Missing documentation on all setter/getter methods](https://github.com/HansenKirk/sdl_android/blob/a4aec6b667377adae01e850da383547d9cbefc07/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/AirbagStatus.java#L109-284)


It doesn't look like this PR is ready for review yet, so I'm removing the "scheduled" tag.  Once it's ready, please let me know and we'll add it back to the agenda.

mikeburke106:
@HansenKirk  we are starting to follow the [GitFlow Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow), which requires feature branches to be merged into the develop branch of the main project.  This PR is attempting to be merged into master, which is supposed to only contain releases.  Here's a good picture explaining the merge process.



Please re-create this PR and request to merge into develop instead of master.

Issue #48, opened 2014-12-18
PutFile doesn't handle file data correctly


mikeburke106:
PutFile has a setter/getter method to set/get FileData, which [takes in a byte array and stores it in the hashtable](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/PutFile.java#L138-147).

```java
    public void setFileData(byte[] fileData) {
        if (fileData != null) {
            parameters.put(RPCStruct.KEY_BULK_DATA, fileData);
        } else {
        	parameters.remove(RPCStruct.KEY_BULK_DATA);
        }
    }
    public byte[] getFileData() {
        return (byte[]) parameters.get(RPCStruct.KEY_BULK_DATA);
    }
```

This is incorrect behavior, as binary data is sent through the binary protocol, not through the JSON protocol.  When I send a PutFile using this method to set the file data, the request is rejected saying "Binary data empty".  When I send a PutFile using setBulkData() instead of setFileData(), it works correctly.

Recommend removing these two functions entirely or calling the bulk data methods directly from these methods.

PR #54, opened 2015-01-22
Fix for how PutFile handles bulk data

mikeburke106:
This was PR #49, but I messed up the branches on my fork, so I'm re-submitting this change.

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/PutFile.java
+++ b/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/PutFile.java
@@ -136,14 +136,10 @@ public class PutFile extends RPCRequest {
         return (Boolean) parameters.get(KEY_PERSISTENT_FILE);
     }
     public void setFileData(byte[] fileData) {
-        if (fileData != null) {
-            parameters.put(RPCStruct.KEY_BULK_DATA, fileData);
-        } else {
-        	parameters.remove(RPCStruct.KEY_BULK_DATA);
-        }
+        setBulkData(fileData);
     }
     public byte[] getFileData() {
-        return (byte[]) parameters.get(RPCStruct.KEY_BULK_DATA);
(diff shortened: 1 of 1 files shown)


mikeburke106:
Also related to issue #48.

A main reason that this issue was hard to track down is related to an over-arching issue, which I've pointed out before, but wanted to bring it up again.  When I investigated this issue, I was logging out the JSON messages and was seeing something like this as part of the JSON:

```json
"bulkData": "15901589d"
```

Even after I found that the PutFile was putting the bulk data into the JSON structure, I was confused why I was getting this seemingly random number associated with it.  Since the binary data is an array of bytes, I would've expected to see a JSON array of numbers here, not this single random number.  Then, I remembered the issue I pointed out in [this presentation](https://docs.google.com/a/livioradio.com/presentation/d/1BvbVJRm6wxPUFmGJsZMMQMgL7fK7CaeAzpgudepCT44/edit) on slide 15, relating to the JsonRPCMarshaller class:

> If object in hashtable is something other than RPCStruct, Vector or Hashtable, the result of the object’s toString() method will be stored in the JSON string, which is neither desired nor intuitive

This is a perfect example of why the "catch-all" else statement [in the serializeHashtable method](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/marshal/JsonRPCMarshaller.java#L116-118) is bad.  Since a byte[] object is not specifically checked for as part of the if/else block, it is falling through and simply putting the object in the JSONObject itself.  Therefore, the JsonRPCMarshaller is putting the memory address returned from the toString() method into the JSON, which is what this seemingly random number represents.

I understand that this "should never happen," but we've just seen it happen and it caused a bug that was hard to track down.  I'm going to be working on implementing my RPC re-design to prevent bugs like this, so once I'm done, I hope we can get it reviewed and merged into the project.

PR #68, opened 2015-01-22
[WIP] Feature/Multiplexing

joeygrover:
:smirk_cat:  Needs discussion on next steps :smirk_cat:

## Summary
This PR will involve several pieces to allow Android apps to share a single transport connection. It is important to note that this implementation currently uses the C4 library. For production we should decide what we want to do.

See <a href="https://docs.google.com/presentation/d/1FwLmlInSNgqSgsUiEjdVXZpojtHXhvL4ZRsir8Bbq9E/edit?usp=sharing" >this quick presentation </a> for more details.

1. Introduce single point router service
2. Allow apps to register with router service for transport connection
3. Refactor low level class into more intuitive and parcelable SdlPacket 

### Single point transport
The communication to hardware will take place via a single point service. This router service will build packets based on bytes received, find their sessionID to AppID mapping and send to the correct app address. This will happen through a few mappings.

### Router to App communication
In this instance the communication is based on intents and broadcast receivers. It was discussed that this should possibly be replaced by using normal binding via aidl files. After looking more into the argument, I found that this intent method is still the best. 

### Sdl Packet
Introduced a new class that houses the frame header info as well as the generic payload. This was important because there was too much passing around of byte arrays unattached to protocol frame headers. This simplifies a lot of that communication.

##### General
- [ ] Optimize lower level logic to be less redundant (due to new Sdl Packet class)
- [ ] Either include C4.jar or move classes from project into the SDL lib
- [ ] Clean up logs

##### Single point transport
- [x] Move transport logic into a standalone service decoupled from proxy object
- [x] Create adapter into proxy that allows it to work alongside existing transports
- [x] Make sure newest router service always starts up when possible

##### Router to App communication
- [x] Create registration procedure through custom intent system
- [x] Enable apps to unregister from service
- [x] Set up mapping between app id's and session Id's
- [ ] Possibly switch from intent/broadcast receiver to aidl method

##### Sdl Packet and lower level
- [x] Create new class to handle frame headers and payloads in one class
- [x] Make said class parcelable to be able to send from router service to client without redundant packet parsing logic on both ends
- [x] Replace ProtocolFrameHeader and factory with SdlPacket and factory.


mikeburke106:
I know you're already aware of this, Joey, but other contributors should be aware that features on this project (not on a fork) should be on a branch named as `feature/branch_name` as per the contributing guidelines.  So, instead of `multiplexing`, this branch should be named `feature/multiplexing`.  If you do the work on a fork, you can name your branch whatever you want.

PR #72, opened 2015-01-23
Changes for throwing BLUETOOTH_SOCKET_UNAVAILABLE exception when No SDP ...

### LINKED ISSUE #61: Improve exception handling when bluetooth rfcomm socket is unavailable
If the device does not have an bluetooth rfcomm socket available, a generic exception is thrown instead of the more specific IOException.

anildahiya:
Fixed #61 with commit 654a3f70e1b918ce711673583b28de45c05598ef

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/transport/BTTransport.java
+++ b/sdl_android_lib/src/com/smartdevicelink/transport/BTTransport.java
@@ -172,7 +172,7 @@ public class BTTransport extends SdlTransport {
 			}			
 		} catch (IOException e) {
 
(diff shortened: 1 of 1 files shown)


mikeburke106:
Wow, [this try/catch logic is really messy and complicated](https://github.com/anildahiya/sdl_android/blob/BlueTooth_Socket_Exception/sdl_android_lib/src/com/smartdevicelink/transport/BTTransport.java#L135-207).  It should probably be broken up into multiple functions to be more readable.

"SYNC" should be changed to "SDL" before we can merge this.

anildahiya:
"SYNC" changed to "SDL"

Issue #56, opened 2015-01-22
getAudioPassThruCapabilities getter missing from proxyALM object


mrapitis:
The alm proxy object contains getter methods for the data structures that are returned from a register app interface response, however the proxy object is missing a getter method for AudioPassThruCapabilities.  A getter method needs to be added to return this information to the app developer.

PR #73, opened 2015-01-23
Changes for Implementation of getAudioPassThruCapabilities getter in pro...

mrapitis:
Regarding issue #56 
Commit 3d6e7efee77c572b524b7341c08e769a458a265c


mikeburke106:
Only comment on this one is that [the additions in SdlProxyBase](https://github.com/smartdevicelink/sdl_android/pull/73/files#diff-1) violate the [DRY principle](http://programmer.97things.oreilly.com/wiki/index.php/Don't_Repeat_Yourself).  The common code should be moved into a function so we don't have to modify the code in 2 different places.

Issue #58, opened 2015-01-22
PutFile streaming methods need enhancements to provide progress and completion status


mrapitis:
The putfile streaming logic needs to be enhanced to include call backs for both progress updates and provide a completion status of the entire putfile stream.

joeygrover:
Merged.

PR #74, opened 2015-01-23
Update putfile stream enhancements

mrapitis:
:boom: READY FOR REVIEW :boom: 

Related to issue #58 

<h2>Summary</h2>
The PutFile RPC allows a developer to send a PutFile request using an offset and length.  Sending a series of PutFile requests allows a developer to split a large file into smaller blocks.  While having the ability to split large files into chunks is a great feature supported by SDL, the bookkeeping and management of offsets and lengths for each individual file is a cumbersome process that should be abstracted from app developers.  Having a mobile API that creates and sends a stream of PutFile requests reduces a large burden on app developers while working with the PutFile RPC.

<h2>Objective:</h2>
Provide a simple mobile public facing API that allows a developer to send a file via SDL to the vehicle head unit utilizing a stream of PutFile requests.
  
Existing PutFile (request & response) Definition:

```xml
 <function name="PutFile" functionID="PutFileID" messagetype="request">
    <description>
    	Used to push a binary data onto the SYNC module from a mobile device, such as icons and album art
    	Not supported on first generation SYNC vehicles.
    	Binary data is in binary part of hybrid msg.
    </description>

    <param name="syncFileName" type="String" maxlength="255" mandatory="true">
      <description>File reference name.</description>
    </param>

    <param name="fileType" type="FileType" mandatory="true">
      <description>Selected file type.</description>
    </param>

    <param name="persistentFile" type="Boolean" defvalue="false" mandatory="false">
      <description>
      	Indicates if the file is meant to persist between sessions / ignition cycles.
      	If set to TRUE, then the system will aim to persist this file through session / cycles.
      	While files with this designation will have priority over others, they are subject to deletion by the system at any time.
      	In the event of automatic deletion by the system, the app will receive a rejection and have to resend the file.
      	If omitted, the value will be set to false.
      </description>
    </param>

    <param name="systemFile" type="Boolean" defvalue="false" mandatory="false" >
      <description>
      	Indicates if the file is meant to be passed thru core to elsewhere on the system.
      	If set to TRUE, then the system will instead pass the data thru as it arrives to a predetermined area outside of core.
      	If omitted, the value will be set to false.
      </description>
    </param>

    <param name="offset" type="Float" minvalue="0" maxvalue="100000000000" mandatory="false">
      <description>Optional offset in bytes for resuming partial data chunks</description>
    </param>
    <param name="length" type="Float" minvalue="0" maxvalue="100000000000" mandatory="false">
      <description>
      	Optional length in bytes for resuming partial data chunks
      	If offset is set to 0, then length is the total length of the file to be downloaded
      </description>
    </param>  

  </function>

<function name="PutFile" functionID="PutFileID" messagetype="response">
    <description>Response is sent, when the file data was copied (success case). Or when an error occured.</description>
    <description>Not supported on First generation SYNC vehicles. </description>
  	<param name="success" type="Boolean" platform="documentation">
		<description> true, if successful; false, if failed </description>
	</param>
	
    <param name="resultCode" type="Result" platform="documentation">
      <description>See Result</description>
      <element name="SUCCESS"/>
      <element name="INVALID_DATA"/>
      <element name="OUT_OF_MEMORY"/>
      <element name="TOO_MANY_PENDING_REQUESTS"/>
      <element name="APPLICATION_NOT_REGISTERED"/>
      <element name="GENERIC_ERROR"/>      
      <element name="REJECTED"/>
      <element name="UNSUPPORTED_REQUEST"/>                     
    </param>
    
    <param name="spaceAvailable" type="Integer" minvalue="0" maxvalue="2000000000">
          <description>
          	Provides the total local space available in SDL Core for the registered app.
          	If the transfer has systemFile enabled, then the value will be set to 0 automatically.
          </description>
    </param>
    
    <param name="info" type="String" maxlength="1000" mandatory="false" platform="documentation">
      <description>Provides additional human readable info regarding the result.</description>
    </param>
  </function>
```
<h2>Goals:</h2>
- [x] Create a public API that allows developers to provide a path to a particular file on the devices file system, as well as other optional and mandatory putfile parameters (see above).  An automatically calculated stream of putfile requests will then be created utilizing the provided file contents.

Public facing API used to initiate a putfile stream request:
```java
/**
	 * Used to push a stream of putfile RPC's containing binary data from a mobile device to the module.
	 * Responses are captured through callback on IProxyListener.
	 *
	 * @param sPath - The physical file path on the mobile device.
	 * @param sdlFileName - The file reference name used by the putFile RPC.
	 * @param iOffset - The data offset in bytes, a value of zero is used to indicate data starting from the beginging of a file.
	 * A value greater than zero is used for resuming partial data chunks.
	 * @param fileType - The selected file type -- see the FileType enumeration for details
	 * @param bPersistentFile - Indicates if the file is meant to persist between sessions / ignition cycles.
	 * @param  bSystemFile - Indicates if the file is meant to be passed thru core to elsewhere on the system.
	 * @param correlationID - A unique ID that correlates each RPCRequest and RPCResponse.
	 * @return boolean - True if the putfile stream was started successfully, false if an exception occurred during stream creation. 
	 * @throws SdlException
	*/	
	public boolean PutFileStream(String sPath, String sdlFileName, Integer iOffset, FileType fileType, Boolean bPersistentFile, Boolean bSystemFile, Integer iCorrelationID) throws SdlException 

```
- [x] Notify the developer if the PutFileStream was started successfully, on failure log an error to the devices logcat. (accomplished using the api call displayed above available within sdlproxybase.)

- [x] After the stream has started, notify the developer the status of the putfile stream utilizing the OnStreamRPC notification (new).  The notification will include the file name, the bytes completed and the total file size.

- [x] After the stream has finished, notify the developer of the status of the completion (success or failure) using the StreamRPCResponse (new) with a matching correlation id from the initial putfile stream request.

Future enhancements:
- [x] Provide handlers in the new proxy framework to send notifications for each event (start of stream, notification for progress, overall result of stream) via listener interfaces. (To be covered in feature/Framework branch)

PutFileStream Flow
```
App requests a new PutFileStream
proxy.PutFileStream(params)----------->
App receives a status of stream creation
<-----------------PutFileStream success
App receives OnStreamRPC Notifications indicating the current transfer progress
<-----------------OnStreamRPC notification
<-----------------OnStreamRPC notification
<-----------------OnStreamRPC notification
App receives StreamRPCResponse indicating the overall result of the transfer request.
<-----------------StreamRPCResponse response
```

<h3>Other TODO items:</h3>

- [X] The PutFile RPC needs to be updated to accept Longs for setLength and setOffsetValues as we might run into overflows since a Long can store a wider range than an Integer.  A Long is returned for the size of the file from the Android OS, however we have to truncate to an Integer to accommodate the existing PutFile definition.  To be covered in #150

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
+++ b/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
@@ -1,7 +1,7 @@
 package com.smartdevicelink.Dispatcher;
 
-public interface IDispatchingStrategy<messageType> {
-	public void dispatch(messageType message);
+public interface IDispatchingStrategy<T> {
+	public void dispatch(T message);
 	
 	public void handleDispatchingError(String info, Exception ex);
 	
--- a/sdl_android_lib/src/com/smartdevicelink/Dispatcher/ProxyMessageDispatcher.java
+++ b/sdl_android_lib/src/com/smartdevicelink/Dispatcher/ProxyMessageDispatcher.java
@@ -5,17 +5,17 @@ import java.util.concurrent.PriorityBlockingQueue;
 
 import com.smartdevicelink.util.DebugTool;
 
-public class ProxyMessageDispatcher<messageType> {
-	PriorityBlockingQueue<messageType> _queue = null;
+public class ProxyMessageDispatcher<T> {
+	PriorityBlockingQueue<T> _queue = null;
 	private Thread _messageDispatchingThread = null;
-	IDispatchingStrategy<messageType> _strategy = null;
+	IDispatchingStrategy<T> _strategy = null;
 
 	// Boolean to track if disposed
 	private Boolean dispatcherDisposed = false;
 	
-	public ProxyMessageDispatcher(String THREAD_NAME, Comparator<messageType> messageComparator, 
-			IDispatchingStrategy<messageType> strategy) {
-		_queue = new PriorityBlockingQueue<messageType>(10, messageComparator);
+	public ProxyMessageDispatcher(String THREAD_NAME, Comparator<T> messageComparator, 
+			IDispatchingStrategy<T> strategy) {
+		_queue = new PriorityBlockingQueue<T>(10, messageComparator);
 		
 		_strategy = strategy;
 		
@@ -38,7 +38,7 @@ public class ProxyMessageDispatcher<messageType> {
 	private void handleMessages() {
 		
 		try {
-			messageType thisMessage;
+			T thisMessage;
 		
 			while(dispatcherDisposed == false) {
 				thisMessage = _queue.take();
@@ -53,7 +53,7 @@ public class ProxyMessageDispatcher<messageType> {
 		}
 	}
 		
-	public void queueMessage(messageType message) {
+	public void queueMessage(T message) {
 		try {
 			_queue.put(message);
 		} catch(ClassCastException e) { 
--- a/sdl_android_lib/src/com/smartdevicelink/SdlConnection/ISdlConnectionListener.java
+++ b/sdl_android_lib/src/com/smartdevicelink/SdlConnection/ISdlConnectionListener.java
@@ -26,7 +26,4 @@ public interface ISdlConnectionListener {
 	
 	public void onHeartbeatTimedOut(byte sessionID);
 	
-	public void onOnStreamRPC(IPutFileResponseListener putFileListener, OnStreamRPC rpcNote);
-	
-	public void onStreamRPCResponse(IPutFileResponseListener putFileListener, StreamRPCResponse result);	
 }
--- a/sdl_android_lib/src/com/smartdevicelink/SdlConnection/ISdlConnectionListener.java
+++ b/sdl_android_lib/src/com/smartdevicelink/SdlConnection/ISdlConnectionListener.java
@@ -2,9 +2,7 @@ package com.smartdevicelink.SdlConnection;
 
 import com.smartdevicelink.protocol.ProtocolMessage;
 import com.smartdevicelink.protocol.enums.SessionType;
-import com.smartdevicelink.proxy.interfaces.IPutFileResponseListener;
-import com.smartdevicelink.proxy.rpc.OnStreamRPC;
-import com.smartdevicelink.proxy.rpc.StreamRPCResponse;
+
 
 public interface ISdlConnectionListener {
 	public void onTransportDisconnected(String info);
--- a/sdl_android_lib/src/com/smartdevicelink/SdlConnection/ISdlConnectionListener.java
+++ b/sdl_android_lib/src/com/smartdevicelink/SdlConnection/ISdlConnectionListener.java
@@ -2,6 +2,9 @@ package com.smartdevicelink.SdlConnection;
 
 import com.smartdevicelink.protocol.ProtocolMessage;
 import com.smartdevicelink.protocol.enums.SessionType;
+import com.smartdevicelink.proxy.interfaces.IPutFileResponseListener;
+import com.smartdevicelink.proxy.rpc.OnStreamRPC;
+import com.smartdevicelink.proxy.rpc.StreamRPCResponse;
 
 public interface ISdlConnectionListener {
 	public void onTransportDisconnected(String info);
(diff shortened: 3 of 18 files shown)


mikeburke106:
We can talk about this in the meeting, but I need some time to review further.

mikeburke106:
Is there some way we can test this code?  Is this supported in the SyncProxyTester application?

Could you prepare a design review document that we can discuss on the next call?  This PR is making a lot of changes on pretty much every layer of the project, so I would like to get a better understanding of how everything works.  I'm particularly curious about why so much connection code had to change when this seems like a feature that could be handled as part of the SdlProxyBase class.

Here are my comments after review:

  * Missing documentation for new APIs (is this being worked on as part of the JavaDoc PR?)
  * Do the streaming RPC methods belong in the ISdlConnectionListener interface or should they be added to a new interface?  Streaming RPCs doesn't have anything to do with connection, so I'm not sure this is the appropriate place to define these methods.
  * JavaDoc for [FileInputStream.available()](http://developer.android.com/reference/java/io/FileInputStream.html#available()) states `Note that this method provides such a weak guarantee that it is not very useful in practice.`  Therefore, relying on this value to determine size of RPC "chunks" ([I think that's what it's doing???](https://github.com/mrapitis/sdl_android/blob/update_putfile_stream_enhancements/sdl_android_lib/src/com/smartdevicelink/SdlConnection/SdlConnection.java#L294)) is probably not very reliable.
  * I think you mentioned on the last call that new "RPCs" were introduced, but they aren't actually received or sent to the head-unit?  So, the notifications and responses are just faked in this case?  If so, what's the point of defining them as "RPCs"?  I have the same question in regards to OnLockScreenStatus notification, which also does not come from the head-unit and is faked in the mobile libraries.  I am concerned about sending fake data through the proxy as if it came from the head-unit, as it's very deceiving to developers.

mikeburke106:
* [AbstractPacketizer buffer size](https://github.com/mrapitis/sdl_android/blob/update_putfile_stream_enhancements/sdl_android_lib/src/com/smartdevicelink/streaming/AbstractPacketizer.java#L17) increased from 1,488 bytes to 150,000 bytes.  What is the reason for such a drastic increase to the buffer size?  Since this is an abstract class, will there be side-effects to other subclasses (other than requiring 100x more memory for the buffer)?
  * [Commented out code](https://github.com/mrapitis/sdl_android/blob/update_putfile_stream_enhancements/sdl_android_lib/src/com/smartdevicelink/streaming/StreamRPCPacketizer.java#L23) should be deleted from the codebase prior to merge.

mrapitis:
Yes, the functionality is available within the SyncProxyTester application for testing.  To test with the app: select the putFile RPC and select a file from the sd card on the device, next check the transfer using stream checkbox and click ok.

I will prepare a design document, however I'm not sure if it will be ready for the next call as I was out of the office all of last week and I am currently in the process of catching up.

I agree the JavaDocs need to be updated with the new API.
Some of my intent was to avoid placing a dependency on SdlProxyBase as the streaming packetizer classes work independently of the base proxy, however I will investigate how this could be simplified and cleaned up more.
Currently the FileInputStream.available() method is  used to calculate the size of the file, the call has been reliable but since the javadocs mention the weak guarantee I will replace the call.

We probably should discuss other details and concerns further during our next call, especially in regards to the increased buffer size.

joeygrover:
Hey Markos,

I'm in attempts to look over this pull request. I might be late to the party on this, but why don't we use something like a PutFile UpdateListener interface rather than introducing a new internal RPC/"global" callback? 

So in the PutFile.java we just add an interface such as
```` Java
    public interface UpdateListener{
    	public void onStarted();
    	public void onProgressUpdate();
    	public void onFinished();
    }
````

Then each putfile request would have their own listener attached and preventing the developers from having to keep track of each request. I do realize this would require a good deal of overhaul.

mrapitis:
I like the suggestion, it may take some overhaul... I'll investigate and keep you posted about feasibility.

mrapitis:
Hey Joey,

I looked further into your idea this morning, but I'm not sure that it would make sense for the problem that putfile streaming is targeting.  In the case of putfile streaming, the app never actually creates or builds any putfile requests (the proxy creates all of these requests).  The app simply kicks off the process by requesting a putfile stream with provided parameters (location of file in memory, offset,  filetype, etc) and waits for notifications on progress and a final response.

Thanks

joeygrover:
Alright, I thought about this some more and I think it can still be done pretty well. I'm going to work through my thoughts and commit something to get your thoughts on it.

Issue #59, opened 2015-01-22
Broadcast logging can cause an exception if the apps service becomes unavailable


mrapitis:
If the apps service becomes unexpectedly unavailable.  In the code below, if the service becomes unavailable the getApplicationContext call will fail causing an exception.

```java
private void sendBroadcastIntent(Intent sendIntent)
	{
		Service myService = null;		
		if (_proxyListener != null && _proxyListener instanceof Service)
		{
			myService = (Service) _proxyListener;				
		}
		else if (_appService != null)
		{
			myService = _appService;
		}
		else
		{
			return;
		}
		Context myContext = myService.getApplicationContext();
		if (myContext != null) myContext.sendBroadcast(sendIntent);		
	}
```

mikeburke106:
I think this is going to be a higher-level "logging" discussion, possibly related to #51.  Sending logs via broadcast intent doesn't really make much sense to me, especially from a background service.

PR #76, opened 2015-01-23
Catching an exception in sendbroadcast if service unexpectedly becomes u...

mrapitis:
Regarding issue #59 
Commit 6b005f22a83f0c72cb5864a15a0426888879bcf7


mikeburke106:
This solution doesn't address the underlying design issue.  In general, exceptions happen for a reason, so ignoring them is usually not a good idea.

Issue #57, opened 2015-01-22
Typedef exceptions can occur when utilizing vehicle data getter methods that return Doubles


mrapitis:
Currently this is an issue for getVehicleData and SubscribeVehicleData rpc's.  
When receiving float based values from sdl core that do not include a decimal (e.g. 80) internal objects are stored as integers causing casting exceptions to occur when values are returned via rpc getter methods (e.g. getSpeed).

mikeburke106:
I don't think this is an issue with the Android libraries, this is an issue with versions of SDL core not meeting the specification, but yeah, the Android libs need to support this incorrect behavior from SDL core.  An issue should be filed for SDL core as well, unless this has been fixed core-side already.

I think this needs to be maintained as part of a list of "hacks" required to support backwards compatibility issues.

PR #77, opened 2015-01-23
Changes for checking type on Vehicle Data

mrapitis:
Regarding issue #57 
Commit 5e386668060d54dc907a664eaa9c63c82abc8b0f

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java
+++ b/sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java
@@ -2,8 +2,11 @@ package com.smartdevicelink.util;
 
 public class SdlDataTypeConverter {
 	
-	public static Double ObjectToDouble(Object obj) {
-    	Double D = null;
+	public static Double objectToDouble(Object obj) {
+    	if (obj == null) {
+    		return null;
+    	}
+		Double D = null;		
     	if (obj instanceof Integer) {
     		int i = ((Integer) obj).intValue();
     		double d = (double) i;
--- a/sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java
+++ b/sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java
@@ -1,0 +1,18 @@
+package com.smartdevicelink.util;
+
+public class SdlDataTypeConverter {
+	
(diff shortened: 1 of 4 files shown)


mikeburke106:
Violates DRY principle and [new method](https://github.com/smartdevicelink/sdl_android/pull/77/files#diff-3) doesn't follow Java naming conventions.

sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java
```diff
@@ -0,0 +1,18 @@
+package com.smartdevicelink.util;
+
+public class SdlDataTypeConverter {
+	
+	public static Double ObjectToDouble(Object obj) {
```
justinjdickow: method should be called objectToDouble

sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java
```diff
@@ -0,0 +1,18 @@
...
+    	Double D = null;
+    	if (obj instanceof Integer) {
```
justinjdickow: if obj is null just return null immediately before all of the reflection

sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java
mikeburke106: Also, for completeness (I realize this isn't relevant to the case we're addressing), this method might want to allow conversions from floats and possibly strings as well.  If a float like `3.14159f` is passed in, the method will return null.  Similarly, if a string like `"3.14159"` is passed in, the method will return null.  Technically, both of these values could be converted to a double.

If someone wants to use this method sometime in the future for some other purpose, it might have undesired behaviors in these cases.

mikeburke106:
What if a float or a string value representing an integer is passed into [this function](https://github.com/mrapitis/sdl_android/blob/68a0c1964f4e5f1a1cc946e0c5fbae43869bea0a/sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java#L5-19)?  I realize that's not the use-case we're looking at, but if someone else wants to re-use this method in the future, it may have unexpected results (returning null) instead of returning an integer representation of that object.

justinjdickow:
Need to make a note in the javadocs that the conversion function will only convert integer or double

PR #78, opened 2015-01-23
[WIP]Feature/multiplexing

joeygrover:
:smirk_cat:  Fixed the new issue :smirk_cat:

## Summary
This PR will involve several pieces to allow Android apps to share a single transport connection. It is important to note that this implementation currently uses the C4 library. For production we should decide what we want to do.

See <a href="https://docs.google.com/presentation/d/1FwLmlInSNgqSgsUiEjdVXZpojtHXhvL4ZRsir8Bbq9E/edit?usp=sharing" >this quick presentation </a> for more details.

1. Introduce single point router service
2. Allow apps to register with router service for transport connection
3. Refactor low level class into more intuitive and parcelable SdlPacket 

### Single point transport
The communication to hardware will take place via a single point service. This router service will build packets based on bytes received, find their sessionID to AppID mapping and send to the correct app address. This will happen through a few mappings.

### Router to App communication
In this instance the communication is based on intents and broadcast receivers. It was discussed that this should possibly be replaced by using normal binding via aidl files. After looking more into the argument, I found that this intent method is still the best. 

### Sdl Packet
Introduced a new class that houses the frame header info as well as the generic payload. This was important because there was too much passing around of byte arrays unattached to protocol frame headers. This simplifies a lot of that communication.

### Make work with old gen of Applink 1. 
After testing with the older TDK, the multiplexing solution doesn't work for it. It only allows only one session per RCOMM channel. Joey is working on how to handle this case. The next possibility will to have the bluetooth class open up 'n' accept sockets where 'n' is equal to how many apps are waiting for sessions. This would require a user to hit the 'Find new apps" twice, however, per earlier discussions this was acceptable. 

##### General
- [x] Optimize lower level logic to be less redundant (due to new Sdl Packet class)
- [x] Either include C4.jar or move classes from project into the SDL lib
- [x] Clean up logs
- [x] Make work with old gen of Applink 1. 
- [x] Create documentation on what app dev's will need to do to incorporate this and how old gen is handled. 

##### Single point transport
- [x] Move transport logic into a standalone service decoupled from proxy object
- [x] Create adapter into proxy that allows it to work alongside existing transports
- [x] Make sure newest router service always starts up when possible

##### Router to App communication
- [x] Create registration procedure through custom intent system
- [x] Enable apps to unregister from service
- [x] Set up mapping between app id's and session Id's
- [x] Switch from intent/broadcast receiver to binding method
- [x] Create BluetoothTransportConfig as just an extension of Multiplexing config

##### Sdl Packet and lower level
- [x] Create new class to handle frame headers and payloads in one class
- [x] Make said class parcelable to be able to send from router service to client without redundant packet parsing logic on both ends
- [x] Replace ProtocolFrameHeader and factory with SdlPacket and factory.
- [x] Implement the PSM into USB and WiFi transports.

## Pros/Cons
During this process it has come to light that certain quirks will be in place because of this change. They are as followed:

#### Pros
- Theoretically unlimited apps connected to head unit 
- Once connected, newly installed apps will just appear without the need for user interaction.
- Newest transport layer fixes will be applied via the newest router service check
- If an app crashes, the router service still remains up and the crashed app can gracefully recover

#### Cons
- If the app that is running the router service process is forced closed through the Android settings, the router service will be shutdown and all apps will be shown as disconnected. It will propagate itself again, however the user will need to press "Find new apps.." again.
- If an app is upgraded while connected to the head unit, it will cause the router service to close and all apps will be shown as disconnected. 
- There is a buffer limit for IPC, usually around 1MB. Could be lower for older/lower cost phones.


joeygrover:
Just a note about cleaning up some of the code in previous classes. There are a lot of unused variables that are in the protected space so they don't flag a warning. Might write an issue, however, I am only finding this out by looking at the classes that affect this pull request.

joeygrover:
New issue added to support old head units.

joeygrover:
Creating a wiki page for developers https://github.com/smartdevicelink/sdl_android/wiki/All-About-Multiplexing

PR #82, opened 2015-01-27
Return null instead of throwing an exception when reverse-lookup fails in DisplayType

### LINKED ISSUE #79: DisplayType valueForString throws a runtime exception
[DisplayType enum throws an exception](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/DisplayType.java#L28) when valueForString is called with a value that isn't defined in the enum.  Other similar enums simply return null, which I think is the appropriate action.

```java
public static DisplayType valueForString(String value) {
    for (DisplayType type : DisplayType.values()) {
        if (type.toString().equals(value)) {
            return type;
        }
    }

    throw new IllegalArgumentException("Unknown value " + value);
}
```

mikeburke106:
Fixes #79.  Return null instead of throwing an exception when reverse-lookup fails

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/DisplayType.java
+++ b/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/DisplayType.java
@@ -25,7 +25,7 @@ public enum DisplayType {
             }
         }
 
-        throw new IllegalArgumentException("Unknown value " + value);
+        return null;
     }
 
     @Override


mikeburke106:
Marking for 3-day offline review. Anticipated merge date is 1/31/2015 at 3pm.

As a hotfix, this should also be merged into the develop branch.

mikeburke106:
Actuallly, this one might be backwards.  According to Java documentation for [Enum.valueOf](http://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#valueOf(java.lang.Class,%20java.lang.String)), if the passed-in string is not associated with any enum value, it throws an IllegalArgumentException.

Therefore, any enums that use the `valueOf` method will also throw an IllegalArgumentException if the string value doesn't exist in the enums.  However, the enum classes that don't use the `valueOf` method typically return null if the string value doesn't exist.  This leads to inconsistent behavior across enum classes.

This issue needs some more discovery and will not be a hotfix since more enums (presumably) will change.  Therefore, I'm rejecting and closing this PR.

PR #83, opened 2015-01-27
Modified layout capabilities when display layout is successfully changed.

### LINKED ISSUE #53: SdlProxyBase doesn't update DisplayCapabilities on SetDisplayLayoutResponse
Mike: SdlProxyBase keeps a global variable containing the display capabilities for the head-unit.  This global variable is only set on register app interface - it isn't updated when the layout changes.  When the layout changes, size of images, number of text fields, etc may change, so the global variable should be updated to reflect these changes.

mikeburke106:
Fixes #53.  Modified layout capabilities when display layout is successfully changed.


mikeburke106:
Need to investigate any side-effects of making this change.  SdlProxyBase is a large, complicated beast and all of these mutable variables are returned directly (in SdlProxyALM), meaning they could be modified in any part of the system.  Assuming no other parts of the system are messing with these variables, this PR should be good to go.

mikeburke106:
Feature requests should be merged into develop, not master.  Will create a new PR for this.

Issue #47, opened 2014-12-17
Support for Lock Screen Icon


mikeburke106:
See issue on SDL core [here](https://github.com/smartdevicelink/sdl_core/issues/3).

Head-unit wants to add support for showing a different image on the app's lock screen based on what type of radio is connected.  This requires adding a new request type to OnSystemRequest called ```LOCK_SCREEN_ICON_URL```.  The head-unit will send OnSystemRequest with request type ```LOCK_SCREEN_ICON_URL``` and URL contained in the URL field after it receives the RegisterAppInterface command from the app.  The app is responsible for handling this command in onOnSystemRequest callback, downloading the image from the URL and displaying the image on the lock screen.

Only thing that needs to change in the Android library is to add the new request type.

PR #95, opened 2015-01-30
Implementation of lock screen icon via OnSystemRequest

mikeburke106:
Added new RequestType enum value for lock screen icon.  See #47 for more info.

Developers can now choose to download the icon at any time by calling `getLockScreenIcon` in SdlProxyBase.  This will either download the image from the OnSystemRequest URL or return the image directly if it has already been downloaded.


ojmakke:
I see that LOCK_SCREEN_ICON_URL, "ends with comma" is followed directly by a ";"

mikeburke106:
This is intentional and is [valid Java syntax](http://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9).

The idea with this formatting is that if we need to add a new item into the enum, we don't have to modify the line containing the semicolon, we can simply add a new item with the comma at the end.  It turns a 2-3 line change into a 1 line change when the 2 files are compared, making the diff easier to read and compare.

Let's say we want to add an item to the following enum:

```java
public enum Number{
    ONE,
    TWO;
}
```

It would look something like this:

```java
public enum Number{
    ONE,
    TWO, // <-- had to delete the semicolon here and replace it with a comma
    THREE; // <-- added this line
}
```

Whereas, the way I formatted the semicolon would look like this:
```java
public enum Number{
    ONE,
    TWO,
    ;
}
```

And adding the new enum value looks like this:
```java
public enum Number{
   ONE,
   TWO,
   THREE, // <-- added this line
   ;
}
```

joeygrover:
@justinjdickow need to confirm this works with latest SDL_CORE version

justinjdickow:
Status: We confirmed that the system request is received by the application in the current version of core with this pull request.

Next Steps: is the developer facing portion of this pull request good to go in the proxy?

PR #113, opened 2015-02-16
Release 3.3.0

mikeburke106:
Merging the changes from last week into master branch as release 3.3.0.  All changes have already been reviewed, but due to API changes, this will be considered a "minor" version under semantic versioning, hence the pull request for review as well.

Here are the issues that are addressed in this version:

  * Update HU capabilities after successful setDisplayLayoutResponse (#53 via #112)
  * Refactor fuel level state methods to follow Java naming conventions (#84 via #96)
  * Made VrCapabilities follow project guidelines (#89 via #92)
  * Fix enable/disable of SiphonServer (#60 via #75)

Issue #89, opened 2015-01-29
VrCapabilities valueForString doesn't follow project design


mikeburke106:
[VrCapabilities](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/VrCapabilities.java) valueForString method doesn't follow the same design as the rest of the enums in the project.

Most enums look like this:

```java
public static PrerecordedSpeech valueForString(String value) {
    return valueOf(value);
}
```

But, this class looks like this:

```java
public static VrCapabilities valueForString(String value) {
    if (value.toUpperCase().equals(VrCapabilities.Text.toString().toUpperCase()))
    {
    	return VrCapabilities.Text;
    }
    return null;
}
```

Not sure why we're ignoring case on this one, I don't believe that is correct.  Likewise, if another value is added to this enum, additional work would need to be done to add any new values to this method, which is bad practice.  

Recommend refactoring this method to follow the same design as other enums.

mrapitis:
This change was initially implemented for legacy compatibility.  Looks like the original request suggested using the toUpperCase method that is in question.  Details can be found here:
https://bitbucket.org/fmcalt/al_android_dev/issue/29/protect-vrcapabilities-enum-value-for

mikeburke106:
Ok, this needs to be added to list of known hacks and corner cases.

Ideally, this should use [`equalsIgnoreCase`](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#equalsIgnoreCase(java.lang.String)) instead of calling `toUpper` on both strings.  There is still the issue that if any other enum values are added to this enum, additional logic will have to be implemented as well because this method will return null if the new enum string is passed into this method.  So, instead of just returning null, we should probably return `valueOf(value)`.

mikeburke106:
Not everyone can see this private bitbucket page, so here's the information:

`Stefan Bankowski created an issue 2013-08-13`
>The XML specification had been updated to make all the enumerated values consistent. In doing so, the vrCapabilities (currently only one) value of "Text" is now the proper "TEXT".
>
>The proxy needs to handle both cases and treat them the same. The suggested method is a "toUpper()" on the enumerated plain-text value.

I also notice that values of `TExt`, `TeXt`, etc will be accepted using this implementation.  Not sure if that's acceptable or not, but just wanted to make a note of it.

mikeburke106:
[...]

Issue #71, opened 2015-01-23
SdlChoice constructor isn't visible


mikeburke106:
In [OnSdlChoiceChosen](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSdlChoiceChosen.java), the constructor for SdlChoice is not visible outside of the package.  Since there is a [setter method for SdlChoice](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSdlChoiceChosen.java#L150-156), developers should be able to create an SdlChoice object to set.  They are currently unable to do so since the constructor is not public.

mikeburke106:
Possibly related to #14 because SdlChoice is more-or-less the same as Choice, which is redundant and violates the DRY principle.

PR #118, opened 2015-02-18
Enum improvements

### LINKED ISSUE #79: DisplayType valueForString throws a runtime exception

### LINKED ISSUE #91: Enum valueForString methods are not consistent
See PR #82 for more details.

Most enum classes simply use the `name()` method to define the value of the JSON value for each particular enum value.  Therefore, the `valueForString` method typically calls and returns the result of the `valueOf()` method.

If `valueForString` is called with a JSON value that is not contained in the enum, the [`valueOf` method will cause an IllegalArgumentException](http://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#valueOf(java.lang.Class,%20java.lang.String)).

However, in enum classes that define their own internal name to represent the JSON value, the `valueForString` method loops through all enum values and returns the value if it's found.  If it's not found, it will return null.

All enums should behave identically in this case, so we need to determine if the correct course of action is to throw a RuntimeException (causing a crash) or to return null.

### LINKED ISSUE #102: ButtonName indexForPresetButton throws NullPointerException
Passing a null value into [indexForPresetButton](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/ButtonName.java#L133-170) throws a null pointer exception.

### LINKED ISSUE #119: SdlDisconnectedReason convertAppInterfaceUnregisteredReason throws NullPointerException
Passing a null value into [convertAppInterfaceUnregisteredReason](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/SdlDisconnectedReason.java#L29-69) throws a null pointer exception.

mikeburke106:
This hotfix will fix any reported minor issues with RPC enum classes.

  - [x] Fix #79
  - [x] Fix #91 
  - [x] Fix #102
  - [x] Fix #119
  - [x] Remove try/catch blocks when calling valueForString

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/proxy/RPCResponse.java
+++ b/sdl_android_lib/src/com/smartdevicelink/proxy/RPCResponse.java
@@ -4,8 +4,8 @@
 package com.smartdevicelink.proxy;
 
 import java.util.Hashtable;
+
 import com.smartdevicelink.proxy.rpc.enums.Result;
-import com.smartdevicelink.util.DebugTool;
 
 /**
  * Result sent by SDL after an RPC is processed, consists of four parts: 
@@ -130,13 +130,7 @@ public class RPCResponse extends RPCMessage {
         if (obj instanceof Result) {
             return (Result) obj;
         } else if (obj instanceof String) {
-            Result theCode = null;
-            try {
-                theCode = Result.valueForString((String) obj);
-            } catch (Exception e) {
-                DebugTool.logError("Failed to parse " + getClass().getSimpleName() + "." + RPCResponse.KEY_RESULT_CODE, e);
-            }
-            return theCode;
+            return Result.valueForString((String) obj);
         }
         return null;
     }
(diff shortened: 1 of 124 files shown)


mikeburke106:
When this PR is merged, make sure to note #92 added `valueOf()` where it didn't exist before.  Ensure VrCapabilities enum is updated with this change once #92 is merged as part of #113.

mikeburke106:
Trivial change is being marked for 3-day offline review. Anticipated merge date is 2/23 at 3pm.

mikeburke106:
Here are the results of the current unit tests for this PR.

Tests Run | Test Passes (%) | Test Failures (%) | Test Errors (%)
---------------|-------------------|-------------------|-----------------
1486 | 1476 (99%) | 9 (<1%) | 1 (<1%)

Of the 10 failed tests, 9 are related to #71, and 1 is related to #89.  Both of these issues are slated to be merged by version 3.3.0, so all unit tests should be passed by March 1 or earlier.

PR #120, opened 2015-02-18
Fix for setBulkData ignoring null input

### LINKED ISSUE #103: BulkData ignores null inputs
The [setBulkData](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/RPCStruct.java#L61-67) method in RPCStruct ignores null values.  So, once a value is set for bulk data, it can never be removed unless a new array is passed in.

mikeburke106:
Fixes #103


mikeburke106:
Trivial change is being marked for 3-day offline review. Anticipated merge date is 2/23 at 3pm.

PR #123, opened 2015-02-23
Better bulk data handling in OnSystemRequest

### LINKED ISSUE #86: OnSystemRequest has a lot of bugs
[OnSystemRequest](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java) doesn't really follow the guidelines and practices of the rest of the project.  It seems like it was hurriedly thrown together and therefore, it has a lot of bugs.

I can see [here](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java#L44-57) that the class is attempting to create a JSON object from the binary bulk data.  If the BulkData is valid, it will save 2 global variables, `httpreqparams` and `myJSONObj`.  If the bulk data is not valid, both of these variables will be null.

In the [setHeaders method](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java#L125-131), if the `httpreqparams` is null, it will cause a NullPointerException crash.

The body of the request is even more confusing.  In the [getBody method](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java#L86-114), it is trying to read from `httpreqparams` if that's not null.  Even though `httpreqparams` is already the object returned from the `"HTTPRequest"` key, the code is reading the exact same value from the original bulk data and then reading the `"body"` key from that.  If the `httpreqparams` is null, the code just returns the original bulk data as a string, which I don't really understand.  Why are we just blindly assuming that whatever is in the bulk data is the body of an HTTP request in this one case?  In addition, in the [setBody method](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java#L116-122), the body information is stored in the parameters hashtable, which isn't even where it came from and obviously doesn't belong here.

I think this whole class should be re-written, to be honest.

mikeburke106:
Fixes #86.

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java
+++ b/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java
@@ -75,7 +75,6 @@ public class OnSystemRequest extends RPCNotification {
         String result = null;
         
         try{
-            // TODO: does body need to conform to the funky logic in existing getBody method?
             result = httpJson.getString("body");
         }catch(JSONException e){
             Log.e("OnSystemRequest", "\"body\" key doesn't exist in bulk data.");
--- a/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java
+++ b/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java
@@ -6,6 +6,8 @@ import java.util.List;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import android.util.Log;
+
 import com.smartdevicelink.marshal.JsonRPCMarshaller;
 import com.smartdevicelink.protocol.enums.FunctionID;
 import com.smartdevicelink.proxy.RPCNotification;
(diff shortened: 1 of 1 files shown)


mikeburke106:
This is technically a hotfix since no APIs have changed, but I wanted to make sure the appropriate parties have time to review, so I'm marking this one for the regularly scheduled review.

I ran unit tests on this PR and all tests were passed, so I'm confident that it's good to go.

Only open question is whether the body field needs to conform to [this weird logic](https://github.com/smartdevicelink/sdl_android/blob/a381ea4fac2676e209b0530fee251a5e935acd28/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/OnSystemRequest.java#L88-113) in the existing code.

```java
JSONObject jLayer1 = null;
String sReturn = null;        
try
{
    if (httpreqparams != null)
    {
        jLayer1 = myJSONObj.getJSONObject("HTTPRequest");        	
        sReturn = jLayer1.getString("body");
        return sReturn;
    }
    else if (myJSONObj != null)
    {
        return myJSONObj.toString();
    }
    else
    {
        return null;
    }
}
catch (Exception e) 
{
    return null;        	
}
```

The first if statement is the "normal" case where it reads the body from the HTTPRequest JSON message.  However, if the HTTPRequest JSON message doesn't exist, it is assumed the entire bulk data is the body.  Does the new code need to conform to this requirement?  Does the request ever come across with just the body like this?  Is there different behavior in different message protocol versions?

Any info I could get on this behavior would be very useful.  Thanks.

mrapitis:
I ran a round testing with the changes and so far things look good.  As far as the weird logic you mentioned, we should be safe in removing this… at one point the json inside of the binary data that we received from the module for a policy table update request differed in format from the json (in the bin data) that we received for a vehicle update request.  Since the module now provides the data in the same format for both types of onSystemRequests, we can safely remove the else if that you referenced.

mikeburke106:
Thanks @mrapitis.  I will remove my TODO comment and this PR should be all set.

PR #124, opened 2015-02-23
Deprecated Jingles class and added Jingle enum

### LINKED ISSUE #110: Jingles should be an enum instead of a class
[Jingles.java](https://github.com/smartdevicelink/sdl_android/blob/5a9cc2c6067a6063a1e1441a0c0194ca4215cf2a/sdl_android_lib/src/com/smartdevicelink/proxy/constants/Jingles.java) is defined as a class with only static variables defined.  This should be changed to be an enum instead.

mikeburke106:
Fixes #110.


mikeburke106:
This is a simple change, but since an API was deprecated, this counts as a minor version change according to semantic versioning.  Slating for 3.4.0 release on 3/16.

PR #125, opened 2015-02-23
Delete Ford-proprietary SiphonServer.

### LINKED ISSUE #51: Purpose/neccessity of SiphonServer?
Is [this class](https://github.com/smartdevicelink/sdl_android/blob/5a9cc2c6067a6063a1e1441a0c0194ca4215cf2a/sdl_android_lib/src/com/smartdevicelink/transport/SiphonServer.java) still needed? It seems like it's mostly just for debugging and I think there are much better ways to handle debugging rather than this. (Lot less overhead).

mikeburke106:
Fixes #51


idamaster:
@MikaylaRay44 did we actually pull siphon server from the proxy, or are we waiting for #190 to address this problem?

MikaylaRay44:
The siphon server, as of now, is still in the project but the change was made (along with related & replacement logging) in the logging consolidation branch #190 that is awaiting merge approval. I closed this to avoid duplication and the work involved in updating this branch before merging since it is already going to be taken care of in the other.

PR #153, opened 2015-03-23
Updated rpcs to v4.0 

joeygrover:
:cat2: Ready for review :cat2: 

The following enums need to be updated to match the latest version of the 4.0 mobile api spec:

- [x] LayoutMode (Add and deprecate)
- [x] AppInterfaceUnregisteredReason
- [x] TextFieldName
- [x] ImageFieldName
- [ ] AmbientLightStatus (Did not see updates needed)
- [x] VehicleDataResultCode
- [x] KeyboardEvent
- [x] RequestType
- [ ] Result (Didn't see updates needed, added descriptions to existing values)

The following classes need to be updated to match the latest version of the 4.0 mobile api spec:

- [x] TouchEvent    (Changed timestamp (ts) to use Long instead of Integers)
- [x] HMICapabilities (new)
- [x] FunctionID
- [x] RegisterAppInterface (response)
- [x] PutFile (request)  (Changed offset and length to use Long instead of Integers)
- [x] DialNumber (new - request, response)
- [x] OnSystemRequest (notification) (Changed offset and length to use Long instead of Integers)

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
+++ b/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
@@ -1,7 +1,7 @@
 package com.smartdevicelink.Dispatcher;
 
-public interface IDispatchingStrategy<messageType> {
-	public void dispatch(messageType message);
+public interface IDispatchingStrategy<T> {
+	public void dispatch(T message);
 	
 	public void handleDispatchingError(String info, Exception ex);
 	
--- a/sdl_android_lib/src/com/smartdevicelink/Dispatcher/ProxyMessageDispatcher.java
+++ b/sdl_android_lib/src/com/smartdevicelink/Dispatcher/ProxyMessageDispatcher.java
@@ -5,17 +5,17 @@ import java.util.concurrent.PriorityBlockingQueue;
 
 import com.smartdevicelink.util.DebugTool;
 
-public class ProxyMessageDispatcher<messageType> {
-	PriorityBlockingQueue<messageType> _queue = null;
+public class ProxyMessageDispatcher<T> {
+	PriorityBlockingQueue<T> _queue = null;
 	private Thread _messageDispatchingThread = null;
-	IDispatchingStrategy<messageType> _strategy = null;
+	IDispatchingStrategy<T> _strategy = null;
 
 	// Boolean to track if disposed
 	private Boolean dispatcherDisposed = false;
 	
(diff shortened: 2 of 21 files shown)


mrapitis:
The changes in LayoutMode are very subtle, basically just an added 'S' on two enums.

https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/LayoutMode.java#L5

https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/LayoutMode.java#L7

joeygrover:
Oh got ya, should they be changed inline?(breaking change) Or added and deprecate the old enum values without the 's'? (non-breaking change) 

My guess is that we need both the old and new versions of the enum which is not ideal.

joeygrover:
Also, should enums that have 'internal_name'. and 'name be updated to be INTERNAL_NAME(NAME). HMI_STATUS is an example of that.

However, I feel the enum values should remain the actual values/names. The encompassing enum type should be descriptive enough and there's no since creating more code complexity if we can avoid it.

Heath-FoMoCo:
I think the Layout Mode ENUM changes are actual spec fixes.  I think the implementation in both proxies are correct (with S).

joeygrover:
This is whats on the develop branch for Android https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/enums/LayoutMode.java

They're missing the 'S' so we need to confirm what core sends pre v4.

joeygrover:
Alright I believe I have most of the changes implemented, however I'm still unclear on what changed in  the AmbientLightStatus and Result enums. Also, we need confirmation on the LayoutMode issue. 

After that we can schedule a review.

stefanek22:
The LayoutMode change should be removed (meaning there should be no move to add a plural "S").

joeygrover:
Thank you @stefanek22 for clearing this up. This is definitely best. I will make the changes.

justinjdickow:
For AmbientLightStatus we'll use `AmbientLightStatus.UNKNOWN`

idamaster:
Can we get a quick update on this PR? Panasonic is asking for updated functionality, which I think is dependent on this...?

joeygrover:
I have to check with the confirmed MOBILE_API.XML v4 spec since the one supplied for this pull request was incorrect.

joeygrover:
Ready for review during the next SDL call.

PR #155, opened 2015-03-24
Fixed issue with onProxyClosed not always called in multiplexing scenari...

### LINKED ISSUE #152: onProxyClosed not always received in app multiplexing scenarios
When a transport error occurs (such as bt disconnect), an onProxyClosed notification should be sent to each multiplexed app.  Sometimes a concurrent modification exception occurs in the InternalMsgDispatcher class that prevents the onTransportError method from being called.  As a result, the onProxyClosed method is never called and the app never receives the notification.  The listenerlist Vector in SdlConnection needs to be changed to a list that is thread safe.  The logic in the onTransportError for loop also need to be revised.

mrapitis:
Fixes #152 in commit ID b16708c3075025efd2ace23ff1207502704df5ab

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
+++ b/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
@@ -1,7 +1,7 @@
 package com.smartdevicelink.Dispatcher;
 
-public interface IDispatchingStrategy<messageType> {
-	public void dispatch(messageType message);
+public interface IDispatchingStrategy<T> {
+	public void dispatch(T message);
 	
 	public void handleDispatchingError(String info, Exception ex);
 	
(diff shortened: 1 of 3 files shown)


justinjdickow:
Joey to investigate lock removal and optimize for CopyOnWriteArrayList convenience methods

joeygrover:
@mrapitis Can you give me a quick step-by-step on how you were testing this? Or what scenario made the adverse effects take place?

mrapitis:
From a single app, create two proxy instances with different app names / app id's.  After the apps have registered, ignition cycle the tdk.  A concurrent modification exception will sometimes occur before the 2nd app has a chance to notify the app with an onProxyClosed message.

joeygrover:
Ok I've tested my thoughts and they work.

1) Remove SESSION_LOCK as it's no longer needed with CopyOnWriteArrayList. Only one write operation can occur at a time by default.

2) https://github.com/smartdevicelink/sdl_android/blob/b16708c3075025efd2ace23ff1207502704df5ab/sdl_android_lib/src/com/smartdevicelink/SdlConnection/SdlConnection.java#L310

Can change to just:

``` Java
		listenerList.addIfAbsent(registerListener);
```

If you have any questions, let me know.

PR #164, opened 2015-04-17
RPC Callbacks

joeygrover:
:smirk_cat:  Ready for review :smirk_cat:

## Summary
Purpose of this pull request is to allow developers to get callbacks when specific RPCs are finished. This is most useful for RPC request that contain useful information in their responses. The callbacks are set up so in the future we can extend them to include specific information about that RPC or use the general callback to get a response for any type of RPC.

##### General
- [x] Create basic OnUpdateListener callback
- [x] Add the callback to every RPC during the parse process
- ~~Create logic that would check for a callback before sending it through the generic interface~~ This should be discussed at a later time.

### Code Example:
This is how a developer could set their callback. Follows the Android standard. 
##### NOTE: Only onFinish is required to be overwritten. The others are optional.

````Java
					Show show = new Show();
                                        //Fill out show request
					show.setOnRPCUpdateListener(new OnRPCUpdateListener(){	    	
						@Override
						public void onStart(int correlationId) {
							Log.w("OnUpdateListener", "onStart ");
						}

						@Override
						public void onError(int correlationId,Result resultCode,String info) {
							Log.w("OnUpdateListener",  "onError " + resultCode + " - " + info);
						}

						@Override
						public void onFinish(int correlationId, RPCMessage message) {
							Log.w("OnUpdateListener", "onFinish ");
							
						}
				    	
				    });
````

This is an example of putfile callback (This is how specific rpc callbacks can be created)
````Java
					PutFile putFile = new PutFile();
					putFile.setFileType(FileType.GRAPHIC_PNG);
					putFile.setSdlFileName(ICON_SYNC_FILENAME);
					putFile.setCorrelationID(iIconCorrID);
					putFile.setBulkData(contentsOfResource(R.raw.fiesta));
					putFile.setOnPutFileUpdateListener(new OnPutFileUpdateListener(){	    	
						@Override
						public void onStart(int correlationId,long totalSize) {
							Log.w("OnPutFileUpdateListener", "onStart ");
						}

						@Override
						public void onUpdate(int correlationId,long bytesWritten,long totalSize) {
							Log.w("OnPutFileUpdateListener", "onUpdate ");
						}

						@Override
						public void onError(int correlationId,Result resultCode,String info) {
							Log.w("OnPutFileUpdateListener",  "onError " + resultCode + " - " + info);
						}

						@Override
						public void onFinish(int correlationId, RPCMessage message, long totalSize) {
							Log.w("OnPutFileUpdateListener", "onFinish ");
							
						}
				    	
				    });
````


joeygrover:
Look into creating tiers of listeners to better conform to specific RPCs.

@Heath-FoMoCo

justinjdickow:
@Heath-FoMoCo Can you please review Joey's changes here to see if they align with your needs for callbacks?

mrapitis:
@joeygrover 

A couple of minor things noticed in testing the branch: READ_DID is missing one of the onPacketFinish calls.  Please see below:

https://github.com/smartdevicelink/sdl_android/blob/feature/rpc_callbacks/sdl_android_lib/src/com/smartdevicelink/proxy/SdlProxyBase.java#L2451-L2464

Also the test for equality below is failing since we need to use .equals instead of == for objects (this was causing the subscribe button RPC request to fail to receive call backs).

https://github.com/smartdevicelink/sdl_android/blob/feature/rpc_callbacks/sdl_android_lib/src/com/smartdevicelink/proxy/SdlProxyBase.java#L1554-L1556

May also need to check for other places in the SdlProxyBase source code to make sure we use the proper equality checks as a separate issue.

joeygrover:
Thanks for the feedback @mrapitis . Committed those fixes, and pushed.

mrapitis:
@joeygrover this is very minor, but it looks like we have introduced two new warnings with the merge of this PR (we cleaned up all the warnings in a previous PR).  Would you like me to write a new issue so we can target?

joeygrover:
Yea that sounds good.

Issue #181, opened 2015-05-07
SendLocation ClassCastException From JSON Value


MikaylaRay44:
> com/smartdevicelink/proxy/rpc/SendLocation

While I was writing the unit tests for the SendLocation class a ClassCastException from a getter method was revealed on one of the three calls made to test it.

The first test on the getLatitudeDegrees method, which returns a Float, was the null value test and it passes.
```java
SendLocation msg = new SendLocation();
assertNotNull("Null object creation failed.", msg);
assertNull("Latitude wasn't set, but getter method returned an object.", msg.getLatitudeDegrees());
```
> Results:
> Latitude Float Type Test: false
> Latitude Integer Type Test: false
> Latitude Double Type Test: false
> Latitude Object Type Test: false
> Latitude Null Value Test: true

The second test was the manual value test through setLatitudeDegrees, and it passes this also.
```java
SendLocation msg = new SendLocation();
msg.setLatitudeDegrees(42f);
assertEquals("Latitude didn't match input latitude.", 42f, msg.getLatitudeDegrees());
```
> Results:
> Latitude Float Type Test: true
> Latitude Integer Type Test: false
> Latitude Double Type Test: false
> Latitude Object Type Test: true
> Latitude Null Value Test: false

However, it fails on the third test, where the value is pulled from the JSON RPC test files.
```java
// Other json code ...
JSONObject parameters = JsonUtils.readJsonObjectFromJsonObject(body, RPCMessage.KEY_PARAMETERS);
assertEquals("Latitude doesn't match input latitude.", JsonUtils.readDoubleFromJsonObject(parameters, SendLocation.KEY_LAT_DEGREES).floatValue(), 42f);
```
> Results:
> Latitude Float Type Test: false
> Latitude Integer Type Test: false
> Latitude Double Type Test: true
> Latitude Object Type Test: true
> Latitude Null Value Test: false

As you can see the type that is being read in is viewed as a double, because JSON does not have any specific format for handling floats and views all numbers with decimals as doubles.

```java
public Float getLatitudeDegrees(){    	
    return (Float) parameters.get(KEY_LAT_DEGREES);
}
```

The ClassCastException occurs because you cannot cast the type Double to a Float. A solution to this issue, if the value is read from the JSON file as a Double or Integer, to use the wrapper classes .floatValue() method.

```java
public Float getLatitudeDegrees(){    	
    if (parameters.get(KEY_LAT_DEGREES) instanceof Integer) {
[...]

PR #182, opened 2015-05-07
Fixes issue #181 by adjusting Float casting.

MikaylaRay44:
See issue #181 for details.

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
+++ b/sdl_android_lib/src/com/smartdevicelink/Dispatcher/IDispatchingStrategy.java
@@ -1,7 +1,7 @@
 package com.smartdevicelink.Dispatcher;
 
-public interface IDispatchingStrategy<messageType> {
-	public void dispatch(messageType message);
+public interface IDispatchingStrategy<T> {
+	public void dispatch(T message);
 	
 	public void handleDispatchingError(String info, Exception ex);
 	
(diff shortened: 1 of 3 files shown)


mikeburke106:
This is a developer-facing breaking change that could break developer code.

```java
Float longitude = sendLocation.getLongitudeDegrees(); // won't compile after this change
```

If this is a change that really needs to be made, technically the methods with Float type should be deprecated and replaced so existing developer code isn't affected.  This also allows the change to be brought in in 4.* instead of 5.0 under semantic versioning.

MikaylaRay44:
The problem with deprecating the getter and setter methods is that there will be library conflicts that won't compile.

```java
public Float getLongitudeDegrees()
public Double getLongitudeDegrees() // This will cause a conflict
```

However, the code will compile that you mentioned before, but it will cause a `ClassCastException` because you cannot cast a Double to a Float in that manner, you would need to call .floatValue() or change your current value to a Double.

So in light of that, even renaming the legacy code method will still cause this runtime exception to occur. Any ideas on how to solve that issue then? @mrapitis @mikeburke106

mrapitis:
Can we please discuss further in our next PR review.

mikeburke106:
@MikaylaRay44 Ahh, good point.  I don't think you have a choice, then, you'll have to make it a breaking change and it won't be able to be released until 5.0 under semantic versioning.  It also needs to be documented for developers to change their type to Double to fix the error that will be introduced in their code.

mikeburke106:
Also, instead of doing your reflection (`instanceof`) in each of the methods, you should use the [SdlDataTypeConverter.objectToDouble method](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/util/SdlDataTypeConverter.java#L5-19) to be consistent with other classes.  Once the breaking change is made, you should no longer have to account for the floating point type.

Issue #198, opened 2015-05-19
Mime NullPointerExceptions


MikaylaRay44:
> _com/smartdevicelink/trace/Mime_

`NullPointerException` is thrown by the following unit tests on the [Mime](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/trace/Mime.java) class.

```java
String actualNullResult1 = Mime.base64Encode((byte[]) null); // Exception
String actualNullResult2 = Mime.base64Encode((String) null);
String actualNullResult3 = Mime.base64Encode(null, 0, 0);  // Exception

// Invalid/Null Tests
assertNull(Test.NULL, actualNullResult1);
assertNull(Test.NULL, actualNullResult2);
assertNull(Test.NULL, actualNullResult3);
```

The first test fails because it calls the third, which has no parameter checking.

However, the second test does not fail ...

```java
public static String base64Encode(String str) {
    String b64String = "";
	try {
		byte[] strBytes = str.getBytes("US-ASCII");
		b64String = base64Encode(strBytes);
	} catch (Exception ex) {
		// Don't care?
	} 
	return b64String;
}
```

It does nothing with the exception that is caught. Instead of having the `// Don't care?` comment it should return that the encoding has failed via a null value.

Solution : To fix both occurrences of the exception some null-value parameter checking needs to be done.

Issue #195, opened 2015-05-18
DiagLevel NullPointerExceptions


MikaylaRay44:
> _com/smartdevicelink/trace/DiagLevel_

`NullPointerException` is thrown by the following methods int the [DiagLevel](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/trace/DiagLevel.java) test class.

``` java
try {
    // Invalid/Null Tests
    assertFalse(Test.FALSE, DiagLevel.isValidDetailLevel(testInvalidS));
    assertFalse(Test.FALSE, DiagLevel.isValidDetailLevel(null)); // Exception here
    assertNull(Test.NULL, DiagLevel.getLevel(null)); // Exception here
    DiagLevel.setLevel(null, null); // Exception here
    DiagLevel.setAllLevels(null); // Exception here
} catch (NullPointerException e) {
    fail("NullPointerException was thrown when attempting to set null values.");
}
```

Solution : The methods should verify that the parameters are not null before attempting to manipulate the data.

bilal-alsharifi:
Fixed in this commit https://github.com/smartdevicelink/sdl_java_suite/commit/19363b7b89621d4cfd0b571734fd112fcfbd3af1

Issue #193, opened 2015-05-18
BitConverter NullPointerExeptions


MikaylaRay44:
> _com/smartdevicelink/util/BitConverter_

The following unit tests for the [BitConverter](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/util/BitConverter.java) class caused a `NullPointerException` to occur.

```java
// Invalid/Null Tests
assertNull(Test.NULL, BitConverter.bytesToHex(null)); // Exception
...
int actualNullBytes = BitConverter.intFromByteArray(null, 0); // Exception
assertEquals(Test.MATCH, (int) 0, actualNullBytes);
...
short  actualNullBytes = BitConverter.shortFromByteArray(null, 0); // Exception
assertEquals(Test.MATCH, (short) 0, actualNullBytes);
```

Solution : Each of these methods should verify that the parameters given are not null before attempting to manipulate them. Similar methods in this class do check, but not all of them.

Issue #191, opened 2015-05-15
ByteEnumer ClassCastExceptions


MikaylaRay44:
> _com.smartdevicelink.util.ByteEnumer_

Calling the `get` methods in the [ByteEnumer](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/util/ByteEnumer.java), which accepts a Vector of ***any*** type as a parameter, will cause a `ClassCastException` if provided a Vector of any type other than [ByteEnumer](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/util/ByteEnumer.java).

```java
// Test
String testString1 = "test_1";
String testString2 = "test_2";

Vector<String> testInvalidList = new Vector<String>();
testInvalidList.add(testString1);
testInvalidList.add(testString2);

assertNull("Value should be null.", ByteEnumer.get(testInvalidList, testByte1));
assertNull("Value should be null.", ByteEnumer.get(testInvalidList, testString1));

// Methods Called
public static ByteEnumer get(Vector<?> theList, byte value) {
	Enumeration<?> enumer = theList.elements();
	while (enumer.hasMoreElements()) {
		ByteEnumer current = (ByteEnumer)enumer.nextElement();
		if (current.getValue() == value) {
			return current;
		}
	}
	return null;
}
	
public static ByteEnumer get(Vector<?> theList, String name) {
	Enumeration<?> enumer = theList.elements();
	while (enumer.hasMoreElements()) {
		ByteEnumer current = (ByteEnumer)enumer.nextElement();
		if (current.getName().equals(name)) {
			return current;
		}
	}
	return null;
}

// Result
java.lang.ClassCastException: java.lang.String cannot be cast to com.smartdevicelink.util.ByteEnumer
```

There are two possible solutions to this :
1)  If these methods are going to allow a list of any type, then they should _at least_ do some validation to avoid this exception from occurring.
2) Replace the `<?>` notation that allows any type of list with `<ByteEnumer>`, however, I believe the original intention of this was to allow list of classes that extend [ByteEnumer](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/util/ByteEnumer.java) to be searched through so the first proposal is the best and will not cause any potentially bad trickle effects.

PR #204, opened 2015-05-29
Minor exception cleanups.

MikaylaRay44:
See #191, #193, #195, and #198.

Note: So I just realized some of these have already been handled and accepted in other pull requests, I can remove them if you'd like, but they are already accepted changes so let me know. (i.e. issue 191 was pull request 192).

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/trace/DiagLevel.java
+++ b/sdl_android_lib/src/com/smartdevicelink/trace/DiagLevel.java
@@ -3,50 +3,105 @@ package com.smartdevicelink.trace;
 import com.smartdevicelink.trace.enums.DetailLevel;
 import com.smartdevicelink.trace.enums.Mod;
 
-
 public class DiagLevel {
 
 	static private DetailLevel[] levels;
 	
-	static {   // this is a static c-tor!!
+	// This is a static constructor.
+	static {
 		levels = new DetailLevel[Mod.values().length];
 		setAllLevels(DetailLevel.OFF);
 	}
 	
-	public static void setAllLevels(DetailLevel thisDetail) {
-		for (int i=0; i < levels.length; i++) {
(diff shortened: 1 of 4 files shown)


mrapitis:
@MikaylaRay44 Can you please double check the for loops that have been replaced?

https://github.com/smartdevicelink/sdl_android/blob/feature/exception_cleanup/sdl_android_lib/src/com/smartdevicelink/util/BitConverter.java#L64 always initializes the hexArrayIndex value to 0.... its also generally preferred to initialize variables outside of the for loop.
https://github.com/smartdevicelink/sdl_android/blob/feature/exception_cleanup/sdl_android_lib/src/com/smartdevicelink/util/BitConverter.java#L84-85
https://github.com/smartdevicelink/sdl_android/blob/feature/exception_cleanup/sdl_android_lib/src/com/smartdevicelink/util/BitConverter.java#L64-67

We may also want to add some checking around offset values to avoid index out of bounds situations.

thanks

MikaylaRay44:
Yes I don't know why I moved that index value, I will move it back.

As for the first for loop:

```java
for (int i = 0; i < hexString.length(); i += 2) {
	String byteString = hexString.substring(i, i + 2);
	byte theByte = (byte)Integer.parseInt(byteString, 16);
	theBytes[i/2] = theByte;
}
```

I didn't change this from how it appears on the develop branch, but I agree it looks concerning because of the downward cast.

I would suggest changing it to this to prevent potential loss of data, but it's not a *needed* change. I have added it to the pull request, I can remove it if you think it's not necessary.

```java
byte theByte = ((Integer) Integer.parseInt(byteString, 16)).byteValue();
```

Also the offset values do have checks in the Mime and BitConverter classes, but not all of them are documented so I've updated that. Did you see them in a different class?

Issue #194, opened 2015-05-18
TransportType Class Location


MikaylaRay44:
> _com/smartdevicelink/transport/TransportType

Shouldn't the [TransportType](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/transport/TransportType.java) enum class be located in it's own package `com.smartdevicelink.transport.enums` to keep consistant with the other enum packages like `com.smartdevicelink.rpc.enums` and `com.smartdevicelink.trace.enums`?

Shouldn't the structure of all of the packages be uniform? Interface sub-packages and enum sub-packages included. See #207 for implementation.

PR #206, opened 2015-06-02
transport.enums package

MikaylaRay44:
See issue #194 for details.

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/SdlConnection/SdlConnection.java
+++ b/sdl_android_lib/src/com/smartdevicelink/SdlConnection/SdlConnection.java
@@ -8,6 +8,7 @@ import java.io.PipedOutputStream;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import android.util.Log;
+
 import com.smartdevicelink.exception.SdlException;
 import com.smartdevicelink.protocol.AbstractProtocol;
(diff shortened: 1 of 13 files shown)


mrapitis:
Hi Mikayla,

Can you double check the imports on this pull request as well?  Some of the files have been changed to add imports without corresponding removals.

Thanks,
Markos

MikaylaRay44:
For the `SdlConnection` class a bulk import was used for the transport package classes because it uses so many of them, but because the TransportType enum was moved to a different package namespace it would need to have a separate package import. I can't remove the bulk import because of the other dependencies, this specific import was added because the class moved locations otherwise it was covered in the bulk import as well.

As for the other transport package classes that did not have import removals like `BTTransport`,`BTTransportConfig`, etc. the TransportType class was previously in the same package as those classes so an extra import was not needed, but now it is in a new package namespace so it needs an import statement.

Even though we call the enums package a sub-package of the transport package there is no parental hierarchy between the two in Java so they are treated as completely different. That's all that's happening, I only added these statements where such errors occurred because of the move.

mrapitis:
@MikaylaRay44 thanks for the follow up, have you also verified that no new warnings have been added to the project as a result of the changes?

MikaylaRay44:
Yep, no warnings or suppressed warnings are seen on my end.

PR #225, opened 2015-06-23
Removes deprecation warnings in SdlProxyBase class.

MikaylaRay44:
Removes deprecation warnings by deprecating old methods related to the change with integer offset/length values and replaced them with the proper methods that accept long values. Documentation included.


Heath-FoMoCo:
Two issues with this request.

First, and most important, this pull request did not follow the process.  It was merged in the same day it was committed.  It never got tagged for offline review.  And there was no discussion around it what-so-ever.  It goes without saying that there is no point in having a process if we do not follow it.  Please don't do this again.

Second, this breaks putfile streaming.  The methods replacing the deprecated methods are not the equivalent.  We will submit an issue against this with more detail.

Heath-FoMoCo:
Issue #226 
https://github.com/smartdevicelink/sdl_android/issues/226

Issue #243, opened 2015-08-05
Update proxy to track both incoming and outgoing heartbeats


mrapitis:
The proxy needs to be updated to keep track of when a heartbeat request is received from the module and provide a corresponding ACK back.  In addition the proxy should keep track of when a heartbeat request is sent to the module and its corresponding ACK.  In both cases, if a response is not received within the specified timeout, a heartbeat timeout should occur.

justinjdickow:
@mrapitis Does "heartbeat timeout" mean anything other than disconnecting the proxy and notifying the developer?

mrapitis:
@justinjdickow That is its entire purpose.

PR #244, opened 2015-08-05
Feature/heartbeat update

mrapitis:
Regarding issue #243


sdl_android_lib/src/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -282,16 +282,21 @@ public void onProtocolSessionStarted(SessionType sessionType,
...
-				 HeartbeatMonitor heartbeatMonitor = new HeartbeatMonitor();
-				 heartbeatMonitor.setInterval(_transportConfig.getHeartBeatTimeout());
-	             sdlSession.setHeartbeatMonitor(heartbeatMonitor);
-			 }			
+			setWiProVersion(version);	
 			
-			if (sessionType.eq(SessionType.RPC)) {			
+			if (sessionType.eq(SessionType.RPC)) {	
+
+				 if ( (_transportConfig.getHeartBeatTimeout() != Integer.MAX_VALUE) && (version > 2))
```
justinjdickow: On Core we use 0 to indicate not using heartbeat, is it possible to align with that paradigm?

mrapitis:
I'm pretty sure we could make that change, but can we create the request as a separate issue to target and investigate as its not something that was changed in this pr?

Issue #247, opened 2015-08-12
Improve SDK inline documentation 


mrapitis:
The inline documentation on our develop branch is out of date and needs improvement. I will submit a pull request with content (from the team) with some initial improvements, but this will be an on-going task.

PR #250, opened 2015-08-12
Feature/documentation update

mrapitis:
Regarding issue #247


joeygrover:
@mrapitis Is there any more work that plans to be done here? Or should we merge in what's here?

mrapitis:
I think we should be good to merge, I can remove the WIP in the title.

PR #251, opened 2015-08-12
Update USBTransport connection info to match SDL

### LINKED ISSUE #224: USB Transport explicitly checks for Ford head unit metadata
https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/transport/USBTransport.java#L471-L478

The manifest of an app should define the manufacturers that are supported and filter out ones that are not. The explicit checks should be removed so that apps are not only connecting to Ford head units.

justinjdickow:
Resolves #224


justinjdickow:
The usb_metadata in sdl_core is merged, so this needs to be merged in order for the Android libraries to work with AOA against release/4.0.0 in core

PR #275, opened 2015-12-14
Hotfix/mtu size for v3 & v4 protocols

mrapitis:
The mtu size that is specified by core for versions 3 and 4 protocols do not match the value in the mobile library.  Core (and the gen 3 module) set the value to 131072-12 as can be seen in the smartDeviceLink.ini file.  This change updates WiProProtocol.java to match these values.


joeygrover:
As a note, this was actually reverted and not merged into any branch.

PR #281, opened 2016-02-04
Added localization class.

kshala-ford:
The purpose of the localization class is to help out when the phone/app language differs from the SDL/head unit language. The class can load a bundle instance based on a specified language/region.


joeygrover:
How will this class be used? Where in the proxy should it be implemented? Need to make sure there is a proper use case before merging in new classes.

Issue #289, opened 2016-02-24
Proxy doesn't handle hybrid service packets correctly


joeygrover:
Currently the proxy will throw away any packets that come through the Hybrid/bulk data service.

```Java
if (message.getSessionType().equals(SessionType.RPC)) {
...
}else{
     // Handle other protocol message types here
}
```

See https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/SdlProxyBase.java#L1332

Quickest fix is to just add an or statement that includes if it's type Hybrid/bulk data
```Java
if (message.getSessionType().equals(SessionType.RPC) || message.getSessionType().equals(SessionType. BULK_DATA ) {...}
````

PR #290, opened 2016-02-25
Fix proxy to not discard hybrid or bulk data service packets

joeygrover:
In reference to #289


mrapitis:
Just a note, our infosys team is currently testing this change against various hardware platforms.

mrapitis:
Quick status update, all test results were positive regarding this pr.

PR #300, opened 2016-05-24
[WIP] Feature/api lockscreen

tstrayer:
This pull request is dependent upon https://github.com/smartdevicelink/sdl_android/pull/299 and must be merged after it.

This feature provides automatic lock screen management for developers. It provides the following features:
- LockScreenActivity that is overridden to specify the desired layout for the lock screen
- LockScreenActivity enforces trapping of the onBackPressed method to prevent users from navigating away from the lock lock screen while it is displayed.
- LockScreenConfig and associated builder that is used to configure the lock screen to be used by the lock screen management
- Automatic lock screen management from the SdlManager class that will non destructively add and remove the lock screen to the Application's back stack.


tstrayer:
This pull request should be targeted to develop not master.

PR #301, opened 2016-05-24
[WIP] Feature/api view

tstrayer:
This pull request is dependent upon https://github.com/smartdevicelink/sdl_android/pull/299 and must be merged after it.

This feature provides view structure for handling the head unit HMI.
- Provides SdlTemplateView for automatically picking, updating and verifying template support based on child views supplied.
- Provides SdlTextView, SdlGraphicView, and SdlButtonView corresponding to the different regions of the layouts provided by the SDL spec.
- Automatically manages Putfile requests for views that contain graphics by checking for their presence in the registry and sending them if they are not present.
- Will send all non-graphic elements and any pre-loaded graphic elements immediately and will add other graphics as they are transferred to the module.
- Support for sending and reacting to Alerts sent as AlertDialogs automatically limited to HMI_FULL
- Support for sending and reacting to PushNotifications sent as AlertDialogs if permissions are available.


tstrayer:
This is also depends on https://github.com/smartdevicelink/sdl_android/pull/302 and must be merged after it. 

The AlertDialog and PushNotification classes rely on permission information to ensure that they are allowed operations to prevent the transmission of RPCs that will be rejected.

PR #302, opened 2016-05-24
[WIP] Feature/permission manager

tstrayer:
This feature adds centralized processing and management of permissions.

Developers can register a permission listener with the permission manager to be notified when permissions they are interested in change. This listener will be called IFF permissions in the filter provided to the manager with the listener change.

Developers can also poll permission information about specific permissions from the manager.

An Enum is provided that describes all of the potential permissions that are returned in the OnPermissionChange object.

The PermissionManager will also listen to the HMI state of the proxy and use that information to correctly drive permission listener callbacks based on the current state to notify the listener of what actions can be taken at the present time instead of requiring that this statefulness is handled by the developer.


tstrayer:
This should be targeted to develop.

Issue #307, opened 2016-06-21
Multiple instances of SDLRouter running confuses SDLProxy, not connecting to the router with a HU connection.


jthrun:
In the generic situation this is probably an edge case where two SDL apps check for a running router almost simultaneously and, finding none, each start their routers resulting in two routers running at the same time.  Our situation is different in that we have a companion app that only starts its router once it detects a connection to a known head unit.  The companion app's router will be started regardless of other routers running since only the companion app's router will get a connection from the HU SDLCore.  Unfortunately, the other SDL apps do not always find and connect to this router even though it broadcasts the 'sdl.router.startservice' message for triggering SDLProxy startup.  What I see in the log is that SdlBroadcastReceiver validates the router that broadcast the message but the proxy (or actually the TransportBroker) later tries to check for a transport connected router and always finds the non-connected router first, short-circuiting and returning false.

In order to repro this you need a "companion" app that will start its SdlRouter when it gets a connection to a "known" HU.  This app's SdlRouter needs to use a custom BT UUID which the SDLCore on the "known" HU will connect to to the exclusion of other SdlRouters broadcasting with the standard UUID.

1. Install the "companion" app and other SDL multiplexer-enabled apps on the HS device
2. Restart the HS
3. Start the HU, SDLCore, HMI
4. Try to run one of the SDL apps

What I see in the logs is that the "companion" app's router gets a connection from SDLCore on the HU and then broadcasts the TransportConstants.START_ROUTER_SERVICE_ACTION message.  Then, other apps' SdlBroadcastReciever check and validate the broadcasting router.  Finally, SDLProxies for those apps check to see of a running router has a transport connection but get back a false response because they checked the wrong running router.  I'm attaching a section of logcat logs which contains messages from two routers (pid 3504, 5368) and one of the apps (pid 3630) which shows what happens.  Short story:
```
06-21 13:30:26.974  3630  3630 V Sdl Broadcast Receiver: SDL enabled by router service from com.uievolution.companionapp compnent package com.uievolution.companionapp - com.main.companionapp.sdl.SdlRouterService
...
06-21 13:30:27.028  3630  3630 D SdlTransportBroker: Bound to service ComponentInfo{com.uievolution.companionapp/com.main.companionapp.sdl.SdlRouterService}
...
06-21 13:30:27.031  3630  5561 D Sdl Broadcast Receiver: Looking for Service: sdlrouterservice
[...]

PR #308, opened 2016-06-21
Potential fix for issue 307

jthrun:
https://github.com/smartdevicelink/sdl_android/issues/307.

One possible way of propagating the broadcasting router info down to the transport connection.


joeygrover:
I'll take a look at this tomorrow, but I think you're on the right path from just looking at the changes.

sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -31,6 +31,8 @@
 		
 	public static ComponentName runningBluetoothServicePackage = null;
 
+	private static ComponentName broadcastServicePackage = null;
```
joeygrover: You don't need to store this, have the TransportBroker already has this info

joeygrover:
You should have the TransportBroker just supply the component name of the service it's bound to [here](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/transport/TransportBroker.java#L163) using this [variable](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/transport/TransportBroker.java#L47) .

jthrun:
Joey,
I've updated the code per your suggestions.  Can you take another look?

joeygrover:
Keeping this open until #312 is reviewed and accepted, but it should take care of the issue this PR solves.

joeygrover:
#312 was merged and #307 was confirmed to be solved. Closing this PR.

PR #312, opened 2016-06-27
Remove deprecated methods and correct hardware connect check

joeygrover:
Removed the dependency on the shared preference model that was deprecated by Android a long time ago. Replaced it with a small class that will bind to the router service and request the connection status. Classes interested in finding the connection status out can create an instance of this class and pass in a callback that will notify them of the connection status once the router service responds.

This also takes care of #307 by eliminating the transport broker from using the shared preferences but instead the connection information is just passed to it from the router service during the client's registration.


mrapitis:
Testing so far looks good, just a couple of minor things:

Is there a way for us to avoid the handler leak here:
https://github.com/smartdevicelink/sdl_android/blob/6b0cfb157e6b2b5e7d1415ec4e40e52b99f84d84/sdl_android_lib/src/com/smartdevicelink/transport/SdlRouterStatusProvider.java#L103

Should we rename isTransportConnected since it now returns a void with a call back passed in instead?
https://github.com/smartdevicelink/sdl_android/blob/6b0cfb157e6b2b5e7d1415ec4e40e52b99f84d84/sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java#L205

joeygrover:
Alright, renamed the method and fixed the handler leak issue. Both good suggestions.

mrapitis:
:thumbsup:

PR #315, opened 2016-07-01
Fixed ClassCastException in openConnection

thomasheike:
From the crash reports in PlayStore I receive some ClassCastExceptions every now and then on devices (seems to be exclusive to tablets) running Android 4.1 and 4.2.
The exceptions occur in BTTransport.openConnection when some exception is cast to SdlException.

The reports contain no line numbers ... I suspect line 185 to be the cause of the error, since this is the only line containing an unchecked cast to SdlException.


joeygrover:
Closing due to it being marked to go into master. Please resubmit this pull request with it marked to merge into develop and not master.

PR #323, opened 2016-07-07
Fixed ClassCastException in openConnection

thomasheike:
From the crash reports in PlayStore I receive some ClassCastExceptions every now and then on devices (seems to be exclusive to tablets) running Android 4.1 and 4.2.
The exceptions occur in BTTransport.openConnection when some exception is cast to SdlException.

The reports contain no line numbers ... I suspect line 190 to be the cause of the error, since this is the only line containing an unchecked cast to SdlException.


joeygrover:
If you can [sign the CLA](https://docs.google.com/a/livioradio.com/forms/d/1VNR8EUd5b46cQ7uNbCq1fJmnu0askNpUp5dudLKRGpU/viewform) I'll be able to review/accept this.

thomasheike:
Done. :)

Issue #259, opened 2015-09-03
StreamPackatizer limited to 1488


joeygrover:
https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/streaming/StreamPacketizer.java#L61

````Java
length = is.read(buffer, 0, 1488);
````

This should be changed to check the current MTU - HeaderSize

joeygrover:
Fixed with pull request #324

Issue #172, opened 2015-04-28
TLS Handshake implementation


justinjdickow:
Is TLS handshake implemented in the Android proxy available on the develop branch?

@joeygrover @mrapitis

mrapitis:
As far as I know, the TLS handshake is not implemented on any branch yet.

joeygrover:
Fixed with merge of #324

PR #324, opened 2016-07-20
Feature/external security support

mrapitis:
Summary

Regarding issue #172 

This PR adds the capability to hook in a security library to the SDL library. The library will take care of automatically instantiating an object based on the make passed alongside it matching the make of the vehicle connected through an RAI. The library defines a protocol that must be followed by the security library's public interface.


joeygrover:
Looks pretty good without really being able to test, but there are a few concerns.

- Is there a test/dummy external security lib that I can test with?
- Have you created any unit tests that can be merged into the unit test branch after this is merged?
- In SdlProxyBase, [there are multiple hardcoded ints that should be a constant](https://github.com/smartdevicelink/sdl_android/pull/324/files#diff-c6637341d1fcd4efdb60197db8912badR3444)
- In RPCStruct, [the method to check if the payload is protected should be isPayloadProtected not getPayloadProtected](https://github.com/smartdevicelink/sdl_android/pull/324/files#diff-7dc1f9d993e0ab28db16d4d47c19a498R133)
- The library was scrubbed of the "compression" bit in favor of the encryption nomenclature. Are we sure this has never been used? Seems more accurate to deprecate first than to totally replace.
- More documentation is needed in general, but specifically [SdlSecurityBase](https://github.com/smartdevicelink/sdl_android/pull/324/files#diff-6de361308e74721aa1b68b0168652386). The class itself should have documentation on what it is, the abstract methods should be explained as to what their purposes are, etc.
- When merged it looks as if this will also fix issue #259

mrapitis:
@joey good feedback and suggestions, thanks. Please see response below:
<<Is there a test/dummy external security lib that I can test with? >>
Yes, we will provide access to a test security lib that can be used to establish a secure session with sdl core.
<<Have you created any unit tests that can be merged into the unit test branch after this is merged?>>
We can initially supply some basic tests that can be merged on top of the unit test branch. We will continue to build new cases for better testing.
<<In SdlProxyBase, there are multiple hardcoded ints that should be a constant>>
Good point we will get this updated.
<<In RPCStruct, the method to check if the payload is protected should be isPayloadProtected not getPayloadProtected>>
Good point we will get this updated.
<<The library was scrubbed of the "compression" bit in favor of the encryption nomenclature. Are we sure this has never been used? Seems more accurate to deprecate first than to totally replace.>>
Internally the compression flag was never used, but I’m not sure that we can guarantee that from an open perspective.  Since the protocol spec changed the field description and usage entirely the code was updated to match the new definition.
<<More documentation is needed in general, but specifically SdlSecurityBase. The class itself should have documentation on what it is, the abstract methods should be explained as to what their purposes are, etc.>>
Agree, we can add additional info.  Would it be possible to target this as a separate github task and PR as this may take more time?
<<When merged it looks as if this will also fix issue #259>>
Glad we can resolve this one as well.

mrapitis:
@joeygrover can you please re-review when you have a moment?  We have added some commits that should address some suggestions.  Also the unit test branch can be found here:
https://github.com/smartdevicelink/sdl_android/tree/feature/unit_tests_security

Thanks!

PR #325, opened 2016-07-27
Add a library to secure preferences used in library

joeygrover:
This PR is looking to add a library that will help secure each apps stored preferences. This layer of security adds onto the existing layer provided through Android's private preferences to begin with.

The change is very minor code wise.


mrapitis:
@joeygrover code side changes look to be minimal, but can we get more information about the secured_prefs.jar library included?
thanks!

Heath-FoMoCo:
This is the first time we are hearing about this change.  I assume this is going to go thru the evolution process as agree for all further enhancements?

joeygrover:
The secured_prefs.jar is actually a closed-source project that will be owned and maintained by the SDL maintainers and governing body. All of the qualified parties will have access. This is due to its nature of improving security. This is of course up for discussion, however, I believe it is the best course of action for this enhancement. 

This change was actually discussed back in the September/October time frame of last year (2015) with interested parties, at which there was no dissention. The project has been in a private repo and shared with the appropriate parties since that time. However, I'm more than willing to go through the evolution process for this feature since no one has went through it before and this can serve as an example. So please refer to [my proposal. ](https://github.com/smartdevicelink/sdl_evolution/pull/1)

PR #337, opened 2016-10-07
AOA improvements

joeygrover:
This pull request is to make it much easier for developers to use AOA/USB. Currently the way to make this work is very confusing. 

This will allow developers to use the onSdlEnabled callback in the SdlBroadcastReceiver to start their service with the proper USB extras. Then in the onStart method in their service, they can check the intent for the proper extras.

``` java
public void startProxy(Intent intent) {
        if (proxy == null) {
            try {
                BaseTransportConfig transport = null;
                if(intent!=null && intent.hasExtra(UsbManager.EXTRA_ACCESSORY)){ //If we want to support USB transport
                    if(android.os.Build.VERSION.SDK_INT <= android.os.Build.VERSION_CODES.HONEYCOMB){
                        Log.e(TAG, "Unable to start proxy. Android OS version is too low");
                        return;
                    }
                    //We have a usb transport
                    transport = new USBTransportConfig(getBaseContext(),(UsbAccessory)intent.getParcelableExtra(UsbManager.EXTRA_ACCESSORY));
                }else{
                    //If we don't want anything but USB then would just do a return here and bail
                   // transport = new MultiplexTransportConfig(getBaseContext(), APP_ID);
                    Log.e(TAG, "Not usb intent. Shutting down");
                    stopSelf();
                    return;
                }
                proxy = new SdlProxyALM(this, APP_NAME,
                        true, APP_ID,
                        transport);
            } catch (SdlException e) {
                e.printStackTrace();
                // error creating proxy, returned proxy = null
                if (proxy == null) {
                    stopSelf();
                }
            }
        }
    }
```

A more detailed guide can be written around exactly how to connect the app via USB, but this is the main point.


mrapitis:
Would it be possible to push a hello_sdl_android branch that could be used with to test the changes?

joeygrover:
Yea I can do. I'll tag it here when pushed.

joeygrover:
See hello sdl repo for example: https://github.com/smartdevicelink/hello_sdl_android/tree/feature/aoa_improvements

Issue #344, opened 2016-10-18
Add API's associated with evolution task Last Mile Navigation


mrapitis:
Please see accepted SDL Evolution proposal below for details:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0003-last-mile-navigation.md

joeygrover:
Completed with #346

PR #346, opened 2016-10-18
API's associated with Last Mile Navigation evolution task

mrapitis:
Please see issue #344 for details.

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/protocol/enums/FunctionID.java
+++ b/sdl_android_lib/src/com/smartdevicelink/protocol/enums/FunctionID.java
@@ -53,6 +53,9 @@ public enum FunctionID{
     SYSTEM_REQUEST(38, "SystemRequest"),
     SEND_LOCATION(39, "SendLocation"),
     DIAL_NUMBER(40, "DialNumber"),
+    GET_WAY_POINTS(45, "GetWayPoints"),
+    SUBSCRIBE_WAY_POINTS(46, "SubscribeWayPoints"),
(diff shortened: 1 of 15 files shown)


mrapitis:
Thanks for the review.  We will get the enumeration style updated and push a new commit to address.  Could we target the float to double change in a separate issue and pull request as we are trying to match the previously accepted evolution task.  We can submit an issue for the lat & long difference and PR in parallel with this task so that it doesn't lose visibility.

joeygrover:
This PR should reflect the correct params over the evolution proposal. Implementation details are not 100% coming out of the evolution process as many unknowns could be discovered during the actual generation of code. The review, comments, and new commit should serve as a a good history and will be easier to follow for future references.

mrapitis:
Should we deprecate the existing methods and add new methods utilizing floats since they are already in use?

joeygrover:
Which methods are already in use? Any new methods introduced in this PR should just be changed since they were never in an official release they do not need to be deprecated.

mrapitis:
We can update the referenced Coordinate.java file since it applies directly to this PR, but the complete scope of this discussion https://github.com/smartdevicelink/sdl_core/issues/823 seems like it should span across to include the sendLocation getter and setter methods (which are currently using doubles).  It sounds like we will be targeting those changes as a separate task.

joeygrover:
Got ya. Yes the SendLocation updates can be a separate PR and those methods should be deprecated.

mrapitis:
@joeygrover can you please re-review when you have a moment? thanks!

Issue #394, opened 2017-01-23
Transport performance issues


jacobkeeler:
When looking into smartdevicelink/sdl_core#835, I found a few performance issues in the transport classes.

There are currently two messages ([here](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/TCPTransport.java#L388) and [here](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/TCPTransport.java#L404)) in TCPTransport.java on the master branch which are logged out every time a byte is read, slowing down the transport significantly. 

After removing these two logs, the transport seemed to speed up by around a factor of 3 (OnAudioPassThru notifications took ~3.5s to transfer rather than ~10s).

In addition, [TCPTransport](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/TCPTransport.java#L390), [BTTransport](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/BTTransport.java#L390), and [MultiplexBluetoothTransport](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/MultiplexBluetoothTransport.java#L793) each read from the input stream one byte at a time, which can be significantly less efficient than using a buffer in my experience. 

A simple fix to this without changing behavior would be to read into a buffer, then use a for loop to process the received bytes one at a time. 

Testing these changes seemed to have more of an effect on the TCP transport than anything. After implementing this change, I saw OnAudioPassThru notifications appear once every second (which is exactly how fast Core sends them out). 

Oddly, making the same optimization seemed to have very little effect on the BT transport, and I have not yet been able to test it's effect on the MultiplexBluetoothTransport.

I will create a PR with these proposed changes.

PR #395, opened 2017-01-23
Fix performance issues with TCP, BT and MultiplexBT transports

jacobkeeler:
Removed 2 incredibly verbose logs from TCP transport and modified TCP and BT transports to read using a buffer rather than byte-by-byte.

Corresponding issue: #394

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/transport/BTTransport.java
+++ b/sdl_android_lib/src/com/smartdevicelink/transport/BTTransport.java
@@ -30,7 +30,9 @@ public class BTTransport extends SdlTransport {
 	private final static UUID SDL_V4_MOBILE_APPLICATION_SVC_CLASS = new UUID(0x936DA01F9ABD4D9DL, 0x80C702AF85C822A8L);
 
 	private static final String SDL_LIB_TRACE_KEY = "42baba60-eb57-11df-98cf-0800200c9a66";
-	
+
+	private static final int READ_BUFFER_SIZE = 4096;
+
 	private BluetoothAdapter _adapter = null;
 	private BluetoothSocket _activeSocket = null;
 	private UUID _listeningServiceUUID = SDL_V4_MOBILE_APPLICATION_SVC_CLASS;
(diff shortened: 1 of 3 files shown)


jacobkeeler:
I created a spreadsheet for the metrics [here](https://docs.google.com/spreadsheets/d/1P9Pa8UtxHLHAFkkRSGotCmh6xwtz72BbupYyEqxT2mQ/edit?usp=sharing).

Fortunately, I was able to get uniform-length packets (88 kB each) for these tests because Core repeatedly sends out a dummy 1-second audio file to fake `AudioPassThru` when `EXTENDED_MEDIA_MODE` is disabled, so message length could be eliminated as a variable.

Each test is done by sending a PerformAudioPassThru message and logging the time it takes to construct the "OnAudioPassThru" notifications that are sent in response (calculating the time between when the PSM goes from START_STATE to FINISHED_STATE).

I only performed one run for each of the TCP cases, as these seemed to present more of an obvious difference in the performance of the transport. Because the Bluetooth and MultiplexBT transports seemed to be significantly less consistent with their timings, I ran each test 5 times and presented the average time it took to receive a packet.

For the TCP Transport, it is fairly straightforward that the buffering makes a large difference in performance. When the verbose log was present, the `PerformAudioPassThru` message had a high chance of timing out due to how long the notifications were taking to be received/processed, therefore I was unable to get more than a few points of data for this case. With the removal of the verbose log, the performance of the transport improved by around a factor of 3 (~2200ms per message vs ~6000ms). Once the buffering optimization was implemented, however, the time it took to receive and process the full message was reduced to a trivial ~50ms on average.

When using the Bluetooth transport, OnAudioPassThru messages were completed ~500ms faster when using a buffer compared to without. This gives the appearance that the buffering optimization does help the Bluetooth transport slightly, but it is a bit difficult to draw any conclusions because the times vary rather wildly.

MultiplexBT appears to be able to receive packets with significantly more consistent timings when using the buffer than without, although the first buffered test was significantly slower than each test after. All in all, however, the inclusion of the buffer appears to improve the MultiplexBT transport by around a factor of two on average.

chenbd:
i guess read buffer size will impact performance significantly when transfer huge amount of data like video stream.
this is true not only for this library but also for sdl core.
for example:when using usb transport transfer video stream, small read buffer in sdl core also cause performance issue.

Issue #375, opened 2017-01-19
java.lang.OutOfMemoryError when allocating byte array in SdlPsm.java


askirk:
Spotify reported a few occurances of OOM errors and provided the following stacktrace:

```
Fatal Exception: java.lang.OutOfMemoryError
       at com.spotify.music.internal.crashes.SpotifyExceptionHandler.processOutOfMemory(SourceFile:1054)
       at com.spotify.music.internal.crashes.SpotifyCrashlyticsCrashManager$3.uncaughtException(SourceFile:81)
       at java.lang.ThreadGroup.uncaughtException(ThreadGroup.java:693)
       at java.lang.ThreadGroup.uncaughtException(ThreadGroup.java:690)
Caused by java.lang.OutOfMemoryError: Failed to allocate a 1769173877 byte allocation with 4194304 free bytes and 251MB until OOM
       at com.smartdevicelink.transport.SdlPsm.transitionOnInput(SourceFile:156)
       at com.smartdevicelink.transport.SdlPsm.handleByte(SourceFile:54)
       at com.smartdevicelink.transport.MultiplexBluetoothTransport$ConnectedThread.run(SourceFile:795)
```

This issue relates to spotify crash numbers 722722 and 711556 and occurs in https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/SdlPsm.java#L199

askirk:
L199 has been surrounded with try/catch since version 4.2.1

PR #397, opened 2017-01-24
Bugfix/issue_375

askirk:
Resolving #375 

In first case, we don't need a try/catch because is the dataLength is over 1500, we know the data has been garbled.

In the second case we surround allocation of byte array with try/catch to catch the OOM case.
Not much else we can do about an erroneously large dataLength.

L178-182 is Mirroring https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/SdlPsm.java#L198-L202

-Tested with Gen 3 TDK

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/protocol/WiProProtocol.java
+++ b/sdl_android_lib/src/com/smartdevicelink/protocol/WiProProtocol.java
@@ -19,7 +19,7 @@ public class WiProProtocol extends AbstractProtocol {
 	byte _version = 1;
 	private final static String FailurePropagating_Msg = "Failure propagating ";
 
-	private static final int V1_V2_MTU_SIZE = 1500;
+	public static final int V1_V2_MTU_SIZE = 1500;
 	private static final int V3_V4_MTU_SIZE = 131072;
 	private static int HEADER_SIZE = 8;
 	private static int MAX_DATA_SIZE = V1_V2_MTU_SIZE  - HEADER_SIZE;
--- a/sdl_android_lib/src/com/smartdevicelink/protocol/WiProProtocol.java
+++ b/sdl_android_lib/src/com/smartdevicelink/protocol/WiProProtocol.java
@@ -22,7 +22,7 @@ public class WiProProtocol extends AbstractProtocol {
 	public static final int V1_V2_MTU_SIZE = 1500;
 	public static final int V3_V4_MTU_SIZE = 131072;
 	public static final int V1_HEADER_SIZE = 8;
-	public static final int V2_V4_HEADER_SIZE =12;
+	public static final int V2_HEADER_SIZE = 12;
 	private static int HEADER_SIZE = 8;
 	private static int MAX_DATA_SIZE = V1_V2_MTU_SIZE  - HEADER_SIZE;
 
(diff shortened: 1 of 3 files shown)


sdl_android_lib/src/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -19,7 +19,7 @@
 	byte _version = 1;
 	private final static String FailurePropagating_Msg = "Failure propagating ";
 
-	private static final int V1_V2_MTU_SIZE = 1500;
+	public static final int V1_V2_MTU_SIZE = 1500;
 	private static final int V3_V4_MTU_SIZE = 131072;
```
joeygrover: Also change this constant to be consistent

sdl_android_lib/src/com/smartdevicelink/transport/SdlPsm.java
```diff
@@ -153,7 +155,11 @@ private int transitionOnInput(byte rawByte, int state){
 					if(dataLength==0){
 						return FINISHED_STATE; //We are done if we don't have any payload
 					}
-					payload = new byte[dataLength];
+					if(dataLength <= V1_V2_MTU_SIZE){ // size from protocol/WiProProtocol.java
```
joeygrover: This should actually be MTU-HEADER_SIZE. Introduce statics into WiProProtocol for standard Header sizes. V1_HEADER_SIZE = 8, V2_V4_HEADER_SIZE = 12

sdl_android_lib/src/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -19,8 +19,10 @@
 	byte _version = 1;
 	private final static String FailurePropagating_Msg = "Failure propagating ";
 
-	private static final int V1_V2_MTU_SIZE = 1500;
-	private static final int V3_V4_MTU_SIZE = 131072;
+	public static final int V1_V2_MTU_SIZE = 1500;
+	public static final int V3_V4_MTU_SIZE = 131072;
+	public static final int V1_HEADER_SIZE = 8;
+	public static final int V2_V4_HEADER_SIZE =12;
```
joeygrover: This should just be V2_HEADER_SIZE. If we move to v5 without changing header size we'd have to change names, or leave it to be confusing.

sdl_android_lib/src/com/smartdevicelink/transport/SdlPsm.java
```diff
@@ -170,7 +177,11 @@ private int transitionOnInput(byte rawByte, int state){
 				if(dataLength == 0){
 					return FINISHED_STATE; //We are done if we don't have any payload
 				}
-				payload = new byte[dataLength];
+				try {
```
joeygrover: Since there is a check for a v1 packet right before this, we know only a v1 packet could enter this if statement and therefore would have to be less than or equal to V1_V2_MTU_SIZE - V1_HEADER_SIZE

askirk: should I just try/catch this then? this is where we got the OOM

joeygrover: Right, but you can actually use the same check as earlier that way we don't have to resort to using the try/catch
```    
    if(dataLength <= V1_V2_MTU_SIZE - V1_HEADER_SIZE){ 
        payload = new byte[dataLength];
    }else{
        return ERROR_STATE;
    }
```

Issue #382, opened 2017-01-20
NPE in MultiplexBluetoothTransport.ConnectedThread.cancel() : mmInStream


askirk:
Stacktrace:

```
Fatal Exception: java.lang.NullPointerException
       at android.bluetooth.BluetoothSocket.close(BluetoothSocket.java:465)
       at android.bluetooth.BluetoothInputStream.close(BluetoothInputStream.java:44)
       at com.smartdevicelink.transport.MultiplexBluetoothTransport$ConnectedThread.cancel(SourceFile:822)
       at com.smartdevicelink.transport.MultiplexBluetoothTransport.stop(SourceFile:290)
       at com.smartdevicelink.transport.MultiplexBluetoothTransport.stop(SourceFile:280)
       at com.smartdevicelink.transport.MultiplexBluetoothTransport.connectionLost(SourceFile:3348)
```

Occurs at https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/MultiplexBluetoothTransport.java#L822.

The if statement evaluates to true, and the value of mmInStream is set to null (assuming it is closed by some other cancel() call) before the inside of the if statement is executed.

askirk:
Propsal is to catch NPE as well as IOException in the catch block. Comment to specify that the purpose is to protect for mmInStream and no other variable.

Issue #380, opened 2017-01-20
NPE in MultiplexBluetoothTransport.ConnectedThread.run() : mmInstream is null


askirk:
Stacktrace:

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.io.InputStream.read()' on a null object reference
       at com.smartdevicelink.transport.MultiplexBluetoothTransport$ConnectedThread.run(SourceFile:793)
```

Can only be set to null in constructor, therefore socket.getInputStream() must have returned null at https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/MultiplexBluetoothTransport.java#L771

PR #399, opened 2017-01-24
Handling NULL mmInStream cases

askirk:
-addresses cases where mmInStream is null
-aimed at #380 and #382 
-tested by running sample app on Gen 3 TDK and toggling Bluetooth on and off


joeygrover:
Changed from pushing to master to develop.

Issue #376, opened 2017-01-20
Process does not have BLUETOOTH permission: java.lang.SecurityException


askirk:
Error

`Caused by java.lang.SecurityException: Need BLUETOOTH permission: Neither user 10227 nor current process has android.permission.BLUETOOTH.
       at android.os.Parcel.readException(Parcel.java:1620)
       at android.os.Parcel.readException(Parcel.java:1573)
       at android.bluetooth.IBluetooth$Stub$Proxy.getState(IBluetooth.java:1291)
       at android.bluetooth.BluetoothAdapter.getState(BluetoothAdapter.java:1086)
       at com.smartdevicelink.transport.SdlRouterService$3.onReceive(SourceFile:279)` 

was reported. It occured at

https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/SdlRouterService.java#L279

The following permissions should already be in the App Manifest for the app, need to investigate why they weren't in this instance.
```
<uses-permission android:name="android.permission.BLUETOOTH" />
<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
```

askirk:
https://developer.android.com/training/permissions/requesting.html#perm-check

askirk:
https://developer.android.com/reference/android/support/v4/content/ContextCompat.html#checkSelfPermission%28android.content.Context,%20java.lang.String%29

Issue #378, opened 2017-01-20
NPE in SdlRouterService.bluetoothAvailable() : BluetoothAdapter.getDefaultAdapter().isEnabled() call


askirk:
Stacktrace:

```
Caused by java.lang.NullPointerException
       at android.os.Parcel.readException(Parcel.java:1493)
       at android.os.Parcel.readException(Parcel.java:1436)
       at android.bluetooth.IBluetooth$Stub$Proxy.isEnabled(IBluetooth.java:596)
       at android.bluetooth.BluetoothAdapter.isEnabled(BluetoothAdapter.java:452)
       at com.smartdevicelink.transport.SdlRouterService.bluetoothAvailable(SourceFile:1027)
       at com.smartdevicelink.transport.SdlRouterService.shouldServiceRemainOpen(SourceFile:1051)
       at com.smartdevicelink.transport.SdlRouterService.onStartCommand(SourceFile:868)
```

Snippet in question:

```
private boolean bluetoothAvailable(){
		boolean retVal = (!(BluetoothAdapter.getDefaultAdapter()==null) && BluetoothAdapter.getDefaultAdapter().isEnabled());
		//Log.d(TAG, "Bluetooth Available? - " + retVal);
		return retVal;
}
```

The exception is raised within the isEnabled method, and it is not an error due to invoking a method on a null object reference. (BluetoothAdapter.getDefaultAdapter() is not null.)

askirk:
Proposed replacement:
```
private boolean bluetoothAvailable(){
		boolean retVal = (BluetoothAdapter.getDefaultAdapter() != null) ? BluetoothAdapter.getDefaultAdapter().isEnabled() : false;
		//Log.d(TAG, "Bluetooth Available? - " + retVal);
		return retVal;
}
```

askirk:
This could relate to #376

askirk:
Could try/catch for NPE, return false if we get the exception

PR #405, opened 2017-01-25
Check for BT Permission in SdlRouterService.onCreate()

askirk:
- To resolve issue #376 
- this may contribute to #378 
- We just stop the service if permission isn't there, any further handling will be done at a future time

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 23
+    compileSdkVersion 18
     buildToolsVersion "23.0.2"
     
     lintOptions {
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 18
+    compileSdkVersion 23
     buildToolsVersion "23.0.2"
     
     lintOptions {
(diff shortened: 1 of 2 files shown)


sdl_android_lib/src/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -53,6 +54,7 @@
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.os.RemoteException;
+import android.support.v4.content.ContextCompat;
```
joeygrover: We don't want to use any of the support libraries. That would force our developers to use the same support library. I suggest trying to use the normal context method if android version is 23+, or fully qualified contextcompat with a try catch

sdl_android_lib/src/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -783,6 +785,11 @@ public void onCreate() {
 			stopSelf();
```
joeygrover: This section is getting a bit large and should be moved to a single function with a name like initCheck() that returns a boolean if the checks passed.

build.gradle
```diff
@@ -13,7 +13,7 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 18
+    compileSdkVersion 23
```
joeygrover: We can't up the compile version with approval from the SDLC

sdl_android_lib/src/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -851,6 +875,8 @@ public void startUpSequence(){
 
 	@Override
 	public int onStartCommand(Intent intent, int flags, int startId) {
+		if(!initPassed)
```
joeygrover: Please include brackets around this if statement

Issue #381, opened 2017-01-20
NPE in SdlBroadCastReceiver.isRouterServiceRunning()


askirk:
Stacktrace:

```
Fatal Exception: java.lang.RuntimeException: Unable to bind to service com.spotify.mobile.android.service.SpotifyService@1cf86fa8 with Intent { cmp=com.spotify.music/com.spotify.mobile.android.service.SpotifyService }: java.lang.NullPointerException: Attempt to read from field 'android.content.ComponentName com.android.server.am.ServiceRecord.name' on a null object reference
       at android.app.ActivityThread.handleBindService(ActivityThread.java:2787)
       at android.app.ActivityThread.access$1900(ActivityThread.java:147)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1369)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:135)
       at android.app.ActivityThread.main(ActivityThread.java:5253)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:900)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:695)
Caused by java.lang.NullPointerException: Attempt to read from field 'android.content.ComponentName com.android.server.am.ServiceRecord.name' on a null object reference
       at android.os.Parcel.readException(Parcel.java:1546)
       at android.os.Parcel.readException(Parcel.java:1493)
       at android.app.ActivityManagerProxy.getServices(ActivityManagerNative.java:3091)
       at android.app.ActivityManager.getRunningServices(ActivityManager.java:1579)
       at com.smartdevicelink.transport.SdlBroadcastReceiver.isRouterServiceRunning(SourceFile:201)
```

Error occurs at
https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java#L196

askirk:
Proposed solution:
Wrap the inside of for loop with
`if (service != null)`

*may want to also check that ActivityManager that is grabbed is not null

askirk:
grab the list and do a regular for loop

PR #406, opened 2017-01-27
Replace range based for loop with index-based

askirk:
- Issue #381 
- Iterating through list indices instead of range-based for loop

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
+++ b/sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
@@ -1,5 +1,6 @@
 package com.smartdevicelink.transport;
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Vector;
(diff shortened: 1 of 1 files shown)


sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -193,8 +195,16 @@ private static boolean isRouterServiceRunning(Context context, boolean pingServi
...
-	    for (RunningServiceInfo service : manager.getRunningServices(Integer.MAX_VALUE)) {
-	    	//We will check to see if it contains this name, should be pretty specific
+		List<RunningServiceInfo> runningServices = null;
+		try{
+			runningServices = manager.getRunningServices(Integer.MAX_VALUE);
+		}catch(NullPointerException e){
+			Log.e(TAG, "Can't get list of running services");
+			return false;
+		}
+	    for (Iterator<RunningServiceInfo> i = runningServices.iterator(); i.hasNext();) {
```
joeygrover: `for(RunningServiceInfo service : runningServices) {....}`

Issue #392, opened 2017-01-20
java.lang.SecurityException starting Service in SdlBroadcastReceiver.OnReceive.isRouterServiceRunning()


askirk:
Stacktrace:

```
Caused by java.lang.SecurityException: Unable to start service Intent { cmp=smartdevicelink.service.SdlRouterService (has extras) }: Unable to launch app for service Intent { cmp=smartdevicelink.service.SdlRouterService }: process is bad
       at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1358)
       at android.app.ContextImpl.startService(ContextImpl.java:1306)
       at android.content.ContextWrapper.startService(ContextWrapper.java:606)
       at android.content.ContextWrapper.startService(ContextWrapper.java:606)
       at com.smartdevicelink.transport.SdlBroadcastReceiver.isRouterServiceRunning(SourceFile:211)
       at com.smartdevicelink.transport.SdlBroadcastReceiver.wakeUpRouterService(SourceFile:157)
       at com.smartdevicelink.transport.SdlBroadcastReceiver.onReceive(SourceFile:141)
```

Occurs at https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java#L206

It may be useful to look at Manifest file.

askirk:
Add try/catch and bail if we get the exception.

PR #409, opened 2017-01-31
Bugfix/issue 392

askirk:
-Protecting against SecurityException found in issue #392

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
+++ b/sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
@@ -203,7 +203,12 @@ public abstract class SdlBroadcastReceiver extends BroadcastReceiver{
 	            	Intent intent = new Intent();
 	            	intent.setClassName(service.service.getPackageName(), service.service.getClassName());
 	            	intent.putExtra(TransportConstants.PING_ROUTER_SERVICE_EXTRA, pingService);
-	            	context.startService(intent);
+	            	try {
+						context.startService(intent);
+					}catch (SecurityException e){
+						Log.e(TAG, "Process is bad");
+						android.os.Process.killProcess(android.os.Process.myPid()); // Let's exit, we can't start our service
+					}
 	            }
 	        }
 	    }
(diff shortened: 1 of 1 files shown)


sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -203,7 +203,12 @@ private static boolean isRouterServiceRunning(Context context, boolean pingServi
 	            	Intent intent = new Intent();
 	            	intent.setClassName(service.service.getPackageName(), service.service.getClassName());
 	            	intent.putExtra(TransportConstants.PING_ROUTER_SERVICE_EXTRA, pingService);
-	            	context.startService(intent);
+	            	try {
+				context.startService(intent);
+			}catch (SecurityException e){
+				Log.e(TAG, "Process is bad");
+				android.os.Process.killProcess(android.os.Process.myPid()); // Let's exit, we can't start our service
```
joeygrover: We don't want to do this. Killing myPid process will actually kill the entire app. We just want to fail silently here, add a return statement there instead.

sdl_android_lib/src/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -203,7 +203,12 @@ private static boolean isRouterServiceRunning(Context context, boolean pingServi
 	            	Intent intent = new Intent();
 	            	intent.setClassName(service.service.getPackageName(), service.service.getClassName());
 	            	intent.putExtra(TransportConstants.PING_ROUTER_SERVICE_EXTRA, pingService);
-	            	context.startService(intent);
+	            	try {
+                        context.startService(intent);
+					}catch (SecurityException e){
+						Log.e(TAG, "Security exception, process is bad");
+						return false; // Let's exit, we can't start the service
```
joeygrover: So we actually can't just return false blindly here. If you take a look at the return statement after this for loop you can see we need to actually return something meaningful. I'd move the ping code into a separate method with the try/catch so that we can continue through the loop and return something meaningful. 

Something like this:
```java 
	protected void pingRouterService(Context context,String packageName, String className){
		if(context == null || packageName == null || className == null){
			return;
		}
		try{
			Intent intent= new Intent();
			intent.setClassName(packageName, className);
			intent.putExtra(TransportConstants.PING_ROUTER_SERVICE_EXTRA, true);
    		context.startService(intent);
		}catch(SecurityException e){
			
		}
	}
```

Issue #420, opened 2017-02-10
USBTransportConfig keeps accessory reference after disconnect


mrapitis:
When creating a USBTransportConfig object a UsbAccessory can be provided, however after a transport disconnect occurs the reference to the accessory is never cleared leaving the transport in a bad state with a reference to the unavailable accessory.  

https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/transport/USBTransport.java#L354-L442

PR #421, opened 2017-02-10
Bugfix/issue 420

mrapitis:
See issue #420

### DIFF
--- a/sdl_android_lib/src/com/smartdevicelink/transport/USBTransport.java
+++ b/sdl_android_lib/src/com/smartdevicelink/transport/USBTransport.java
@@ -360,6 +360,8 @@ public class USBTransport extends SdlTransport {
         }
         _disconnecting = true;
 
+        mConfig.setUsbAccessory(null);
+
         final State state = getState();
         switch (state) {
             case LISTENING:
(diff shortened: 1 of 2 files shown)


sdl_android_lib/src/com/smartdevicelink/transport/USBTransport.java
```diff
@@ -487,7 +491,7 @@ private void initializeAccessory() {
      * @param accessory Accessory to check
      * @return true if the accessory is right
      */
-    private boolean isAccessorySupported(UsbAccessory accessory) {
+    public static boolean isAccessorySupported(UsbAccessory accessory) {
```
askirk: Why make it public?

mrapitis: The app has ability to create a transport config object with a UsbAccessory object (as seen in the link below), however their is no publicly available method for the app to use to check if the accessory is supported.  

https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android_lib/src/com/smartdevicelink/transport/USBTransportConfig.java#L29

PR #422, opened 2017-02-12
fix compiler error for "catch (IOException|NullPointerException e)" and   	remove redundant "public|abstract" for interface methods

chenbd:


joeygrover:
Hi chenbd,

Thanks for taking interest in contributing to the SmartDeviceLink project. There are a few notes here to help you  get started:

1.  You should check out the [Contributing Guide ](https://github.com/smartdevicelink/sdl_android/blob/develop/CONTRIBUTING.md)and you'll need to sign the CLA. The link to the form is linked in the contributing guide. 
2. All pull requests should be finely scoped. .Combining too many fixes into one PR can make it difficult for the reviewer to ensure a thorough review, and also introduces the risk of possibly missing something that should be changed. For example, the change to change the catch statement back to a pre Java 7 style should be its own pull request with no other changes included. 
3. While I agree with a lot of the code changes here (redundant modifiers in method signatures) It's important to note that every code change has to be evaluated and tested. Therefore, making code changes that mostly affect the cosmetics/redeable of code vs the functionality should be weighed accordingly to their worth in terms of review and testing.
4. All SDL projects follow semantic versioning. Therefore, whenever changing public methods in a library project such as this it requires the library to increase its major version. So something as simple as changing `constructPacket` to `toByteArray` will force the library to move from 4.x to 5.x. force the library to move from 4.x to 5.x. We'd also recommend as a best practice to avoid simply removing public methods without first deprecating them for a few releases.

For these reasons I will close this pull request, but encourage you to look over the provided material and open new pull requests that follow those guidelines. Again, thank you for your interest in contributing to the project and we look forward to having your help!

chenbd:
@joeygrover 
1, CLA has been signed
2, my mistake, too many commits to the same branch
3, agree with you
4, all public methods in library change needs version up? even if the methods only used in library and not used by app?

joeygrover:
Unfortunately yes. Since we have no way of knowing if a developer uses one of the public methods (for whatever reason) we need to ensure that we don't break their implementations.

PR #423, opened 2017-02-15
fix compiler error for "catch (IOException|NullPointerException e)"

chenbd:
This pull request fix compiler error for "catch (IOException|NullPointerException e)" under java 7


joeygrover:
This is addressed by a recent PR to up the Java version used to 1.7.

Issue #445, opened 2017-03-31
There has a stream performance effect in present realize of PipedInputStream and PipedOutputStream


BrandonHe:
We found that currently in PipedInputStream in java.io package is using function `write()` to stream **Byte by Byte**, it will have performance effect:
```
    @Override
    public void write(int oneByte) throws IOException {
        PipedInputStream stream = target;
        if (stream == null) {
            throw new IOException("Pipe not connected");
        }
        stream.receive(oneByte);
    }
```
This will have a big influence in massive video stream data transmission, will cause very frequently read/write thread switch, reduce the transmission rate speed and increase the usage of CPU resource.

Additional, during the mobile navigation app development, the video stream performance is not as good as we expected, such as fuzzy during drag in HMI side.

There is a better way to write data to PipedInputStream: read the length of write buffer one time and write these bytes via `System.arraycopy(bytes, offset, buffer, in, bytesToTransfer)`. For more code details, please refer [here](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/io/PipedInputStream.java#226)

We have tested with the rewrite PipedInputStream and PipedOutputStream code in Ubuntu and Windows, both have a result of performance improvement.

I will send a PR for this feature improvement.

PR #446, opened 2017-04-03
Feature/enhance video streaming

BrandonHe:
### Summary:
This PR is relative to #445, in order to enhance the video stream performance.

### In Depth:
In I/O operation, we're using class `PipedInputStream` and `PipedOutputStream`.  For writing thread, we found that data was written into buffer by each single Byte. After per byte was written into buffer, the reading thread was waked up (Thread state is changed from a blocked state to an executable state). It starts polling while the CPU is idle, run the read thread(Thread state is changed from an executable state to implementation state). 

But at this point, writing thread may occupy the critical resource of the thread, leading to read thread blocking(Thread state is changed from implementation state to blocking state). Frequently switch CPU state will reduce the CPU operating efficiency, and increase power consumption. 

In addition, read thread seize the critical resources in an inappropriate time will also reduce the efficiency of proxy transmission. For example, after write thread just writes a few bytes, almost at the same time, read thread gets the resources begin to read, the data in buffer will have to split into many frames to transmit.

So, we rewrite the class `PipedInputStream` and class `PipedOutputStream`, for more details, please reference #445 and modified codes.
### Performance Test:
**PC: Debian 32-bit, Windows 32-bit**
**Phone: Android4.4(XiaoMI mi3)**
**Connection: USB_AOA**
**Result:**
The data in test are false video stream data, aims to get the max value in transmission.

|             | SDL(Debian)|
|:---------|:------:|
|Proxy (**NOT** include this PR) | 1408.868Kb/s  |
|Proxy (**include** this PR)         | 2227.984Kb/s  |

|             | SDL(Windows)|
|:---------|:------:|
|Proxy (**NOT** include this PR) | 1.2Mb/s  |
|Proxy (**include** this PR)         | 7.5Mb/s  |

### DIFF
--- a/sdl_android/src/main/AndroidManifest.xml
+++ b/sdl_android/src/main/AndroidManifest.xml
@@ -1,4 +1,3 @@
 <manifest package="com.smartdevicelink" xmlns:android="http://schemas.android.com/apk/res/android">
 	<uses-sdk android:minSdkVersion="8"/>
-    <application android:debuggable="true"/>
 </manifest>


BrandonHe:
@joeygrover 
Could you help to review this PR?
Thanks!

joeygrover:
Hi Brandon,
In the future if you're PR is ready for review you can request a reviewer and we will get notified as well. However, looking at this PR I think this might need an evolution proposal. This PR is to basically create a custom implementation of a java class that we would have to support moving forward. I'd like to know if any other solutions were considered (different java classes, refactor code to not need pipes, etc). All of that could be explained and laid out in an evolution proposal. So before I can review and accept I would need to ensure the SDLC gets a chance to review the high level idea and approve.

BrandonHe:
@joeygrover 
Hi Joey,
Thanks for your response. I'd like to create an evolution proposal if it's necessary. But for now, we don't have other solutions for this PR.

joeygrover:
I don't feel as though I can review/merge this unless the SDLC accepts this as a proposal. You already have most of the info between the issue and the PR. So please submit a proposal with that information and if accepted I will review/merge this PR.

chenbd:
@BrandonHe about video streaming performance, have you tried sending video data directly from encoder using byte buffer and handler instead of pipe to stream packetizer?
from my experience this will result in best performance.

BrandonHe:
@joeygrover 
Thanks for your explanation! I will submit a proposal to ask SDLC to review first. 
Based on your information, can I understand that submit a proposal is our standard code contribution process?

BrandonHe:
Hi @chenbd, Thanks for your suggestion, we have tried directly send encoder data using byte buffer, it doesn't have much improvement compared with this solution. I'm also afraid it will not be accepted by SDLC using byte buffer and handler to stream.

joeygrover:
@BrandonHe[ this document ](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals_versus_issues.md)should help you determine when a PR should have a proposal or not. 

The improvement of just sending the packets straight from the encoder would be to avoid having to create our own custom pipe streams.

Issue #447, opened 2017-04-03
Possible NPE when calling handleInitResult


mrapitis:
The SdlSecurityBase class can encounter a null pointer exception when calling handleInitResult and the session is null.  Please see attached stacktrack for details.  We should be able to check the session for null before calling onSecurityInitialized

https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/security/SdlSecurityBase.java#L47

` Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.smartdevicelink.SdlConnection.SdlSession.onSecurityInitialized()' on a null object reference
       at com.smartdevicelink.security.SdlSecurityBase.handleInitResult(SdlSecurityBase.java:48)
`

PR #448, opened 2017-04-04
Fix a NPE when calling handleInitResult

mrapitis:
Fixes issue #447

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
@@ -43,9 +43,10 @@ public abstract class SdlSecurityBase {
     }
     
     public void handleInitResult(boolean val) {
-    	setInitSuccess(val);
-		if (session != null)
-			session.onSecurityInitialized();
+        if (session == null) return;
+
+        setInitSuccess(val);
+        session.onSecurityInitialized();
     }
     
     public void handleSdlSession(SdlSession val) {
--- a/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
@@ -44,7 +44,8 @@ public abstract class SdlSecurityBase {
     
     public void handleInitResult(boolean val) {
     	setInitSuccess(val);
-    	session.onSecurityInitialized();
+		if (session != null)
+			session.onSecurityInitialized();
     }
     
     public void handleSdlSession(SdlSession val) {


sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
```diff
@@ -44,7 +44,8 @@ public void resetParams() {
     
     public void handleInitResult(boolean val) {
     	setInitSuccess(val);
-    	session.onSecurityInitialized();
+		if (session != null)
+			session.onSecurityInitialized();
```
joeygrover: If it's possible this gets called without a session attached to it, do we need to fire off the `session.onSecurityInitialized();` callback when a session is set? [See where we handle setting a session](https://github.com/smartdevicelink/sdl_android/blob/41a6698a2cbd7b3428fc3be55efe30fc21c3965a/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java#L55)

mrapitis: We still need to onSecurityInitialized to let the session know the security library is ready to go.  The session being null is more of an edge case related to timing around the disconnect as the close method and resetParams have the ability to set the session to null.

https://github.com/smartdevicelink/sdl_android/blob/41a6698a2cbd7b3428fc3be55efe30fc21c3965a/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java#L35

joeygrover: Right, but what I'm wondering is, can the security lib be initialized before a session is set? (session would be null, and the callback wouldn't happen at the changed code) If so, there would be no callback telling the session that security is initialized because it missed the only callback. If the security library is being reset, but the security library is still attempting to start, we should cancel the initialization process of the security library.

mrapitis: A session needs to be established prior to the initialization (we need an RAI response), please see the code snippet below:
https://github.com/smartdevicelink/sdl_android/blob/41a6698a2cbd7b3428fc3be55efe30fc21c3965a/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L1781-L1823

joeygrover: So really this should return immediately if the session is null as it is orphaned and will not be used.

mrapitis: True good point, I can update the PR to return immediately in that case.

PR #449, opened 2017-04-04
Bugfix/issue 443

brettywhite:
## Issue:

Buffer methods called even if buffer was null.

## Fix:

Don't invoke buffer methods if null by surrounding calls in an else statement insuring the buffer is indeed not null

Also added the unit test file RegisteredAppTests that tests is the buffer is null or not. Before, the app would crash with a NPE and now it will not.

### DIFF
--- a/sdl_android/src/androidTest/java/com/smartdevicelink/transport/RegisteredAppTests.java
+++ b/sdl_android/src/androidTest/java/com/smartdevicelink/transport/RegisteredAppTests.java
@@ -1,0 +1,38 @@
+package com.smartdevicelink.transport;
+
+import android.os.Handler;
+import android.os.Looper;
+import android.os.Messenger;
+import android.test.AndroidTestCase;
+
+
+/**
+ * Created by brett on 4/4/17.
+ */
+
+public class RegisteredAppTests extends AndroidTestCase {
+
+    private static final String APP_ID = "123451123";
+
+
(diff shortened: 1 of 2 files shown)


sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -2265,19 +2265,20 @@ public void handleMessage(int flags, byte[] packet){
 			}
 			if(buffer == null){
 				Log.e(TAG, "Unable to assemble message as buffer was null/not started");
```
joeygrover: I don't think we really need this log message. For the sake of clean code it should be removed and the this made into a single if statement, or null check if with a return

sdl_android/src/androidTest/java/com/smartdevicelink/transport/RegisteredAppTests.java
```diff
@@ -0,0 +1,40 @@
...
+
+                // Call Handle Message
+                app.handleMessage(TransportConstants.BYTES_TO_SEND_FLAG_LARGE_PACKET_START,bytes);
+
+                // Insure that the buffer is not null, if it is the test will fail
+                assertNotNull(app.buffer);
+
+            }
+        });
+    }
```
brettywhite: see latest commit

Issue #473, opened 2017-04-27
Update buffer read size for encrypted payloads to match max tls record size


mrapitis:
Currently the android library reads a maximum of 1024 bytes (see link below) when an encrypted payload is utilized for video and audio services -- this value needs to be updated to reflect the maximum value available for a TLS record (16384 bytes) minus any overhead used by the record such as the tls record header (5 bytes), tls record message authentication code (32 bytes), and the maximum tls record padding (up to 256 bytes) used.

https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/streaming/StreamPacketizer.java#L17

PR #474, opened 2017-04-27
Update buffer read size for encrypted payloads to match TLS record standard

mrapitis:
Fixes issue #473

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/streaming/StreamPacketizer.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/streaming/StreamPacketizer.java
@@ -14,7 +14,14 @@ public class StreamPacketizer extends AbstractPacketizer implements Runnable{
 
 	private Thread t = null;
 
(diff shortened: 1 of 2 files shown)


sdl_android/src/main/java/com/smartdevicelink/streaming/StreamPacketizer.java
```diff
@@ -14,7 +14,14 @@
 
 	private Thread t = null;
 
-	private final static int BUFF_READ_SIZE = 1024;
+
+	private final static int TLS_MAX_RECORD_SIZE = 16384;
+	private final static int TLS_RECORD_HEADER_SIZE = 5;
```
askirk: Is this supposed to be 5 or 4? You said 4 in your issue

mrapitis: @askirk should be 5, looks like I had a type in the issue.  I updated the issue with the proper value. (I also added an issue to the protocol spec with more details) Thanks!

PR #475, opened 2017-04-27
Working gradle script that deploys to Bintray

askirk:
Fill out user and api_key fields. Change version.Name to the correct version. Use `./gradlew bintrayUpload` to upload

### DIFF
--- a/sdl_android/bintray.gradle
+++ b/sdl_android/bintray.gradle
@@ -45,24 +45,27 @@ artifacts {
 }
 
 bintray {
+    Properties props = new Properties()
+    props.load(new FileInputStream("$projectDir/bintray.properties"))
+
     // Authorization
-    user = ''
-    key = ''
+    user = props.getProperty("bintray.user")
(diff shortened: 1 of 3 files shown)


askirk:
After upload you can compile with

`compile 'com.smartdevicelink:sdl_android:{version}'`

sdl_android/bintray.gradle
```diff
@@ -0,0 +1,68 @@
...
+    pkg {
+        repo = 'sdl_android'
+        name = "sdl_android"
+        websiteUrl = siteUrl
+        vcsUrl = gitUrl
+        userOrg = 'smartdevicelink'
+        licenses = ["BSD 3-Clause"]
+        publish = true
+        version {
+            name = '4.3.0'  // Change to release version
```
joeygrover: Is it possible to have this in a separate config? That way we can access it in the code.

askirk: I was trying to do that but it was having trouble finding the config file, I'll look into it more though

Issue #458, opened 2017-04-12
[SDL 0049] Gesture cancellation on video streaming


theresalech:
Proposal: [Gesture cancellation on video streaming](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0049-touch-cancellation.md)

> This proposal defines a gesture cancellation event to prevent false gesture recognition during an interrupted video stream.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/144

Steering Committee Decision:

> The Steering Committee has agreed to accept this proposal, citing the drag and drop example as a valid use case.

PR #477, opened 2017-05-02
Add enum to OnTouchEvent for cancellation gesture

askirk:
In reference to issue #458


joeygrover:
This PR is acceptable as is, however, once we introduce the new VirtualDisplayEncoder we will need to be sure to actually handle this new enum when converting touch events.

PR #479, opened 2017-05-03
Fix TravisCI failures

askirk:
Will require this PR to test


askirk:
The unit tests now run, each TravisCI build takes between 18-22 mins.

PR #482, opened 2017-05-04
bugfix/utest_lockScreenManagerTest

brettywhite:
- test was actually good, there was a NPE in the lock screen manager class. Fixed that

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/proxy/LockScreenManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/proxy/LockScreenManager.java
@@ -41,11 +41,15 @@ public class LockScreenManager {
 	public synchronized void setHMILevel(HMILevel hmiVal)
 	{		
 		hmiLevel = hmiVal;
-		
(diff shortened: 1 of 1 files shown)


sdl_android/src/main/java/com/smartdevicelink/proxy/LockScreenManager.java
```diff
@@ -41,11 +41,15 @@ public synchronized void setDriverDistStatus(boolean bVal)
...
-		else if (hmiVal.equals(HMILevel.HMI_NONE))
-			setUserSelectedStatus(false);				
+
+		if (hmiVal != null) {
+			if ((hmiVal.equals(HMILevel.HMI_FULL)) || (hmiVal.equals(HMILevel.HMI_LIMITED)))
+				setUserSelectedStatus(true);
+			else if (hmiVal.equals(HMILevel.HMI_NONE))
+				setUserSelectedStatus(false);
+		}else{
+			setUserSelectedStatus(false);
```
askirk: Should we be setting this on a null input or just do nothing?

PR #487, opened 2017-05-04
bugfix/utest_abstractPacketizerTests

brettywhite:
- fixed NPE in the second constructor 
- also added the check in the first constructor as well


sdl_android/src/main/java/com/smartdevicelink/streaming/AbstractPacketizer.java
```diff
@@ -30,8 +30,10 @@ public AbstractPacketizer(IStreamListener streamListener, InputStream is, Sessio
 		_rpcSessionID = rpcSessionID;
 		_serviceType = sType;
 		this._session = session;
-		bufferSize = this._session.getMtu();
-		buffer = new byte[bufferSize];
+		if (this._session != null) {
```
joeygrover: This is going to push the problem to all child classes of this abstract class. See [here](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/streaming/StreamPacketizer.java#L28). Maybe it's better to fail hard with an initializer exception?

PR #488, opened 2017-05-04
bugfix/utest_diagLevelTests

brettywhite:
- adding null pointer checks as the tests were failing when null

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/trace/DiagLevel.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/trace/DiagLevel.java
@@ -14,29 +14,38 @@ public class DiagLevel {
 	}
 	
 	public static void setAllLevels(DetailLevel thisDetail) {
-		for (int i=0; i < levels.length; i++) {
-			levels[i] = thisDetail; //  
+		if (thisDetail != null) {
+			for (int i = 0; i < levels.length; i++) {
+				levels[i] = thisDetail; //
+			}
 		}
 	}
 	
     public static void setLevel(Mod thisMod, DetailLevel thisDetail) {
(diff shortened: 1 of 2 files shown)


sdl_android/src/main/java/com/smartdevicelink/trace/DiagLevel.java
```diff
@@ -14,29 +14,38 @@
...
+			for (int i = 0; i < levels.length; i++) {
+				levels[i] = thisDetail; //
+			}
 		}
 	}
 	
     public static void setLevel(Mod thisMod, DetailLevel thisDetail) {
-    	levels[thisMod.ordinal()] = thisDetail;
+		if (thisMod != null && thisDetail != null) {
+			levels[thisMod.ordinal()] = thisDetail;
```
askirk: Should there be another check for `.ordinal != null`?

sdl_android/src/main/java/com/smartdevicelink/trace/DiagLevel.java
```diff
@@ -14,29 +14,38 @@
...
+		}
     }
 	
 	public static DetailLevel getLevel(Mod thisMod) {
-		return levels[thisMod.ordinal()];
+		if (thisMod != null) {
+			return levels[thisMod.ordinal()];
```
askirk: Same thing

sdl_android/src/main/java/com/smartdevicelink/trace/DiagLevel.java
joeygrover: If mod is not null, it is guaranteed to have an ordinal value.

PR #492, opened 2017-05-04
bugfix/utest_Mime

askirk:


sdl_android/src/main/java/com/smartdevicelink/trace/Mime.java
```diff
@@ -18,48 +22,61 @@ public static String base64Encode(String str) {
...
-				break;
-			case 2:
-				b64idx = ((bytesToEncode[idxin-1] << 2)& 0x3c);
-				sb.append(m_base64Chars.charAt(b64idx));
-				sb.append('=');
-				break;
-		}	
-		return sb.toString();
+			return sb.toString();
+		}catch (ArrayIndexOutOfBoundsException e){
```
joeygrover: This seems like a pretty easy exception to prevent, we should probably try that rather than just catch it.

sdl_android/src/main/java/com/smartdevicelink/trace/Mime.java
```diff
@@ -7,6 +7,10 @@
 	private static String m_base64Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 	
 	public static String base64Encode(String str) {
+		if(str == null){
```
joeygrover: Needs documentation explaining what is returned, and if null is provided, null will be returned.

sdl_android/src/main/java/com/smartdevicelink/trace/Mime.java
```diff
@@ -18,48 +22,61 @@ public static String base64Encode(String str) {
 	}
 
 	public static String base64Encode(byte bytesToEncode[]) {
-		return base64Encode(bytesToEncode, 0, bytesToEncode.length);
+		if(bytesToEncode != null){
+			return base64Encode(bytesToEncode, 0, bytesToEncode.length);
+		}
+		return null;
```
joeygrover: Needs documentation explaining what is returned, and if null is provided, null will be returned.

sdl_android/src/main/java/com/smartdevicelink/trace/Mime.java
```diff
@@ -18,48 +22,61 @@ public static String base64Encode(String str) {
...
-		return base64Encode(bytesToEncode, 0, bytesToEncode.length);
+		if(bytesToEncode != null){
+			return base64Encode(bytesToEncode, 0, bytesToEncode.length);
+		}
+		return null;
 	}
 
 	public static String base64Encode(byte bytesToEncode[], int offset, int length) {
+		if(bytesToEncode == null){
+			return null;
```
joeygrover: Needs documentation explaining what is returned, and if null is provided, null will be returned.

PR #493, opened 2017-05-04
bugfix/utest_MultiplexBluetoothTransport

askirk:
- Was previously failing due to need to call Looper.prepare()
- The start() call and subsequent check for changed state failed because the BT adapter is not enabled, there are no bluetooth permissions in this library (TODO: find a way to test around that)


sdl_android/src/androidTest/java/com/smartdevicelink/test/transport/MultiplexBluetoothTransportTest.java
```diff
@@ -17,41 +18,46 @@
 	boolean didCorrectThing = false, isWaitingForResponse = false;
```
brettywhite: didCorrectThing is sort of ambiguous

PR #494, opened 2017-05-04
bugfix/utest_BitConverter

askirk:
Now returns -1 or null object upon null input array


sdl_android/src/main/java/com/smartdevicelink/util/BitConverter.java
```diff
@@ -1,10 +1,19 @@
 package com.smartdevicelink.util;
 
 public class BitConverter {
+	/**
+	 * @param bytes passing in a null byte array returns a null object
```
joeygrover: The description of what's returned should be in the "return" tag Something like:
@param bytes byte array that will be converted to hex
@return the String containing converted hex values or null if byte array is null

Issue #463, opened 2017-04-18
Potential NPE, SdlRouterService.java, in writeBytesToTransport()


Tnnnguyen:
Potential NPE: int count = ... should be moved inside if (packet != null) clause
```
			if(mSerialService !=null && mSerialService.getState()==MultiplexBluetoothTransport.STATE_CONNECTED){
				byte[] packet = bundle.getByteArray(TransportConstants.BYTES_TO_SEND_EXTRA_NAME); 
				int offset = bundle.getInt(TransportConstants.BYTES_TO_SEND_EXTRA_OFFSET, 0); //If nothing, start at the begining of the array
				int count = bundle.getInt(TransportConstants.BYTES_TO_SEND_EXTRA_COUNT, packet.length);  //In case there isn't anything just send the whole packet.
				if(packet!=null){
					mSerialService.write(packet,offset,count);
					return true;
				}
				return false;
			}
```
[Reference](https://github.com/smartdevicelink/sdl_android/blob/303bee02221f9e8360a5a4c449c6dc501518166d/sdl_android_lib/src/com/smartdevicelink/transport/SdlRouterService.java#L1252-L1257)

PR #495, opened 2017-05-04
#463, fix a potential NPE, move packet.length inside null check.

Tnnnguyen:
Fix issue [#463](https://github.com/smartdevicelink/sdl_android/issues/463)


sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1291,8 +1291,8 @@ public boolean writeBytesToTransport(Bundle bundle){
 			if(mSerialService !=null && mSerialService.getState()==MultiplexBluetoothTransport.STATE_CONNECTED){
 				byte[] packet = bundle.getByteArray(TransportConstants.BYTES_TO_SEND_EXTRA_NAME); 
 				int offset = bundle.getInt(TransportConstants.BYTES_TO_SEND_EXTRA_OFFSET, 0); //If nothing, start at the begining of the array
```
joeygrover: We should probabaly move this line under the `packet!=null` check as well since it doesn't make sense to allocate this if the packet is null.

joeygrover:
Code looks good, can you add some unit tests for this method?

PR #496, opened 2017-05-04
bugfix/utest_OnSdlChoiceChosen

askirk:


sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/OnSdlChoiceChosen.java
```diff
@@ -85,42 +85,12 @@ public String toString() {
 		}
 	}
 	
-	public class SdlChoice {
```
joeygrover: We can't just remove this as it is a public class. Which would require a major version change of the library.

sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/OnSdlChoiceChosen.java
```diff
@@ -58,115 +62,125 @@ public String toString() {
...
+
 		public String toString() {
 			return _menuName;
 		}
 	}
-	
-	public class SdlChoice {
-		
+
+	public class SdlChoice extends RPCStruct {
```
joeygrover: We shouldn't change source code just to make unit tests work. We should find a way around not changing it, or possibly removing the current tests and create tests that match this class better.

askirk:
Creating a new PR for this

PR #498, opened 2017-05-04
bugfix/utest_RegisteredAppTest

brettywhite:
- removed main thread calls 
- made sure looper was prepared
- cleaned up double indent

### DIFF
--- a/sdl_android/src/androidTest/java/com/smartdevicelink/transport/RegisteredAppTests.java
+++ b/sdl_android/src/androidTest/java/com/smartdevicelink/transport/RegisteredAppTests.java
@@ -18,47 +18,44 @@ public class RegisteredAppTests extends AndroidTestCase {
 
     public void testHandleMessage() {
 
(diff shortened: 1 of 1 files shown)


sdl_android/src/androidTest/java/com/smartdevicelink/transport/RegisteredAppTests.java
```diff
@@ -18,47 +18,44 @@
 
     public void testHandleMessage() {
 
-        // Run Test in Main Thread
-        new Handler(Looper.getMainLooper()).post(new Runnable() {
-            @Override
-            public void run() {
+        if (Looper.myLooper() == null)
+        {
```
joeygrover: Try to stick to the style of the project and leave the bracket on the same line as the if statement

brettywhite: done

Issue #470, opened 2017-04-21
Suboptimal Behavior when Force Closing App Running Router Service


askirk:
__Current Problem__
Take a situation where three apps A, B, and C are connected to the head unit via Multiplex BT transport, and all through app A's Router Service. When app A is force closed, all apps disconnect from the head unit. Then app B and C reconnect, resorting to using the Legacy BT transport before one of their router service's can start up. 

__Ideal Behavior__
Upon force closing app A, a router service from app B or C should start up. B and C should then use that router service to reconnect to the head unit, still using the Multiplex BT transport. 

__Stacktrace__
Here's a stacktrace from an example I've run across:
```
/* App A is Force Closed and Its Router Service is Shut Down. */
/com.app.B E/SdlProxy: VERSION-INFO: Transport failure: 
/com.app.B.r D/RSVP: Supplied service name of com.app.A.SdlRouterService
/com.app.C W/SdlConnection: SDL Router service isn't trusted. Enabling legacy bluetooth connection.
/com.app.B W/SdlConnection: SDL Router service isn't trusted. Enabling legacy bluetooth connection.
/? I/ActivityManager: Start proc 20127:com.smartdevicelink.router/u0a218 for service com.app.C/.SdlRouterService
/? W/Sdl Router Service: Supplied intent was null, local router service will not contain intent
/? I/Sdl Router Service: SDL Router Service has been created
```

tpulatha:
This would probably need a proposal for behavior change, right?

joeygrover:
No, it should work in the optimal way described, but the current behavior doesn't match.

askirk:
This is where the apps are deciding to fall back on legacy BT:

https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlConnection.java#L97-L106

A possible solution would be to check for any other router services before falling back and enabling legacy BT.

askirk:
/

PR #504, opened 2017-05-10
Bugfix/issue 470 - Suboptimal Multiplexing Behavior

askirk:
In reference to issue #470 ...

Fixing the case where the apps using a router service fall back to legacy BT when that router service is force closed. Now we check if there are additional Sdl Broadcast Receivers (and Router Services) that can be woken before falling back to legacy BT.


askirk:
I'm not sure if the `isAdditionalSdlBroadcastReceiver()` method should check that a found BroadcastReceiver is verified using a method like `isTrustedPackage()`. Currently it does not.

brettywhite:
I'd assume it should 🔎

sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlConnection.java
```diff
@@ -98,12 +98,17 @@ private void constructor(BaseTransportConfig transportConfig,RouterServiceValida
...
 						cachedMultiConfig = (MultiplexTransportConfig) transportConfig;
 						cachedMultiConfig.setService(null);
 					}
-					enableLegacyMode(true,TransportType.BLUETOOTH); //We will use legacy bluetooth connection for this attempt
+
+					if(!rsvp.isAdditionalSdlBroadcastReceiver()) { // If there are no other Router Services to wake up
+						Log.w(TAG, "SDL Router service isn't trusted. Enabling legacy bluetooth connection.");
+						enableLegacyMode(true, TransportType.BLUETOOTH); //We will use legacy bluetooth connection for this attempt
+					}else{
+						Log.w(TAG, "SDL Router service isn't trusted. Attempting to wake up other router services.");
```
joeygrover: This  comment doesn't make sense here.

sdl_android/src/main/java/com/smartdevicelink/transport/RouterServiceValidator.java
```diff
@@ -513,7 +513,7 @@ public boolean isAdditionalSdlBroadcastReceiver(){
 			String packageName = resolveInfo.activityInfo.packageName;
 			if(packageName != null){
 				if(lastClosedService != null){
-					if(isTrustedPackage(packageName, manager) && !lastClosedService.equals(packageName)){
+					if(isTrustedPackage(packageName, manager) && !lastClosedService.toString().equals(packageName)){
 						return true;
```
Tnnnguyen: This condition check will not work the way it's intent to:
`!lastClosedService.toString().equals(packageName)`

The reason is that `lastClosedService` is a ComponentName and its toString method returns: 
`"ComponentInfo{" + mPackage + "/" + mClass + "}"`
which will never equal to `packageName`.

I think it should be: `!lastClosedService.getPackageName().equals(packageName)` if the intention is to exclude the current (last closed) service.

timeyinxingren:
this problem has been fixed? @BrettyWhite

PR #516, opened 2017-05-18
Fixing the failing unit tests in Emulator

askirk:
Adds a DeviceUtil class with isEmulator() function to detect if running in emulator. Aspects like MultiplexBT and BT adapter checks will account for if an emulator is running the test. Changed FTP url in Packetizer tests that was failing in emulator.


askirk:
Wasn't based on latest develop branch

Issue #464, opened 2017-04-18
Potential NPE, SdlRouterService.java, in handleMessage() of AltTransportHandler


Tnnnguyen:
Potential NPE: It appears that we add the bundle to the message object; but nevertheless, should handle bundle being null here at "receivedBundle.containsKey..."
```
if(receivedBundle!=null){
	        			receivedBundle.setClassLoader(loader);//We do this because loading a custom parceable object isn't possible without it
	            	}else{
	            		Log.e(TAG, "Bundle was null while sending packet to router service from alt transport");
	            	}
            		if(receivedBundle.containsKey(TransportConstants.FORMED_PACKET_EXTRA_NAME)){
            			SdlPacket packet = receivedBundle.getParcelable(TransportConstants.FORMED_PACKET_EXTRA_NAME);
    					if(packet!=null){
    						service.onPacketRead(packet);
    					}else{
    						Log.w(TAG, "Received null packet from alt transport service");
    					}
            		}else{
            			Log.w(TAG, "Flase positive packet reception");
            		}
```
[Reference](https://github.com/smartdevicelink/sdl_android/blob/303bee02221f9e8360a5a4c449c6dc501518166d/sdl_android_lib/src/com/smartdevicelink/transport/SdlRouterService.java#L603-L617)

Although we are almost guaranteed to have a non-null bundle in the Message object; we will attempt to handle the rare case here and put the code that accesses the bundle inside the already existed null check block.

Tnnnguyen:
Completed.

PR #521, opened 2017-05-24
Bugfix/issue 464

Tnnnguyen:
Fix issue [#464](https://github.com/smartdevicelink/sdl_android/issues/464)

### Risk

### Testing Plan
Create a Message object with null data, feed the message into the handler and test if handleMessage method can handle null.
New unit test passed.

### Summary
We will attempt to handle message data being null and prevent a npe crash here.

### DIFF
--- a/sdl_android/src/androidTest/java/com/smartdevicelink/transport/SdlRouterServiceTests.java
+++ b/sdl_android/src/androidTest/java/com/smartdevicelink/transport/SdlRouterServiceTests.java
@@ -1,0 +1,85 @@
+package com.smartdevicelink.transport;
+
+
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.os.Message;
+import android.test.AndroidTestCase;
+import android.util.Log;
+
+import com.smartdevicelink.protocol.SdlPacket;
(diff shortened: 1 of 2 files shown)


sdl_android/src/androidTest/java/com/smartdevicelink/transport/SdlRouterServiceTests.java
```diff
@@ -0,0 +1,85 @@
...
+        super.tearDown();
+        //Nothing here for now
+    }
+
+    /**
+     * Test null bundle handling in AltTransportHandler when handling messages. Only test the case of
+     * msg.what == TransportConstants.ROUTER_RECEIVED_PACKET
+     */
+    public void testAlTransportHandlerHandleNullBundle() {
+        Looper.prepare();
```
brettywhite: you need to check if a looper is already available, otherwise it will error:
```java
        if (Looper.myLooper() == null) {
            Looper.prepare();
        }
```

Tnnnguyen: @brettywhite, thanks.

Issue #461, opened 2017-04-18
Potential NPE, RouterServiceValidator


Tnnnguyen:
#### Potential NPE: dereferencing "apps" in for loop
```
List<SdlApp> apps = findAllSdlApps(context);
		
		final JSONObject object = new JSONObject();
		JSONArray array = new JSONArray();
		JSONObject jsonApp;
		
		for(SdlApp app: apps){	//Format all the apps into a JSON object and add it to the JSON array
			try{
				jsonApp = new JSONObject();
				jsonApp.put(JSON_APP_PACKAGE_TAG, app.packageName);
				jsonApp.put(JSON_APP_VERSION_TAG, app.versionCode);
				array.put(jsonApp);
			}catch(JSONException e){
				e.printStackTrace();
				continue;
			}
		}
```
[Reference](https://github.com/smartdevicelink/sdl_android/blob/303bee02221f9e8360a5a4c449c6dc501518166d/sdl_android_lib/src/com/smartdevicelink/transport/RouterServiceValidator.java#L380-L395)

Although we are almost guaranteed to receive a non-null and non-empty list when calling
```
List<SdlApp> apps = findAllSdlApps(context);
```
we will attempt to handle the rare case here. 
1. When receiving a null list, we will return immediately
2. When array is empty, either due to a null list or an empty list, we will also return immediately as the end point to request the trusted list expects at least one app.

PR #524, opened 2017-05-24
Bugfix/issue_461

Tnnnguyen:
Fix issue [#461](https://github.com/smartdevicelink/sdl_android/issues/461)

### Risk

### Testing Plan
Extend RouterServiceValidator.java, overriding findAllSdlApps method and return null for testing then call findAllSdlApps and handle null value.
New unit tests passed.

### Summary
Although it's almost guaranteed to have findAllSdlApps to return a non null list.  We will attempt to handle the null case and prevent a npe crash here.

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/RouterServiceValidator.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/RouterServiceValidator.java
@@ -429,11 +429,6 @@ public class RouterServiceValidator {
 					continue;
 				}
 			}
-		} else {	//Return here and do not bother to make request since there's no app to send
-			if (listCallback != null) {
-				listCallback.onListObtained(true);
-			}
-			return false;
 		}
 		
 		try {object.put(JSON_PUT_ARRAY_TAG, array);} catch (JSONException e) {e.printStackTrace();}
@@ -475,7 +470,7 @@ public class RouterServiceValidator {
 		}
 
 		new HttpRequestTask(cb).execute(REQUEST_PREFIX,HttpRequestTask.REQUEST_TYPE_POST,object.toString(),"application/json","application/json");
-		
+
 		return true;
 	}
 	
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/RouterServiceValidator.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/RouterServiceValidator.java
@@ -351,7 +351,7 @@ public class RouterServiceValidator {
 		List<SdlApp> apps = new ArrayList<SdlApp>();
 		PackageManager packageManager = context.getPackageManager();
 		Intent intent = new Intent();
-		intent.setAction("sdl.router.startservice");
+		intent.setAction(TransportConstants.START_ROUTER_SERVICE_ACTION);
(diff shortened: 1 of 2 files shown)


Tnnnguyen:
I am looking into this.

sdl_android/src/main/java/com/smartdevicelink/transport/RouterServiceValidator.java
```diff
@@ -345,22 +345,22 @@ protected boolean verifyVersion(int version, JSONArray versions){
...
 	 * intent. 
-	 * @return 
+	 * @return the list of Sdl enabled apps if any app is found; null or empty list otherwise
 	 */
-	private static List<SdlApp> findAllSdlApps(Context context){
+	protected static List<SdlApp> findAllSdlApps(Context context){
 		List<SdlApp> apps = new ArrayList<SdlApp>();
 		PackageManager packageManager = context.getPackageManager();
 		Intent intent = new Intent();
 		intent.setAction("sdl.router.startservice");
```
joeygrover: Since there are already changes here I think we should change this from the hardcoded string

Tnnnguyen: Just made the change, thnx.

sdl_android/src/main/java/com/smartdevicelink/transport/RouterServiceValidator.java
```diff
@@ -417,16 +417,17 @@ else if(getSecurityLevel(context) == MultiplexTransportConfig.FLAG_MULTI_SECURIT
...
-		for(SdlApp app: apps){	//Format all the apps into a JSON object and add it to the JSON array
-			try{
-				jsonApp = new JSONObject();
-				jsonApp.put(JSON_APP_PACKAGE_TAG, app.packageName);
-				jsonApp.put(JSON_APP_VERSION_TAG, app.versionCode);
-				array.put(jsonApp);
-			}catch(JSONException e){
-				e.printStackTrace();
-				continue;
+		if (apps != null && apps.size() > 0) {
```
joeygrover: This list should theoretically never be null or size = 0. However, if that is the case I don't believe we should even perform the Http request. We should  add `if(listCallback!=null){listCallback.onListObtained(false);}` and return out of this method with false.

Tnnnguyen: I made this change, thnx.

Tnnnguyen:
The list of Sdl list can be returned as empty; when it's empty, we return false early when trying to create a trusted list. This will fail or cause the flow to freeze for some of the current test codes because they are asserting true (list empty or not). Hence we are checking to see if the list of Sdl apps is non-null and not empty on setup. Only if it is, continue to test as we do previously; otherwise, ignore the now invalid test code.

joeygrover:
Can you pull in develop again to resolve conflict?

joeygrover:
@Tnnnguyen can you pull in develop to resolve conflicts? If not I will be closing the pull request.

Tnnnguyen:
I will pull in and resolve conflicts today or tomorrow.

Issue #467, opened 2017-04-18
Potential NPE, WiProProtocol.java


Tnnnguyen:
Dereferencing "data" may potentially cause a NPE

```
synchronized(messageLock) {
			if (data.length > MAX_DATA_SIZE) {
				
				messageID++;
```

[Reference](https://github.com/smartdevicelink/sdl_android/blob/303bee02221f9e8360a5a4c449c6dc501518166d/sdl_android_lib/src/com/smartdevicelink/protocol/WiProProtocol.java#L163-L166)

PR #531, opened 2017-05-31
Bugfix/issue 467

Tnnnguyen:
Fixes [issue #467](https://github.com/smartdevicelink/sdl_android/issues/467)

### Risk

### Testing Plan
Send a null ProtocolMessage to WiProtocol's SendMessage and test if it can handle.

### Summary
When a message to be sent is null, we should just not attempt to send it and not crash.


sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -172,57 +172,59 @@ else if (protocolMsg.getBulkData() != null) {
...
-	                            ++frameSequenceNumber;
-	                        }
-					} else {
-						frameSequenceNumber = SdlPacket.FRAME_INFO_FINAL_CONNESCUTIVE_FRAME;
-					} // end-if
-					
-					int bytesToWrite = data.length - currentOffset;
-					if (bytesToWrite > MAX_DATA_SIZE) { 
-						bytesToWrite = MAX_DATA_SIZE; 
+			if (data != null) {
```
joeygrover: I'm not sure this is how we want to handle potential null data. If a frame doesn't have any data this logic would prevent it from being sent. I believe we have to incorporate more intelligent logic to handle that case and ensure a packet is still sent.

Tnnnguyen: That was the intention of the fix, don't send anything if data is null. Clarifying question: what is the reason for the need to send the packet anyway if data is null? Is it because we need to keep track of the `messageId` field? Is there any other reason? Thanks.

joeygrover: I don't see any clarifying language that that's the intended outcome. So if that is the case, we should at least add documentation to the method that will inform users that if this method is called with a message that has no data, it will not be sent.

Tnnnguyen: I see. So there are 2 ways we can handle this now.
1. When data is null, don't send anything (fix stays the same) and add documentation to the callers of the method.
2. When data is null, do send the packet with null data like so (similar to the case where data.length < MAX_DATA_SIZE):
`SdlPacket header = SdlPacketFactory.createSingleSendData(sessionType, sessionID, data.length, messageID, _version, data, protocolMsg.getPayloadProtected());`
and replace:
a. `data.length` with either MAX_DATA_SIZE or negative.
b. data = data (null)
Internally when the code tries to create a packet in `SdlPacket.java`, `constructPacket()` method, the null data is not added to the ByteBuffer due to this check:
`if(payload!=null && payload.length>0){
			builder.put(payload);
		}
`
If there is no imperative need to send the packet when data is null, I would vote for 1.

joeygrover: Yea I think option 1 is fine due to the fact that without this PR if data was null it would just crash and not send the packet, so it's not changing behavior just preventing a crash.

PR #543, opened 2017-06-19
Adding FILE enum to SpeechCapabilities

askirk:
- And added associate line in unit test
#419


brettywhite:
tested w/ core, works well

PR #549, opened 2017-06-30
Resend PR about enhance video streaming performance

BrandonHe:
Hi @joeygrover 
Sorry resend this PR again, since we have done something and we'd like you to review again:
- Have reset the commits in #446 
- Rewrite the code in custom class `SdlPipedInputStream` and `SdlPipedOutStream` based on default sdk
- I think it's now under **Apache 2.0 license.**

**Test**:
PC: Debian 32-bit, Windows 32-bit
Phone: Android4.4(API_LEVEL 19)
Connection: USB_AOA
**Result:**
The data in test are false video stream data, aims to get the max value in transmission.

|             | Core(Windows)|
|:---------|:------:|
|Proxy (**NOT** include this PR) | 1.54MB/s  |
|Proxy (**include** this PR)         | 9.11MB/s  |

Could you please review again and give your feedback?

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlSession.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlSession.java
@@ -1,14 +1,5 @@
 package com.smartdevicelink.SdlConnection;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.PipedInputStream;
-import java.io.PipedOutputStream;
-import java.util.List;
-import java.util.ListIterator;
-import java.util.concurrent.CopyOnWriteArrayList;
-
 import android.annotation.SuppressLint;
 import android.os.Build;
 import android.util.Log;
@@ -25,11 +16,20 @@ import com.smartdevicelink.proxy.RPCRequest;
 import com.smartdevicelink.security.ISecurityInitializedListener;
 import com.smartdevicelink.security.SdlSecurityBase;
 import com.smartdevicelink.streaming.IStreamListener;
+import com.smartdevicelink.streaming.SdlPipedInputStream;
+import com.smartdevicelink.streaming.SdlPipedOutputStream;
 import com.smartdevicelink.streaming.StreamPacketizer;
 import com.smartdevicelink.streaming.StreamRPCPacketizer;
 import com.smartdevicelink.transport.BaseTransportConfig;
 import com.smartdevicelink.transport.MultiplexTransport;
 import com.smartdevicelink.transport.enums.TransportType;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 public class SdlSession implements ISdlConnectionListener, IHeartbeatMonitorListener, IStreamListener, ISecurityInitializedListener {
 	private static CopyOnWriteArrayList<SdlConnection> shareConnections = new CopyOnWriteArrayList<SdlConnection>();
@@ -148,12 +148,12 @@ public class SdlSession implements ISdlConnectionListener, IHeartbeatMonitorList
 
 	@SuppressLint("NewApi")
 	public OutputStream startStream(SessionType sType, byte rpcSessionID) throws IOException {
-		OutputStream os = new PipedOutputStream();
+		OutputStream os = new SdlPipedOutputStream();
 		InputStream is = null;
 		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.GINGERBREAD) {
-			is = new PipedInputStream((PipedOutputStream) os, BUFF_READ_SIZE);
+			is = new SdlPipedInputStream((SdlPipedOutputStream) os, BUFF_READ_SIZE);
 		} else {
-			is = new PipedInputStream((PipedOutputStream) os);
+			is = new SdlPipedInputStream((SdlPipedOutputStream) os);
 		}
         if (sType.equals(SessionType.NAV))
         {
@@ -187,8 +187,8 @@ public class SdlSession implements ISdlConnectionListener, IHeartbeatMonitorList
 
 	public OutputStream startRPCStream(RPCRequest request, SessionType sType, byte rpcSessionID, byte wiproVersion) {
 		try {
-			OutputStream os = new PipedOutputStream();
-	        InputStream is = new PipedInputStream((PipedOutputStream) os);
+			OutputStream os = new SdlPipedOutputStream();
+	        InputStream is = new SdlPipedInputStream((SdlPipedOutputStream) os);
 			mRPCPacketizer = new StreamRPCPacketizer(null, this, is, request, sType, rpcSessionID, wiproVersion, 0, this);
 			mRPCPacketizer.start();
 			return os;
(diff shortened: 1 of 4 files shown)


BrandonHe:
@joeygrover 
Could you please take a look at this PR? This is the alternative solution of #446.
Also, I have no permission to request a reviewer, could you give me collaborator permission?

joeygrover:
1. If you copy and pasted this code you need to state very clearly where it came from. There are important legal issues to consider when doing that.
2. We can't introduce new licenses into the project. 
3. If this is copy/pasted from the Android source have you tried just setting the compile version to the Android version that implemented the changes to the piped streams? Does it work the same?

BrandonHe:
@joeygrover 
> 1. If you copy and pasted this code you need to state very clearly where it came from. There are important legal issues to consider when doing that.

Yes, I understand there are important legal issues need to be considered, thanks for your reminding. This code mostly come from android sdk 19, as mentioned previous, rewrite function `write()` and `receive()`

> 2. We can't introduce new licenses into the project.

I am not very understand this, from my view point, it's very common to use **proper** open source licenses from third party libraries or source codes if it's no conflicts with current license we are using. For example in [sdl_core](http://colah.github.io/posts/2015-08-Understanding-LSTMs/). If I am wrong here, just correct me.

> 3. If this is copy/pasted from the Android source have you tried just setting the compile version to the Android version that implemented the changes to the piped streams? Does it work the same?

Yes, it works. we have tested more than once.

BTW, our partners had trial testing with the patches in this PR and they said it has a fantastic performance improvement. But before official deliver to them, I think it's better to ask your professional review in case of  legal issues or others may concerned.

joeygrover:
>This code mostly come from android sdk 19, as mentioned previous, rewrite function write() and receive()

Does that mean you copy and pasted the piped stream classes out of Android and then rewrote the write() and receive() methods? If so, did you obtain the code for those write() and read() methods from anywhere?

> I am not very understand this, from my view point, it's very common to use proper open source licenses from third party libraries or source codes if it's no conflicts with current license we are using. For example in sdl_core. If I am wrong here, just correct me.

The sdl_core and other projects have had any previous licenses grandfathered in as they have been properly reviewed. However, contributing new code under any other license other than BSD3 is specifically against section D in the [contributors agreement](https://docs.google.com/a/livio.io/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform) that is required to be signed before we can accept any code. 


> > If this is copy/pasted from the Android source have you tried just setting the compile version to the Android version that implemented the changes to the piped streams? Does it work the same?

> Yes, it works. we have tested more than once.

So simply upping the compile version and not introducing new code solves the issue? If that's the case why would we introduce new code?

BrandonHe:
> Does that mean you copy and pasted the piped stream classes out of Android and then rewrote the write() and receive() methods? If so, did you obtain the code for those write() and read() methods from anywhere?

Yes, we copy and pasted the piped stream classes out of Android and then rewrote the write() and receive() method, these two functions were not obtained from anywhere but Android, we have checked again. Also, we added `notifyAll()` in the receive() function, which will help to read data as soon as the pipe finished receiving. 

> The sdl_core and other projects have had any previous licenses grandfathered in as they have been properly reviewed. However, contributing new code under any other license other than BSD3 is specifically against section D in the contributors agreement that is required to be signed before we can accept any code.

Thanks for your explanation, do we also need to sign the contributors agreement? The code was contributed from our Ford supplier.

> So simply upping the compile version and not introducing new code solves the issue? If that's the case why would we introduce new code?

Sorry, I misunderstood this, we did testing for this PR **NOT** new compile version, we don't have latest Google phone that support  this version. As I mentioned in [proposal](https://github.com/smartdevicelink/sdl_evolution/issues/200), there are only about 4.7% developers are using the compile version 23, for compile version 24 which support pipe stream, I think the quality are less than compile version 23.

BrandonHe:
@joeygrover 
Do you have any other feedback?

joeygrover:
@BrandonHe yes you need to sign the CLA before we could accept any donations. 

There are a few different versions to declare in Android, compile, target, and minimum. If all we need to increase is the compile version then I think that is a very easy and doable solution. It is recommended by Android to always compile with the latest SDK version. Unlike target and minimum that impose versions onto developers, compile versions do not. See [this article](https://medium.com/google-developers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd) of explanations for all version types.

Otherwise, we will have to have the SDLC legal team look into the legality/ramifications of including lifted code with a different license in the project.

BrandonHe:
@joeygrover 
I have asked supplier to sign the CLA.

Thanks for your sharing, that's great to know it. Here we **do made some changes** beyond the compile version.

That's make sense that legal team look into the legality/ramifications of including lifted code with a different license in the project.

BrandonHe:
@joeygrover 
We still suggest you to merge this PR. Besides mobile navigation application, we also tested with TDK, they have much performance improvement.

We have tested with newest **compileSdkVersion**, but it can not optimize the video performance during the runtime. According to your link [article](https://medium.com/google-developers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd), it also clearly tell us that **changing your compileSdkVersion does not change runtime behavior**
**minSdkVersion** is the lower bound for the app can be installed on. 
**targetSdkVersion** provides forward compatibility, however Android-24(sdk 19) doesn't provide forward compatibility for `PipedInputStream` and `PipedOutputStream`

Consider the distribution status on the [dashboards](https://developer.android.com/about/dashboards/index.html), 82% android device are running below Android 7.0 OS, higher percentage in China market I think.

Another information for your reference:
After optimizing code from both Core and Android proxy, we successfully tested that the video performance latency is less than 300ms.
#### Test env: 
- Arm Ubuntu 16.04, Core v4.3.2
- Samsung Note5, Android 6.0; Samsung S8, Android 7.0
- Bit rate: 1Mbps, 6Mbps
- Resolution: 800x480, 1080x720
- Frame rate: 30fps
- Test time: 1min, 4hrs,

#### Test result:

| phone | resolution | bit rate | average frame rate | test time | video latency|
|---------|:------------:|:---------:|:-------------:|:-----------:|:---------------:|
| SamsungNote5 | 1080x720 | 6Mbps | 30fps | 1min | 219 |
| SamsungS8 | 1080x720 | 6Mbps | 26fps | 1min | 245 |
| SamsungNote5 | 800x480 | 1Mbps | 26fps | 4hrs | 295 |
| SamsungS8 | 800x480 | 1Mbps | 30fps | 4hrs | 234 |


Thanks,

joeygrover:
@BrandonHe 

I'm actually going to be closing this pull request for several reasons:

1. The original proposal for this feature was rejected through the evolution process. A new proposal was never entered.
2. There is still uncertainty about the legality of bringing in clearly copy/pasted code. Not something that can be decided here, only a new proposal would be able to bring  that to a vote within the SDLC.
3. Video streaming in the proxy has moved away from pipped streams via the [New data transfer interface in Android Proxy for streaming proposal](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0100-new-data-transfer-interface-for-streaming-android.md). This has been implemented and is currently in develop.

I recommend testing with the latest code from develop to see if you are able to obtain similar results. If you still believe this is a necessary change please enter a new proposal into the evolution repo. Thank you!

BrandonHe:
@joeygrover 
OK, we will have a testing with the latest develop branch, and give your feedback if there is.
Thanks,

Issue #506, opened 2017-05-10
[SDL 0052] Constructed Payloads


theresalech:
Proposal: [Constructed Payloads](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0052-constructed-payloads.md)

> This proposal is for the addition of constructed payloads for non-RPC type packets. It will allow for clearer definition of payloads and allow for better parsing between versions. Constructed payloads reduce overhead by avoiding having to send an RPC message as well not mixing different services in the process.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/157

Steering Committee Decision:

> As this proposal has now been revised to include BSON, the Steering Committee has agreed to accept this proposal.

PR #577, opened 2017-08-03
506 Constructed Payloads

askirk:
- See Issue #506


sdl_android/src/main/java/com/smartdevicelink/protocol/SdlPacket.java
```diff
@@ -65,6 +67,7 @@
 	int messageId;
 	int priorityCoefficient;
 	byte[] payload = null;
+	HashMap<String, Object> bsonPayload = new HashMap<>();
```
joeygrover: We shouldn't instantiate this unless it is needed. The majority of packets won't use it so it's best to not allocate the memory.

sdl_android/src/main/java/com/smartdevicelink/protocol/SdlPacket.java
```diff
@@ -186,6 +189,11 @@ public long getDataSize() {
 	}
 	
 	public byte[] constructPacket(){
+		if(!bsonPayload.isEmpty()){
```
joeygrover: After changing how this is instantiated, we will also have to check for null

sdl_android/src/main/java/com/smartdevicelink/protocol/SdlPacket.java
```diff
@@ -344,6 +353,22 @@ public SdlPacket createFromParcel(Parcel in) {
 		}
 
     };
-	
-	
+
+	public void putTag(String tag, Object data){
+		bsonPayload.put(tag, data);
```
joeygrover: This is where you should instantiate the `bsonPayload` object `if(bsonPayload==null){bsonPayload = new HashMap<String,Object>());}`

sdl_android/src/main/java/com/smartdevicelink/protocol/SdlPacket.java
```diff
@@ -344,6 +353,22 @@ public SdlPacket createFromParcel(Parcel in) {
...
+	}
+
+	public Object getTag(String tag){
+		if(payload == null){
+			return null;
+		}else if(bsonPayload.isEmpty()){
```
joeygrover: Again will have to add a null check

sdl_android/src/main/java/com/smartdevicelink/protocol/enums/BsonTags.java
```diff
@@ -0,0 +1,12 @@
+package com.smartdevicelink.protocol.enums;
+
+public class BsonTags {
```
joeygrover: We should make this a little better organized; creating subclasses in this class would be fine. Open to suggestions. Also adding descriptions to the constants.

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -1,30 +1,39 @@
 package com.smartdevicelink.protocol;
 
+
+import android.util.Log;
```
joeygrover: Should remove this import if it's not used.

askirk: was going to use it for printing out rejected Params as error log

joeygrover: I missed that. We should pass those params back. I don't think they are helpful to long unless a developer is debugging. Maybe for now just leave a //TODO and with the modifications for the video streaming params we can pass back the rejected params.

askirk:
@joeygrover updated

PR #579, opened 2017-08-08
535 Add video streaming capabilities - BSON

askirk:
Issue #535

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
@@ -9,6 +9,10 @@ import com.smartdevicelink.protocol.enums.FrameDataControlFrameType;
 import com.smartdevicelink.protocol.enums.FrameType;
 import com.smartdevicelink.protocol.enums.MessageType;
 import com.smartdevicelink.protocol.enums.SessionType;
+import com.smartdevicelink.proxy.rpc.ImageResolution;
+import com.smartdevicelink.proxy.rpc.VideoStreamingFormat;
+import com.smartdevicelink.proxy.rpc.enums.VideoStreamingCodec;
+import com.smartdevicelink.proxy.rpc.enums.VideoStreamingProtocol;
 import com.smartdevicelink.security.SdlSecurityBase;
 import com.smartdevicelink.util.BitConverter;
 import com.smartdevicelink.util.DebugTool;
@@ -468,6 +472,17 @@ public class WiProProtocol extends AbstractProtocol {
 							//At this point we have confirmed the negotiated version between the module and the proxy
 							protocolVersion = new Version((String)version);
 						}
+					}else if(serviceType.equals(SessionType.NAV)){
+						SdlSession session = sdlconn.findSessionById((byte) packet.sessionId);
+						ImageResolution acceptedResolution = new ImageResolution();
+						VideoStreamingFormat acceptedFormat = new VideoStreamingFormat();
+						acceptedResolution.setResolutionHeight((Integer) packet.getTag(ControlFrameTags.Video.StartServiceACK.HEIGHT));
+						acceptedResolution.setResolutionWidth((Integer) packet.getTag(ControlFrameTags.Video.StartServiceACK.WIDTH));
+						acceptedFormat.setCodec((VideoStreamingCodec) packet.getTag(ControlFrameTags.Video.StartServiceACK.VIDEO_CODEC));
+						acceptedFormat.setProtocol((VideoStreamingProtocol) packet.getTag(ControlFrameTags.Video.StartServiceACK.VIDEO_PROTOCOL));
+						if(session != null) {
+							session.setAcceptedVideoConfig(acceptedResolution, acceptedFormat);
+						}
 					}
 				}else{
 					if (protocolVersion.getMajor() > 1){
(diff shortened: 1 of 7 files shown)


sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -575,7 +590,21 @@ private void handleSingleFrameMessageFrame(SdlPacket packet) {
 	public void StartProtocolService(SessionType sessionType, byte sessionID, boolean isEncrypted) {
 		SdlPacket header = SdlPacketFactory.createStartSession(sessionType, 0x00, getMajorVersionByte(), sessionID, isEncrypted);
 		if(sessionType.equals(SessionType.NAV)){
-			// TODO: Add bson tags for video service
+			SdlSession videoSession = sdlconn.findSessionById(sessionID);
+			ImageResolution desiredResolution = null;
+			VideoStreamingFormat desiredFormat = null;
```
joeygrover: Instead of using these params, why don't we use a single class like the one created for a different PR, [`StreamingParameters`](https://github.com/smartdevicelink/sdl_android/pull/471/files#diff-df29f04f294a1bcce443240922d7bb69R61)

sdl_android/src/main/java/com/smartdevicelink/proxy/interfaces/ISdlServiceListener.java
```diff
@@ -0,0 +1,11 @@
+package com.smartdevicelink.proxy.interfaces;
+
+
+import com.smartdevicelink.SdlConnection.SdlSession;
+import com.smartdevicelink.protocol.enums.SessionType;
+
+public interface ISdlServiceListener {
```
joeygrover: We should add a HashMap param to each signature for the potential payload

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -3806,19 +3822,31 @@ public void drainEncoder(boolean endOfStream) {
         sdlSession.drainEncoder(endOfStream);
     }
 	
-	private void NavServiceStarted() {
+	private void NavServiceStarted(boolean isEncrypted) {
 		navServiceStartResponseReceived = true;
 		navServiceStartResponse = true;
+
+		if(sdlSession.getServiceListeners() != null && sdlSession.getServiceListeners().containsKey(SessionType.NAV)){
```
joeygrover: This logic for callbacks should all be in the SdlSession class. Just use the logic in methods like [`onProtocolSessionStarted`](https://github.com/smartdevicelink/sdl_android/blob/d8d92bc44399b5a552fa3211e0d9928ab3d07454/sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlSession.java#L488)

sdl_android/src/main/java/com/smartdevicelink/streaming/VideoStreamingParams.java
```diff
@@ -0,0 +1,90 @@
...
+    public VideoStreamingParams(){
+	    format = new VideoStreamingFormat();
+	    format.setProtocol(DEFAULT_PROTOCOL);
+	    format.setCodec(DEFAULT_CODEC);
+	    resolution = new ImageResolution();
+	    resolution.setResolutionWidth(DEFAULT_WIDTH);
+	    resolution.setResolutionHeight(DEFAULT_HEIGHT);
+    }
+
+    public VideoStreamingParams(ImageResolution resolution, VideoStreamingFormat format){
```
joeygrover: This should include all parameters. Even though this works for the use case of the PR it's not a good design to create constructors for specific use cases.

sdl_android/src/main/java/com/smartdevicelink/streaming/VideoStreamingParams.java
```diff
@@ -0,0 +1,90 @@
...
+	private final int DEFAULT_DENSITY = DisplayMetrics.DENSITY_HIGH;
+	private final int DEFAULT_FRAMERATE = 24;
+	private final int DEFAULT_BITRATE = 512000;
+	private final int DEFAULT_INTERVAL = 5;
+
+
+	private int displayDensity = DEFAULT_DENSITY;
+	private int frameRate = DEFAULT_FRAMERATE;
+	private int bitrate = DEFAULT_BITRATE;
+	private int interval = DEFAULT_INTERVAL;
```
joeygrover: I would move all these default settings to the `VideoStreamingParams()` constructor.

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -468,6 +473,20 @@ else if (frameInfo == FrameDataControlFrameType.StartSession.getValue()) {
 							//At this point we have confirmed the negotiated version between the module and the proxy
 							protocolVersion = new Version((String)version);
 						}
+					}else if(serviceType.equals(SessionType.NAV)){
+						SdlSession session = sdlconn.findSessionById((byte) packet.sessionId);
+						ImageResolution acceptedResolution = new ImageResolution();
```
joeygrover: All of this can be moved under the `if(session!=null)` check because if it's null there's no point in performing any of this logic.

Issue #26, opened 2014-10-23
ChangeRegistrationResponse function name


mikeburke106:
While merging the pull requests today, I noticed that [ChangeRegistrationResponse](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android_lib/src/com/smartdevicelink/proxy/rpc/ChangeRegistrationResponse.java) names the function name "ChangeLanguageRegistration" instead of "ChangeRegistration".

Is this correct, or a typo?  Does SDL core respond to ChangeRegistration request with a function name of "ChangeLanguageRegistration" in the response?

I'm going to assume it's correct and define this as a function name in FunctionID.java, but please let me know if this is not correct.

PR #592, opened 2017-08-23
Feature/issue 515 rtp packetizer

shoamano83:
This PR is related to issue https://github.com/smartdevicelink/sdl_android/issues/515

## Summary

This PR adds RTP packetizer implementation which is necessary for SDL-0048.
Also it includes small additions to encoder's output (IEncoderListener) and ProtocolMessage.setData() which are necessary for RTP output.

## Risk

This PR does not make changes in Mobile_API.
The RTP packetizer is not enabled yet and this PR should not make changes in existing flow.
Therefore I think the risk is low.

## Testing Plan

Unit test cases are added for the new packetizer.

## Note

The RTP implementation is NOT enabled yet. We need video format selection mechanism to enable the new format, and it depends on Video Streaming Capabilities and Constructed Payload implementations.

I created this PR before seeing the implementations on develop branch so that we can work on the implementations and code review in parallel.

By the way, please let me know if I should also work on the switching mechanism after implementations of Video Streaming Capabilities and Constructed Payload are done.

Thanks.


shoamano83:
@joeygrover Please kindly review.

sdl_android/src/main/java/com/smartdevicelink/encoder/IEncoderListener.java
```diff
@@ -0,0 +1,72 @@
...
+package com.smartdevicelink.encoder;
+
+/**
+ * A listener that receives a chunk of data from an encoder.
+ */
+public interface IEncoderListener {
+	/**
+	 * Definitions of data formats.
+	 */
+	enum Format {
```
joeygrover: This enum is unnecessary as the values should be pulled from the enums in this [class](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/enums/VideoStreamingCodec.java)

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
```diff
@@ -0,0 +1,488 @@
...
+	public RTPH264Packetizer(IStreamListener streamListener,
+	                         SessionType serviceType, byte sessionID, SdlSession session) throws IOException {
+
+		super(streamListener, null, serviceType, sessionID, session);
+
+		mServiceProtected = session.isServiceProtected(_serviceType);
+
+		if (bufferSize == 0) {
+			// fail safe
+			bufferSize = MAX_DATA_SIZE_FOR_ENCRYPTED_SERVICE;
```
joeygrover: This should be the agreed upon MTU for the video streaming service

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	/**
+	 * Overwrites MTU value. This is only for testing.
+	 *
+	 * @param mtu maximum payload size of SDL frames
+	 */
+	public void setMTU(int mtu) {
+		if (mtu <= 0) {
+			mtu = MAX_DATA_SIZE_FOR_ENCRYPTED_SERVICE;
```
joeygrover: See comment above. Also should check if the encryption flag is set before setting MTU.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
```diff
@@ -0,0 +1,488 @@
...
+				_streamListener.sendStreamPacket(pm);
+
+				frame.position(frame.position() + len);
+			}
+		}
+
+		// XXX: This is added to sync with StreamPacketizer. Actually it shouldn't be here since
+		// it's confusing that a packetizer takes care of End Service request.
+		if (connection != null) {
+			connection.endService(_serviceType, _rpcSessionID);
```
joeygrover: Agreed. This seems better as a callback from a packetizer to indicated the stream is empty/consumed and the implementing class should then be responsible for closing the service. Though it might be out of the scope for this PR and needs to be addressed in all packetizers.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	/**
+	 * Called by the encoder.
+	 *
+	 * @see com.smartdevicelink.encoder.IEncoderListener#onEncoderOutput(Format, byte[], long)
+	 */
+	@Override
+	public void onEncoderOutput(IEncoderListener.Format format, byte[] data, long ptsInUs) {
+		if (data == null || format != Format.H264_BYTE_STREAM) {
```
joeygrover: Will need to updated `H264_BYTE_STREAM` to correct enum

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
```diff
@@ -0,0 +1,488 @@
...
+
+	/**
+	 * Called by the encoder.
+	 *
+	 * @see com.smartdevicelink.encoder.IEncoderListener#onEncoderOutput(Format, byte[], int, int, long)
+	 */
+	@Override
+	public void onEncoderOutput(IEncoderListener.Format format, byte[] data,
+	                            int offset, int length, long ptsInUs) throws ArrayIndexOutOfBoundsException {
+		if (data == null || format != Format.H264_BYTE_STREAM) {
```
joeygrover: See comment above.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
```diff
@@ -0,0 +1,488 @@
...
+		mNALUnitReader.init(data, offset, length);
+		onEncoderOutput(mNALUnitReader, ptsInUs);
+	}
+
+	private void onEncoderOutput(NALUnitReader nalUnitReader, long ptsInUs) {
+		if (mPaused) {
+			return;
+		}
+
+		ByteBuffer nalu;
```
joeygrover: This should have a little bit more descriptive of a name, maybe `nalUnit`. Changes should be applied to the `outputRTPFrames` signature too.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
```diff
@@ -0,0 +1,488 @@
...
+
+		// Version = 2, Padding = hasPadding, Extension = 0, CSRC count = 0
+		frame.put((byte)(0x80 | (hasPadding ? 0x20 : 0)))
+			// Marker = isLast, Payload type = mPayloadType
+			.put((byte)((isLast ? 0x80 : 0) | (mPayloadType & 0x7F)))
+			.putChar(mSequenceNum)
+			.putInt(ptsIn90kHz)
+			.putInt(mSSRC);
+
+		assert frame.position() == FRAME_LENGTH_LEN + RTP_HEADER_LEN;
```
joeygrover: I don't believe `assert` should not be used in production code. It should be replaced with if statements and thrown exceptions.

joeygrover:
All systems capability and video streaming negotiation PRs have been merged. Please merge in `develop` into your branch and add corresponding functionality.

shoamano83:
Thanks @joeygrover, I'll take a look into the additional code.

sdl_android/src/main/java/com/smartdevicelink/encoder/IEncoderListener.java
shoamano83: OK, updated with https://github.com/smartdevicelink/sdl_android/pull/592/commits/e61dc691637a1a468b1066428577f67692b6c116. Since I want to specify both codec and format info (H.264 in byte-stream format in this case), I added VideoStreamingCodec and VideoStreamingProtocol.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
shoamano83: Fixed with https://github.com/smartdevicelink/sdl_android/pull/592/commits/e61dc691637a1a468b1066428577f67692b6c116.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
shoamano83: Fixed with https://github.com/smartdevicelink/sdl_android/pull/592/commits/e61dc691637a1a468b1066428577f67692b6c116.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
shoamano83: Sure, I've renamed all `nalu` to `nalUnit` with commit https://github.com/smartdevicelink/sdl_android/pull/592/commits/eb79ad607e8a1ffe55138db37e07517f38e6e7d9.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
shoamano83: Thanks, implemented with https://github.com/smartdevicelink/sdl_android/pull/592/commits/135a684668340311697bf49ab0b8b61c09c5e1e1.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
shoamano83: Actually this method was intended for testing. I found that private member variable can be altered with reflection, so I've updated the test case and removed this method completely with commit https://github.com/smartdevicelink/sdl_android/pull/592/commits/5325fc61689d6022006ab250b43096934a78750c. Sorry for your confusion!

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
shoamano83: Yes, I prefer to make it out of the scope of this PR.

sdl_android/src/main/java/com/smartdevicelink/streaming/RTPH264Packetizer.java
shoamano83: OK, replaced with if() check and RuntimeException, in commit https://github.com/smartdevicelink/sdl_android/pull/592/commits/9663bc8ba110fc647bdfb54dfc53435fca29f366.

shoamano83:
Hi @joeygrover, I've implemented video format selection mechanism. Please kindly review again.
Thanks.

shoamano83:
There are a few things I should note in the implementation.

* App needs to send `GetSystemCapability` request with systemCapabilityType=VIDEO_STREAMING and wait for its response prior to starting video streaming. Proxy will cache the capability information and will use it for format selection. Also, app should check for `preferredResolution` field to determine the video resolution.
If the app starts streaming without `GetSystemCapability` request/response, proxy will fall back to legacy StartService procedure, i.e. no codec/protocol/width/height information in the constructed payload.

* Currently the format selection is only enabled for `SdlProxyBase.createOpenGLInputSurface()` API. It is disabled for `startH264()` because:
  - InputStream and OutputStream don’t provide timestamp information so there is little point using RTP format.
  - `startH264()` APIs don’t provide width and height inputs from the app, and also lack notifying accepted resolution parameter back to the app.

* Resolution negotiation is not done by proxy. This is because there is no way to notify accepted resolution params back to the app. The app should provide appropriate width/height values to `createOpenGLInputSurface()` API as it does today.

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -3768,27 +3808,113 @@ public Surface createOpenGLInputSurface(int frameRate, int iFrameInterval, int w
...
+     */
+    private static List<VideoStreamingParams> createDesiredVideoParams(
+            VideoStreamingCapability hmiCapability,
+            int frameRate, int frameInterval, int width, int height, int bitrate,
+            VideoStreamingFormat[] availableFormats) {
+        ArrayList<VideoStreamingFormat> formats = new ArrayList<>();
+        if (hmiCapability != null && hmiCapability.getSupportedFormats() != null
+                && availableFormats != null) {
+            // supportedFormat is listed in HMI's preferred order
+            for (VideoStreamingFormat supportedFormat : hmiCapability.getSupportedFormats()) {
```
brettywhite: It cant be implied that the HMI is returning a list in a preferred order, or any order at all - each HMI implementation could be different. The proxy should have a bit of logic in it to prefer the RTP format over RAW, and fall back if RTP is not in the list.  (which is why the default is still raw in `VideoStreamingParams`).

brettywhite: Since right now we support only 2 protocols we could probably just do something like:
```java
List<VideoStreamingFormat> formatList = hmiCapability.getSupportedFormats();
	if (formatList.contains(VIDEO_STREAMING_FORMAT_H264_RTP)){
		// start up preferred RTP format
	} else {
		// iterate list and start up whats available (or just start RAW)
	}
```

shoamano83: Thanks for your comment.

Actually, the proposal has description of how the protocol and codec should be selected. HMI notifies the preferred order, and Proxy will try the most preferable format. Please refer to:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0058-video-streaming-capabilities.md

> Head unit's video capabilities are transferred to SDL proxy based on "System Capabilities Query" RPC proposed by SDL-0055 (https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0055-system_capabilities_query.md). The capabilities include:
> * (snip)
> * list of video formats that the head unit supports, in its preferred order. Here, a video format is a combination of "protocol" and "codec".

> When selecting a video format, SDL proxy should pick a format that is most preferable to the head unit and is also supported by the proxy itself. If HMI should reply a negative response with the format, then the proxy picks the second preferable format and try again. SDL proxy can continue this routine until it gets to the end of video format list. If none of the format is accepted (which is unlikely to happen), then the proxy may emit Start Service Control Frame without any video configuration parameters, to fall back to existing sequence.

Honestly speaking, I could implement the preference in Proxy side since we only have two formats (H.264 RTP and H.264 RAW) now. However, we may have more protocols and codecs in future, and it is possible that some HMIs prefer a codec over another codec, or a protocol over another. I wanted to make the spec and implementation flexible enough to support such cases as well.

So my assumption is, all HMIs return the list of supported format in their preferred order.

brettywhite: Ok, I do see in the proposal where it says the hmi/core decides the best order to return. However in testing, we did not see that behavior. To test I was editing `hmicapabilities.json` in core and it wasn't responding with RTP first, if both RTP and RAW were present under `videoStreamingCapability`. Please let me know if that was correct behavior or if hmi/core needs to be modified to do the choosing. Thank you

shoamano83: Thanks for the info. Let me look into it.
FYI, I found an issue in Core when reading HMI capabilities from a file. I fixed it here https://github.com/smartdevicelink/sdl_core/pull/1763 but I'm not sure if it is related.

shoamano83: I verified that following environment works. RTP is selected first, HMI rejects it, then RAW is selected and accepted.

Core: https://github.com/smartdevicelink/sdl_core/pull/1763/commits/412fef738770e97e0c392e5055ddeafb1a1bb9f0 which includes my PR #1763
Capability file: attached [hmi_capabilities.json.zip](https://github.com/smartdevicelink/sdl_android/files/1316608/hmi_capabilities.json.zip)
HMI: https://github.com/smartdevicelink/sdl_hmi/pull/26/commits/79f24a0ecef40ddba02273ff80eddd332c0ec7fa which includes PR #26
Proxy: latest feature/issue_515_rtp_packetizer branch

Please make sure that your SDL app sends GetSystemCapability RPC request with VIDEO_STREAMING type after first OnHMIStatus notification, and start video streaming after receiving GetSystemCapability response. Otherwise, Proxy will not receive HMI capabilities and will use RAW format.

brettywhite: Thank you for your quick reply. We will test it out in the next few days

brettywhite: @shoamano83 We were able to confirm that it works as you described. Thank you for your help

PR #630, opened 2017-10-05
Bugfix/follow up video streaming codec specific data

shoamano83:
This is a follow-up of PR https://github.com/smartdevicelink/sdl_android/pull/590, which fixes issue https://github.com/smartdevicelink/sdl_android/issues/589.

- The main purpose of this PR is to support the new interface introduced by PR https://github.com/smartdevicelink/sdl_android/pull/623.
- Additionally, I found that BufferInfo.offset value wasn't properly handled. This is also fixed.

This change has been tested on Nexus 6 (Android 5.0.1) and Xepria Z3+ (Android 6.0) phones.

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/encoder/SdlEncoder.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/encoder/SdlEncoder.java
@@ -176,25 +176,28 @@ public class SdlEncoder {
 						Log.i(TAG, "H264 codec specific data not retrieved yet.");
 					}
 				}
-				// append SPS and PPS in front of every IDR NAL unit
-				if ((mBufferInfo.flags & MediaCodec.BUFFER_FLAG_KEY_FRAME) != 0
-						&& mBufferInfo.size != 0
-						&& mH264CodecSpecificData != null) {
-					try {
-						mOutputStream.write(mH264CodecSpecificData, 0,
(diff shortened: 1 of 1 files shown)


sdl_android/src/main/java/com/smartdevicelink/encoder/SdlEncoder.java
```diff
@@ -187,11 +187,15 @@ public void drainEncoder(boolean endOfStream) {
 				}
 
 				if (mBufferInfo.size != 0) {
+					ByteBuffer encoderOutputBuffer = encoderOutputBuffers[encoderStatus];
 					byte[] dataToWrite = new byte[mBufferInfo.size];
-					encoderOutputBuffers[encoderStatus].get(dataToWrite,
-							mBufferInfo.offset, mBufferInfo.size);
```
shoamano83: mBufferInfo.offset indicates the offset of valid data in the codec output buffer, not the destination buffer.
FYI, I tested with Nexus 6 and Xperia Z3+ and this offset value was always 0 on both phones. So it's possible that this value isn't used, or is used only by specific phones.

Please also refer to the "famous" example http://bigflake.com/mediacodec/EncodeAndMuxTest.java.txt.

shoamano83:
@joeygrover Please kindly review.

PR #640, opened 2017-10-23
Check protocol version inside of createOpenGLInputSurface

### LINKED ISSUE #638: createOpenGLInputSurface should check if protocol version < 5
createOpenGLInputSurface in SdlProxyBase is currently failing for protocol versions that are less than 5.
The tryStartVideoStream method will always return null if the protocol version is less than 5, as a result the createOpenGLInputSurface will fail to be called (for any protocol version below 5).  Please see code block below:
https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L4032-L4038

We need to add an else if statement similar to the block linked below from the startVideoStream method where we check to see if the protocol version is less than 5 before attempting to createOpenGLInputSurface. 
https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L3956-L3958

Please see code snippet below for a sample solution:
```
        if (acceptedParams != null) {
            return sdlSession.createOpenGLInputSurface(frameRate, iFrameInterval, width,
                    height, bitrate, SessionType.NAV, sdlSession.getSessionId());
        } else if(getWiProVersion() < 5){
			sdlSession.setAcceptedVideoParams(new VideoStreamingParameters());
			return sdlSession.createOpenGLInputSurface(frameRate, iFrameInterval, width,
					height, bitrate, SessionType.NAV, sdlSession.getSessionId());
		} else {
            return null;
        }
```

mrapitis:
Fixes #638
Corrected the use case when createOpenGLInputSurface  is called with a protocol version lower than 5.

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
@@ -4032,7 +4032,11 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
         if (acceptedParams != null) {
             return sdlSession.createOpenGLInputSurface(frameRate, iFrameInterval, width,
                     height, bitrate, SessionType.NAV, sdlSession.getSessionId());
-        } else {
+        } else if(getWiProVersion() < 5){
+			sdlSession.setAcceptedVideoParams(new VideoStreamingParameters());
+			return sdlSession.createOpenGLInputSurface(frameRate, iFrameInterval, width,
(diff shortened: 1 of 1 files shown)


joeygrover:
Thank you for your contribution @mrapitis. As this PR and #644 were both attempting to accomplish the same thing and I would have had requested changes for both I created a PR that handled both and merged into the 4.4.0-rc1 branch.

PR #644, opened 2017-10-25
Fix/nack start video stream

### LINKED ISSUE #643: Process condition in protocol version less than 5
Here we found two issues in `develop` branch:

1. startVideoStream will execute on NACK
2. createOpenGLInputSurface never executed if protocol version < 5, this issue seems fixed in #638

`StartVideoStream()` method in SdlProxyBase currently return `sdlSession.startVideoStream()` if protocol version is less than 5, but `acceptedParams` has already returned **null**, this not make sense.

Below is the code snippet:
https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L3956-L3958

`tryStartVideoStream` method in SdlProxyBase, in `if(navServiceStartResponse)` condition, if protocol version is less than 5, `getAcceptedVideoParams` will return **null**,  VideoParams will not send to proxy, leads the VideoStream never start.
Below is the code snippet:
https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L4148-L4150

BrandonHe:
This PR is to fix #643 
Add the process code while protocol version < 5 situation.

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
@@ -3953,9 +3953,6 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 		VideoStreamingParameters acceptedParams = tryStartVideoStream(isEncrypted, parameters);
         if (acceptedParams != null) {
             return sdlSession.startVideoStream();
-        } else if(getWiProVersion() < 5){
-			sdlSession.setAcceptedVideoParams(new VideoStreamingParameters());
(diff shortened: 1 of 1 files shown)


BrandonHe:
@joeygrover 

Could you please help to review this? Thanks.

joeygrover:
Thank you for your contribution @BrandonHe. As this PR and #640 were both attempting to accomplish the same thing and I would have had requested changes for both I created a PR that handled both and merged into the 4.4.0-rc1 branch.

PR #653, opened 2017-11-21
bugfix / issue 652

### LINKED ISSUE #652: RemoteDisplay and video stream cannot be started after stopped
### Bug Report
The remote display fails to start successfully after it has been stopped.  Attempting to start the remote display again will result in a NACK received from SDL core.

##### Reproduction Steps
Create a mobile nav app that uses the remote display within the proxy.
Start the remote display and begin projecting video
Stop the remote display either through the onServiceEnded or internally
Start the remote display again, take note of the failure

##### Expected Behavior
The remote display will start successfully again

##### Observed Behavior
The remote display fails to start the service / project video

mrapitis:
Fixes #652

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
@@ -6390,7 +6390,9 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 		@Override
 		public void onServiceEnded(SdlSession session, SessionType type) {
 			if(SessionType.NAV.equals(type)){
-				stopStreaming();
+				if(remoteDisplay!=null){
+					stopStreaming();
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -6386,9 +6390,8 @@ public void onServiceStarted(SdlSession session, SessionType type, boolean isEnc
 		@Override
 		public void onServiceEnded(SdlSession session, SessionType type) {
 			if(SessionType.NAV.equals(type)){
-				dispose();
+				stopStreaming();
```
joeygrover: Changing from `dispose()` to `stopStreaming()` introduces an unnecessary call to stop the video streaming service. `stopStreaming() -> ISDL.onServiceEnded() -> stopStreaming()` We should try to avoid this in this class rather than relying on downstream efforts. I would suggest either simply check if `remoteDisplay != null` or introduce a new flag that denotes if the `VideoStreamingManager` is currently streaming.

mrapitis:
@joeygrover can you please re-review when you have a moment of time? thanks!

Issue #671, opened 2018-01-25
Android O, Unable To Start Router Service And Unable to Connect


Tnnnguyen:
### Summary
Upon startup, apps on devices running Android O try to query for running services using a different flow. When they find no such service, they will attempt to find an eligible service and start it. The problem is that if there is any issue in starting this eligible service, no service is running and apps cannot connect.

Such problem can occur if the eligible service belongs to an app that defines its SdlReceiver.java and SdlRouterService.java classes in two separate packages. See more [here](https://github.com/smartdevicelink/sdl_android/issues/648).

The end result can be a critical bug where none of the Android Sdl enabled multiplexing apps running on Android O devices can connect as long as the problematic eligible app is still installed (or not updated).

**Flow**:
- Wake up router, get a list of installed Sdl enabled apps [here](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java#L205).
- Sort the list, get the first app in the list and build an intent from it [here](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java#L207-L227). If the problematic app is first on the list, it will always end up at first until being deleted or updated.
- Use the first app and start its router service [here](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java#L238-L239) . If it cannot be started, nothing happens resulting in a dead end for apps trying to connect thereafter.

### Reproduction Steps
**Precondition**: 
- Device has no Sdl enabled multiplexing app.
- All apps being used are Sdl enabled multiplexing apps with 4.4.0 Sdl library

**Steps**:
1. Install App A which has SdlReceiver.java and SdlRouterService.java defined in different packages
2. Install App B
3. Install App C

### Expected Behavior
App A starts its own router service and connects through the service
App B either binds to App A's router service or reverts to legacy Bluetooth
App C either binds to App A's router service or reverts to legacy Bluetooth

### Observed Behavior
App A cannot start its own router service; cannot connect
App B tries to start App A's router service unsuccessfully; cannot connect
App C tries to start App A's router service unsuccessfully; cannot connect

### OS & Version Information
* Android Version: Android Oreo
* SDL Android Version: 4.4.0

### Proposed Solution
[...]

PR #672, opened 2018-01-26
bugfix / issue 671 Attempt to resolve intent before starting router service

Tnnnguyen:
Fixes [#671](https://github.com/smartdevicelink/sdl_android/issues/671)

### Risk

### Testing Plan
No API's were changed. Ran unit tests.

### Summary
If the intent to start the eligible router service does not resolve, attempt to start the next eligible router service.

##### Breaking Changes
* No breaking changes

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
@@ -223,14 +223,13 @@ public abstract class SdlBroadcastReceiver extends BroadcastReceiver{
 								}
 							});
 							serviceIntent = new Intent();
-							String pkgName;
-							do {
-								pkgName = apps.remove(0).activityInfo.packageName;
+							for (ResolveInfo app : apps) {
+								String pkgName = app.activityInfo.packageName;
 								serviceIntent.setComponent(new ComponentName(pkgName, pkgName + ".SdlRouterService"));
 								if (packageManager.resolveService(serviceIntent, PackageManager.MATCH_DEFAULT_ONLY) != null) {
 									break;
 								}
-							} while (apps.size() > 1);
+							}
 						} else{
 							Log.d(TAG, "No router service running, starting ours");
 							//So let's start up our service since no copy is running
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
@@ -222,9 +222,10 @@ public abstract class SdlBroadcastReceiver extends BroadcastReceiver{
 									return 0;
 								}
 							});
+							serviceIntent = new Intent();
+							String pkgName;
 							do {
-								serviceIntent = new Intent();
-								String pkgName = apps.remove(0).activityInfo.packageName;
+								pkgName = apps.remove(0).activityInfo.packageName;
 								serviceIntent.setComponent(new ComponentName(pkgName, pkgName + ".SdlRouterService"));
 								if (packageManager.resolveService(serviceIntent, PackageManager.MATCH_DEFAULT_ONLY) != null) {
 									break;
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
@@ -222,11 +222,14 @@ public abstract class SdlBroadcastReceiver extends BroadcastReceiver{
 									return 0;
 								}
 							});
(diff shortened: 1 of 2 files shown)


Tnnnguyen:
@joeygrover, fixed per code review request. Please review, thanks.

sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
```diff
@@ -48,5 +48,4 @@ public static boolean isServiceExported(Context context, ComponentName name) {
 		}
 		return sdlMultiList;
 	}
-
```
joeygrover: Please remove this as part of the commit.

sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -222,9 +222,14 @@ public int compare(ResolveInfo resolveInfo, ResolveInfo t1) {
 									return 0;
 								}
 							});
-							String packageName = apps.get(0).activityInfo.packageName;
-							serviceIntent = new Intent();
-							serviceIntent.setComponent(new ComponentName(packageName, packageName +".SdlRouterService"));
+							do {
```
joeygrover: Since we don't use the do/while syntax anywhere else in the project I would prefer if we stuck to either simple while or for loops.

sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -222,9 +222,14 @@ public int compare(ResolveInfo resolveInfo, ResolveInfo t1) {
...
+								serviceIntent = new Intent();
```
joeygrover: You do not need to create a new intent every iteration. You only need to overwrite the component name. This prevents us from being irresponsible with our memory management.

sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -222,9 +222,14 @@ public int compare(ResolveInfo resolveInfo, ResolveInfo t1) {
...
+								String pkgName = apps.remove(0).activityInfo.packageName;
```
joeygrover: Same comment about creating a new instance of Intent here for String

Tnnnguyen:
1. I will remove the extra line.
2. Creating new intent for each iteration: we can move the `new` operator out before the loop. I have thought about this but was not sure if setting component alone with refresh all properties of the intent and couldn't find anything in the doc about this. Since you are recommending only resetting the component, I will follow your recommendation.
3. I disagree on your comment `Same comment about creating a new instance of Intent here for String`. In Java, new string is created (in the string pool) every time there is an assignment like above.
4. In your recommendation, these 2 lines are repeated twice:

```packageName = apps.remove(0).activityInfo.packageName;```

```serviceIntent.setComponent(new ComponentName(packageName, packageName +".SdlRouterService"));```

all in the span of 7 lines. This is not great although it works.

The reason we don't have the do/while loop in the code base yet because there has not been a need for it. However, we don't have to stick to just while/for loop just because we have only been using them. The do/while loop is designed for this case (it just fits right in) and I think we should utilize it.

Here is what my next change will look like:

```
serviceIntent = new Intent();
do {
   String pkgName = apps.remove(0).activityInfo.packageName;
   serviceIntent.setComponent(new ComponentName(pkgName, pkgName + ".SdlRouterService"));
   if (packageManager.resolveService(serviceIntent, PackageManager.MATCH_DEFAULT_ONLY) != null) {
      break;
   }
} while (apps.size() > 1);
```
and your original recommendation:
```
String packageName = apps.remove(0).activityInfo.packageName;
serviceIntent = new Intent();
serviceIntent.setComponent(new ComponentName(packageName, packageName +".SdlRouterService"));
while (packageManager.resolveService(serviceIntent, PackageManager.MATCH_DEFAULT_ONLY) == null &&  apps.size() > 0) {
    packageName = apps.remove(0).activityInfo.packageName;
    serviceIntent.setComponent(new ComponentName(packageName, packageName +".SdlRouterService"));

}
```
Both accomplish the same task and use the same memory space. What do you think?

joeygrover:
My mistake on the String object, you are correct. The implicit coding style of the project does not include do/whiles and I won't merge code that contains them unless it can't be accomplish using for or while loops. In this instance if we want to avoid the repeated lines we can use the enhanced for loop as such `for(ResolveInfo app : apps){..}` which will accomplish nearly the same logic as the do/while with the only difference being we do not remove the object from the vector. Though the removal was unnecessary anyway as its scope is short lived and will be cleared as the method exits.

joeygrover:
#683 removes this code and takes care of the issue.

PR #681, opened 2018-02-12
Bugfix/issue 467

Tnnnguyen:
Fixes # [467](https://github.com/smartdevicelink/sdl_android/issues/467)

### Risk

### Testing Plan
Call SendMessage() method of WiProProtocol.java sending null data and confirm that it will not crash.

### Summary
If the data part of the message to be sent is null, the code will ignore the message (maintain the same behavior) and not crash.

### DIFF
--- a/sdl_android/src/androidTest/java/com/smartdevicelink/protocol/WiProProtocolTests.java
+++ b/sdl_android/src/androidTest/java/com/smartdevicelink/protocol/WiProProtocolTests.java
@@ -7,6 +7,10 @@ import com.smartdevicelink.SdlConnection.SdlConnection;
 import com.smartdevicelink.protocol.WiProProtocol.MessageFrameAssembler;
 import com.smartdevicelink.protocol.enums.SessionType;
 import com.smartdevicelink.test.SampleRpc;
+import com.smartdevicelink.util.DebugTool;
+import java.io.ByteArrayOutputStream;
+import java.util.Hashtable;
+import com.smartdevicelink.SdlConnection.SdlConnection;
 import com.smartdevicelink.test.SdlUnitTestContants;
 import com.smartdevicelink.transport.BaseTransportConfig;
 import com.smartdevicelink.transport.MultiplexTransportConfig;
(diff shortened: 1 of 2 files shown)


Tnnnguyen:
@joeygrover please review and merge, thanks.

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -231,58 +231,60 @@ else if (protocolMsg.getBulkData() != null) {
...
+					//Send the first frame
+					handlePacketToSend(firstHeader);
 					
-					int bytesToWrite = data.length - currentOffset;
-					if (bytesToWrite > mtu) {
-						bytesToWrite = mtu.intValue();
+					int currentOffset = 0;
+					byte frameSequenceNumber = 0;
+
+					for (int i = 0; i < frameCount; i++) {
```
joeygrover: Can you fix the formatting with this for loop? The tabs/spaces seem to be way off in this loop and included conditionals.

sdl_android/src/androidTest/java/com/smartdevicelink/protocol/WiProProtocolTests.java
```diff
@@ -342,6 +344,35 @@ public void testOverallocatingAccumulator(){
...
+	 * Test data being null in SendMessage() method
+	 *
+	 * @see WiProProtocol#SendMessage(ProtocolMessage)
+	 */
+	public void testNullDataInSendMessage() {
+		Object protocolMsg = null;
+		Field dataField = null;
+		Field msgLockField = null;
+		try {
+			protocolMsg = Class.forName("com.smartdevicelink.protocol.ProtocolMessage").newInstance();
```
joeygrover: Why is reflection necessary here? ProtocolMessage is a public class with a public no argument constructor.

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -231,58 +231,59 @@ else if (protocolMsg.getBulkData() != null) {
...
+						if (i < (frameCount - 1)) {
+		       				++frameSequenceNumber;
```
joeygrover: Still needs to be tabbed in one more time

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -231,58 +231,59 @@ else if (protocolMsg.getBulkData() != null) {
...
+		         			if (frameSequenceNumber == SdlPacket.FRAME_INFO_FINAL_CONNESCUTIVE_FRAME) {
```
joeygrover: This `if` statement still needs to be tabbed in one more time

Tnnnguyen: This could be a git issue.  It looks pretty good on my editors.
screen shot 2018-02-13 at 12 49 01 pm
screen shot 2018-02-13 at 12 48 28 pm

joeygrover: Attatched is the raw view. It looks like there are spaces and tabs mixed in and that is likely causing the issue. image

sdl_android/src/androidTest/java/com/smartdevicelink/protocol/WiProProtocolTests.java
Tnnnguyen: I recall that for this test, without using reflection, setting the `_data` field to null for ProtocolMessage will not work because the flow would be:
```java
ProtocolMessage ptcMsg = new ProtocolMessage();
ptcMsg.setData(null); // or any public setData method
```
The follow through initialization in ProtocolMessage class will always cause a npe.
Reflection allows setting the `_data` field will null without going through the initialization.

Come to think of it now, this `_data` field is next to impossible to be null and I was wondering if we should leave this test out because it only tests if the method can handle (ignore) the case where `_data` is null.

joeygrover: My question is strictly related to the constructor call. Setting the `_data` field to null can only happen through reflection for the reasons you stated. I believe the purpose of the PR is to test for a case in which `_data` is null, so if that's the case I think the test makes sense.

PR #682, opened 2018-02-14
Bugfix/issue 660 UAI Race condition logic

### LINKED ISSUE #660: App gets unexpectedly unregistered when stopped and then restarted on HS.
### Bug Report
I have a case where apps are being stopped (removed from recents list) several times and then restarted (both from dead app ping and from user relaunching app on HS).  Each time this happens the Router sends an UnregisterAppInterface request to SDLCore because "binder died for app..."
At the point where something bad happens, SDLCore sends a delayed (1 minute) UnregisterAppInterface response which router passes on to app's SDLProxy.  Proxy seems to disconnect because it later gets an OnHMIStatus FULL notification but when trying to send further RPCs to SDLCore the SDLProxy throws "SDL_UNAVAILABLE" exceptions.

Here are some relevant snippets from the log.
App is removed from recents list: https://gist.github.com/jthrun/c054c7c2093bd30d11225d203edd4cd5
App reregisters with router and sends registration messages to SDLCore: https://gist.github.com/jthrun/73abb7ce10ad0222d0a8acd737c40da9
App receives unregister response, I'm guessing this is from the createForceUnregisterApp() of one minute earlier: https://gist.github.com/jthrun/7711eb386e88c37307a84a540f3204d0
App then receives `OnHMIStatus` and tries to send RPCs but can't: https://gist.github.com/jthrun/9e43f9007d28cbe7731abb8439df83e7

##### Reproduction Steps
1. Launch _CTS Media 1_ app on the HU (do not launch on the HS) 
2. Allow HS to sit idle for 3 minutes 
3. Launch _Screens_ app on the HU and then HS
4. Launch _Simple_ app on the HU and then HS
5. Press **Back** HK to close _Simple_ app on the HU
6. Swipe close _Screens_ and _Simple_ apps on the HS
7. Launch _Simple_ app on the HU and then HS
8. Launch _Screens_ app on the HU and then HS
9. Press **Back** HK to close _Simple_ app on the HU
10. Swipe close _Screens_ and _Simple_ apps on the HU
11. Repeat steps 3-10 two more times
12. Launch _Simple_ app on the HU and HS
13. Press the **Back** HK to close _Simple_ app on the HU (leave _Simple_ app on HS foreground)
14. Allow HS to sit idle for 2 minutes
15. Launch _Screens_ app on the HS and then HU

Note the "Press **Back** HK" referred to in the above steps is a HK on the OEM's HU.  Pressing this HK takes the HU screen from the app's screen back to the app launcher screen.
Also note, I don't think the waiting periods are necessary but that is how the tester did the repro.

##### Expected Behavior
App should not be given UnregisterAppInterface (success:false) from an old force-unregister and/or should not disconnect from router if success:false.

##### Observed Behavior
App is disconnected even though it continues to get messages from SDLCore.

* Android Version: Samsung Galaxy S7 on 6.0.1
* SDL Android Version: 4.4.0
* Testing Against: Custom HMI; SDLCore internal build based on 4.1.2 tag plus a few hotfixes

I tried to add the APKs for the above apps but apparently github doesn't like big files.  Please contact me if you want the full logcat log or the apps that were used in testing.

brettywhite:
Fixes #660 

### Risk

### Testing Plan

I added 2 tests to handle the 2 cases that might be presented. Using reflection, i create the rare instance where an app who just was cleaned up is beaten by an app registering using the same sessionId. In this case, we check out the sessionHash and if its different (meaning a different app, the UAI packet is dropped). The second test covers the condition where it is the same app and the UAI packet is let through.

### Summary
This PR takes care of a rare race condition. When an app is removed and another app is connecting, there is a chance that the connecting app might be assigned the same session key as the app that was removed. If this happens, the unregister app interface can come back and remove the app that just connected.

Logic:

1. When an app is cleaned up in `attemptToCleanUpModule()`, make note of it
2. When a UAI packet is received, check to see if that session key exists in both the `cleanedSessionMap` and the `sessionHashIdMap`. 
3. If the key exists in both, compare the hash. If it is different, it means another app was assigned that session key and it returns, dropping the UAI packet. If they are the same, it allows it to continue sending the UAI packet

### Remaining Tasks

- [X] - Add tests

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
@@ -1486,15 +1486,15 @@ public class SdlRouterService extends Service{
 						BinaryFrameHeader binFrameHeader = BinaryFrameHeader.parseBinaryHeader(packet.getPayload());
 						if(binFrameHeader!=null && FunctionID.UNREGISTER_APP_INTERFACE.getId() == binFrameHeader.getFunctionID()){
 							Log.d(TAG, "Received an unregister app interface. Checking session hash before sending");
-							// make sure that we don't try and unregister recently added app that might have a
+							// make sure that we don't try to unregister a recently added app that might have a
 							// session ID of a removed app whose UAI was delayed
 							int hashOfRemoved = this.cleanedSessionMap.get(session, -1);
 							int currentHash = this.sessionHashIdMap.get(session, -1);
 							if (hashOfRemoved != -1 && currentHash != -1) {
 								// Current session contains key that was held before
 								if (hashOfRemoved != currentHash){
-									// App assigned same session id but is different app. keep this from being killed
-									Log.i(TAG, "same session id for different apps found, dropping packet");
+									// App assigned same session id but is a different app. Keep this from being killed
+									Log.d(TAG, "same session id for different apps found, dropping packet");
 									this.cleanedSessionMap.clear();
 									return false;
 								}
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
@@ -1447,14 +1447,10 @@ public class SdlRouterService extends Service{
 		 * @return whether or not the sending was successful 
 		 */
 		public boolean sendPacketToRegisteredApp(SdlPacket packet) {
-			Log.i(TAG, "SEND PACKAGE");
 			if(registeredApps!=null && (registeredApps.size()>0)){
-				Log.i(TAG, "INSIDE 1");
 				int session = packet.getSessionId();
 				boolean shouldAssertNewSession = packet.getFrameType() == FrameType.Control && (packet.getFrameInfo() == SdlPacket.FRAME_INFO_START_SERVICE_ACK || packet.getFrameInfo() == SdlPacket.FRAME_INFO_START_SERVICE_NAK);
 	    		String appid = getAppIDForSession(session, shouldAssertNewSession); //Find where this packet should go
-				Log.i(TAG, "APP ID "+ appid);
-				Log.i(TAG, "REGISTERED APPS "+ registeredApps.toString());
 	    		if(appid!=null && appid.length()>0){
 	    			RegisteredApp app;
 	    			synchronized(REGISTERED_APPS_LOCK){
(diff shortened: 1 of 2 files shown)


sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1480,6 +1483,26 @@ public boolean sendPacketToRegisteredApp(SdlPacket packet) {
...
+							// make sure that we don't try to unregister a recently added app that might have a
+							// session ID of a removed app whose UAI was delayed
+							int hashOfRemoved = this.cleanedSessionMap.get(session, -1);
+							int currentHash = this.sessionHashIdMap.get(session, -1);
+							if (hashOfRemoved != -1 && currentHash != -1) {
+								// Current session contains key that was held before
+								if (hashOfRemoved != currentHash){
+									// App assigned same session id but is a different app. Keep this from being killed
+									Log.d(TAG, "same session id for different apps found, dropping packet");
+									this.cleanedSessionMap.clear();
```
bilal-alsharifi: I think we should delete only the element that have the same session key not clean everything in cleanedSessionMap. Because cleaning everything will remove other recently cleaned apps that may have the same race condition later.

sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1480,6 +1483,26 @@ public boolean sendPacketToRegisteredApp(SdlPacket packet) {
 	    				}
 	    			}
 
+					if(packet.getFrameType() == FrameType.Single && packet.getServiceType() == SdlPacket.SERVICE_TYPE_RPC){
```
joeygrover: We should first check the size of `cleanedSessionMap` to see if any entries exist. This will prevent us from having to parse every RPC if there is no need to.

sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1480,6 +1483,26 @@ public boolean sendPacketToRegisteredApp(SdlPacket packet) {
...
+							// make sure that we don't try to unregister a recently added app that might have a
+							// session ID of a removed app whose UAI was delayed
+							int hashOfRemoved = this.cleanedSessionMap.get(session, -1);
+							int currentHash = this.sessionHashIdMap.get(session, -1);
+							if (hashOfRemoved != -1 && currentHash != -1) {
+								// Current session contains key that was held before
+								if (hashOfRemoved != currentHash){
+									// App assigned same session id but is a different app. Keep this from being killed
+									Log.d(TAG, "same session id for different apps found, dropping packet");
+									this.cleanedSessionMap.delete(session);
```
joeygrover: I believe it is actually delete this mapping based on the knowledge that a UAI response has been received for that session id and we had a hash id stored for it. This would help us not have an ever growing list if for some reason Core didn't properly respond with the UAI responses. I believe the function could be:
```java
if (hashOfRemoved != -1 ) {
     if (hashOfRemoved != currentHash){ 
        ...
    }

```

PR #685, opened 2018-02-21
Feature/issue 606 send multiple RPCS

### LINKED ISSUE #606: [SDL 0087] Sequentially Send Multiple RPCs
Proposal: [Sequentially Send Multiple RPCs](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0087-send-multiple-rpcs.md)

> This proposes to add an API allowing the developer to sequentially send multiple RPCs and to be notified when all have completed sending.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/263

Steering Committee Decision:

> The Steering Committee voted to accept this proposal with the following revisions: revise to add Android counterpart to this proposal. Additionally, when implementing, will need to investigate at what rate RPCs can be sent and received, as well as which RPCs do not need to be sent asynchronously.  Revisions to add the Android counterpart are now included within the .md file.

brettywhite:
Fixes #606 

### Risk

### Testing Plan
As testing this is challenging (requiring response from a core instance), the tests were mainly carried out against Manticore. Unit tests were added, but at this point cannot test full functionality as we cannot mock core. 

### Summary
Added 2 public methods:

- sendSequentialRequests - Allows multiple RPCRequests to be sent in an ArrayList and guarantees that they are executed in order.
- sendRequests - Allows multiple RPCRequests to be send in an ArrayList with no guarantee for order.

### Tasks Remaining:
- [x] Finish Async method
- [x] Finish synchronous method
- [x] Unit tests

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
@@ -3480,6 +3480,7 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 						sendSequentialRequests(rpcs, listener);
 					} catch (SdlException e) {
 						e.printStackTrace();
+						listener.onError(correlationId, Result.GENERIC_ERROR, e.toString());
 					}
 				}
 			}
--- a/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
@@ -3431,7 +3431,9 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 
 	/**
 	 * Takes a list of RPCRequests and sends it to SDL in a synchronous fashion. Responses are captured through callback on OnMultipleRequestListener.
-	 * For sending requests asynchronously, use sendRequests
+	 * For sending requests asynchronously, use sendRequests <br>
+	 *
+	 * <strong>NOTE: This will override any listeners on individual RPCs</strong>
 	 *
 	 * @param rpcs is the list of RPCRequests being sent
 	 * @param listener listener for updates and completions
@@ -3465,7 +3467,7 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 			return;
 		}
 
-		RPCRequest rpc = rpcs.get(0);
+		RPCRequest rpc = rpcs.remove(0);
 		if (rpc.getCorrelationID() == null) {
 			rpc.setCorrelationID(CorrelationIdGenerator.generateId());
 		}
@@ -3475,7 +3477,6 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 			public void onResponse(int correlationId, RPCResponse response) {
 				if (response.getSuccess()) {
 					// success
-					rpcs.remove(0);
 					listener.onUpdate(rpcs.size());
 					try {
 						// recurse after successful response of RPC
@@ -3483,10 +3484,12 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 					} catch (SdlException e) {
 						e.printStackTrace();
 					}
-				} else {
-					// fail
-					listener.onError(correlationId, response);
 				}
+			}
+
+			@Override
+			public void onError(int correlationId, Result resultCode, String info){
+				listener.onError(correlationId, resultCode, info);
 			}
 		});
 
@@ -3495,7 +3498,9 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 
 	/**
 	 * Takes a list of RPCRequests and sends it to SDL. Responses are captured through callback on OnMultipleRequestListener.
-	 * For sending requests synchronously, use sendSequentialRequests
+	 * For sending requests synchronously, use sendSequentialRequests <br>
+	 *
+	 * <strong>NOTE: This will override any listeners on individual RPCs</strong>
 	 *
 	 * @param rpcs is the list of RPCRequests being sent
 	 * @param listener listener for updates and completions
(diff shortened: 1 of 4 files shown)


sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -3425,12 +3428,164 @@ public void run() {
...
+			public void onResponse(int correlationId, RPCResponse response) {
+				if (response.getSuccess()) {
+					// success
+					rpcs.remove(0);
+					listener.onUpdate(rpcs.size());
+					try {
+						// recurse after successful response of RPC
+						sendSequentialRequests(rpcs, listener);
+					} catch (SdlException e) {
+						e.printStackTrace();
```
bilal-alsharifi: Isn't it better to throw the exception instead of catching it so the developer can handle it?

brettywhite: Hmm, my thought was that its not really a library problem at that point, if it got a bad response from core, I don't want it to error out the application. That is different than above where passing in empty or null lists is the developer's problem.

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -3425,12 +3428,164 @@ public void run() {
...
+					}
+				} else {
+					// fail
+					listener.onError(correlationId, response);
```
joeygrover: To actually get the error, another method has to be overridden in the `OnRPCResponseListener`, [`onError`](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/listeners/OnRPCResponseListener.java#L67)

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -3425,12 +3428,164 @@ public void run() {
...
+
+		int requestCount = rpcs.size();
+
+		// Break out of recursion, we have finished the requests
+		if (requestCount == 0) {
+			listener.onFinished();
+			return;
+		}
+
+		RPCRequest rpc = rpcs.get(0);
```
joeygrover: This can be `rpcs.remove(0)`. It will return the object that is removed and prevent us from having to make a call again later.

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -3425,12 +3428,164 @@ public void run() {
...
+				throw new SdlException("There is no valid connection to SDL. sendRequests cannot be called until SDL has been connected.", SdlExceptionCause.SDL_UNAVAILABLE);
+			}
+		}
+
+		if (rpcs == null || rpcs.size() == 0){
+			//Log error here
+			throw new SdlException("You must send some RPCs", SdlExceptionCause.INVALID_ARGUMENT);
+		}
+
+		for (int i = 0; i < rpcs.size(); i++) {
```
joeygrover: We should use a temp variable for size since we call it at least twice, and at everytime we have to loop through.

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -3425,12 +3428,164 @@ public void run() {
...
+			RPCRequest rpc = rpcs.get(i);
+			if (rpc.getCorrelationID() == null) {
+				rpc.setCorrelationID(CorrelationIdGenerator.generateId());
+			}
+			listener.addCorrelationId(rpc.getCorrelationID());
+			rpc.setOnRPCResponseListener(listener.getSingleRpcResponseListener());
```
joeygrover: I don't believe this is the right thing to do. If the developer has already added a listener to each of these RPCs this will overwrite them and will prevent a developer's logic from being executed.

brettywhite: Isn't the goal of sending everything at once is that the responses come in on the overridden `onResponse` method in the required `OnMultipleRequestListener`? This allows that.

sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/listeners/OnMultipleRequestListener.java
```diff
@@ -0,0 +1,58 @@
...
+import org.json.JSONException;
+
+import java.util.Vector;
+
+/**
+ * This is the listener for sending Multiple RPCs.
+ */
+public abstract class OnMultipleRequestListener extends OnRPCResponseListener {
+
+	Vector<Integer> correlationIds;
```
joeygrover: This can be final.

sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/listeners/OnMultipleRequestListener.java
```diff
@@ -0,0 +1,58 @@
...
+	OnRPCResponseListener rpcResponseListener;
+	private static String TAG = "OnMultipleRequestListener";
+
+	public OnMultipleRequestListener(){
+		setListenerType(UPDATE_LISTENER_TYPE_MULTIPLE_REQUESTS);
+		if(correlationIds == null){
```
joeygrover: This will always be true. You can remove this check.

sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/listeners/OnMultipleRequestListener.java
```diff
@@ -0,0 +1,58 @@
...
+					onUpdate(correlationIds.size());
+				}else{
+					onFinished();
+				}
+			}
+		};
+	}
+
+	public void addCorrelationId(int correlationid){
+		if(correlationIds == null){
```
joeygrover: The `correlationIds` should never be null if set it to final so this block can be removed.

sdl_android/src/main/java/com/smartdevicelink/proxy/rpc/listeners/OnRPCResponseListener.java
```diff
@@ -12,6 +12,10 @@
 	 * Listener type specific to putfile
 	 */
 	public final static int UPDATE_LISTENER_TYPE_PUT_FILE 		= 1;
+	/**
+	 * Listener type specific to putfile
```
joeygrover: Comment needs to be updated to reflect actual description of listener type.

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -130,12 +132,13 @@
...
 						POLICIES_CORRELATION_ID = 65535;
 	
-	// Sdlhronization Objects
+	// Sdl Synchronization Objects
 	private static final Object CONNECTION_REFERENCE_LOCK = new Object(),
 								INCOMING_MESSAGE_QUEUE_THREAD_LOCK = new Object(),
 								OUTGOING_MESSAGE_QUEUE_THREAD_LOCK = new Object(),
 								INTERNAL_MESSAGE_QUEUE_THREAD_LOCK = new Object(),
 								ON_UPDATE_LISTENER_LOCK = new Object(),
+								ON_UPDATE_MULTIPLE_LISTENER_LOCK = new Object(),
```
joeygrover: This isn't used anywhere so it seems as though it can be removed

PR #692, opened 2018-02-27
Fix issue #691

### LINKED ISSUE #691: Potential bad package added in AndroidTools.java
### Bug Report
We are adding the package info to the list even when the system cannot find the package and throws a NameNotFoundException. This can lead to a bad ResolveInfo and PackageInfo being used to create a SdlAppInfo object and subsequently a library crash. [Reference](https://github.com/smartdevicelink/sdl_android/blob/develop/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java#L104-L112).

```
for (ResolveInfo info : resolveInfoList) {
					PackageInfo packageInfo = null;
					try {
						packageInfo = packageManager.getPackageInfo(info.serviceInfo.packageName, 0);
					} catch (NameNotFoundException e) {
					} finally {
						sdlAppInfoList.add(new SdlAppInfo(info, packageInfo));
					}
				}
```

The code above adds every package to the list regardless because it's being done inside the finally clause. Should it pass on the package that results in an exception?

joeygrover:
Fixes #691

### Risk

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
@@ -105,10 +105,10 @@ public class AndroidTools {
 					PackageInfo packageInfo = null;
 					try {
 						packageInfo = packageManager.getPackageInfo(info.serviceInfo.packageName, 0);
+						sdlAppInfoList.add(new SdlAppInfo(info, packageInfo));
 					} catch (NameNotFoundException e) {
 						//Package was not found, likely a sign the resolve info can't be trusted.
 					}
-					sdlAppInfoList.add(new SdlAppInfo(info, packageInfo));
 
 				}
 			}
--- a/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
@@ -106,9 +106,10 @@ public class AndroidTools {
 					try {
 						packageInfo = packageManager.getPackageInfo(info.serviceInfo.packageName, 0);
 					} catch (NameNotFoundException e) {
-					} finally {
-						sdlAppInfoList.add(new SdlAppInfo(info, packageInfo));
+						//Package was not found, likely a sign the resolve info can't be trusted.
 					}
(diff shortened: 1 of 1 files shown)


sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
```diff
@@ -106,9 +106,10 @@ public static boolean isServiceExported(Context context, ComponentName name) {
 					try {
 						packageInfo = packageManager.getPackageInfo(info.serviceInfo.packageName, 0);
 					} catch (NameNotFoundException e) {
-					} finally {
-						sdlAppInfoList.add(new SdlAppInfo(info, packageInfo));
+						//Package was not found, likely a sign the resolve info can't be trusted.
 					}
+					sdlAppInfoList.add(new SdlAppInfo(info, packageInfo));
```
brettywhite: do we want to wrap this in a null check, or add null package info's in?

Tnnnguyen: This change will still add the bad package. It should be:
```
try {
 						packageInfo = packageManager.getPackageInfo(info.serviceInfo.packageName, 0);	 						packageInfo = packageManager.getPackageInfo(info.serviceInfo.packageName, 0);
sdlAppInfoList.add(new SdlAppInfo(info, packageInfo));
 					} catch (NameNotFoundException e) {	 					} catch (NameNotFoundException e) { // comments here
-					} 
+					
```

joeygrover: From my understanding the `getPackageInfo` method doesn't return null. It will instead throw an exception if the package can't be found. [Reference](https://developer.android.com/reference/android/content/pm/PackageManager.html#getPackageInfo)

joeygrover: Code under a `catch` block is not executed, that is why there is the `finally` block.

Tnnnguyen: There are more than one way to do this. I know my change works. The code that adds the package cannot be outside after the catch block this way. It will be executed when the catch block is done. In this case, it's almost similar to having it in a finally clause.

PR #695, opened 2018-03-02
Bugfix/issue 570 USB disconnecting deadlock [WIP]

### LINKED ISSUE #570: SdlProxyBase.dispose() sometimes deadlocks
We have noticed that sometimes while disconnecting from a TDK headunit using USB, the call to SdlProxyBase.dispose() never returns.  A thread in SdlConnection becomes deadlocked with another thread in SdlProxyBase (see PROTOCOL_REFERENCE_LOCK and CONNECTION_REFERENCE_LOCK)

This was first discovered on our own app, but it is also reproducible on the latest version 4.3 of the hello_sdl_android sample.
Tested on samsung galaxy note 4 (android 6.0.1) and galaxy S8 (android 7)

bilal-alsharifi:
Fixes #570 

This PR is **not ready** for review.

### Risk

### Summary
The deadlock happens when:
* ThreadA locks PROTOCOL_REFERENCE_LOCK in SdlConnection.java
* ThreadB locks CONNECTION_REFERENCE_LOCK in SdlProxyBase.java 
* ThreadB waits to lock PROTOCOL_REFERENCE_LOCK that is already locked by ThreadA 
* After that, in most case, ThreadA releases PROTOCOL_REFERENCE_LOCK and things go without a problem. But in some rare cases, ThreadA waits to lock CONNECTION_REFERENCE_LOCK  before releasing PROTOCOL_REFERENCE_LOCK which causes a deadlock between the two threads.

The scenario is little complicated but basically the reason for this rare case is that: when the connection is broken, this scenario might happen:
* ThreadB tries to clean the proxy, unregister the sessions, and remove them from listenerList
* ThreadB gets delayed and execution is switched to ThreadA
* ThreadA tries to run the following loop: 

for (SdlSession session : listenerList) {
      session.onTransportError(info, e);
}

The loop shouldn't run because ThreadB should have already cleaned listenerList. But because ThreadB got delayed, the loop runs and causes a deadlock to happen

The PR prevents the rare case from happening by putting threadA on wait until one of the following conditions happens:
* ThreadB finishes cleaning the proxy
* Timeout passed 

### Tasks Remaining:
- [ ] [Discuss the solution]
- [ ] [Add test cases]

### CLA


bilal-alsharifi:
Closing this pull request because I was merging with master instead of develop.

PR #696, opened 2018-03-02
[WIP] Bugfix/issue 570 USB disconnecting is causing a deadlock 

### LINKED ISSUE #570: SdlProxyBase.dispose() sometimes deadlocks

bilal-alsharifi:
Fixes #570 

This PR is **not ready** for review.

### Risk

### Testing Plan
I tested the fix manually by trying to connect/disconnect the USB cable 30 times continuously for multiple times. The app didn't freeze after disconnecting the cable.

### Summary
The deadlock happens when:
* USBTransportReader thread locks PROTOCOL_REFERENCE_LOCK in SdlConnection.java
* Main thread locks CONNECTION_REFERENCE_LOCK in SdlProxyBase.java 
* Main thread waits to lock PROTOCOL_REFERENCE_LOCK that is already locked by USBTransportReader thread 
* After that, in most case, USBTransportReader thread releases PROTOCOL_REFERENCE_LOCK and things go without a problem. But in some rare cases, USBTransportReader thread waits to lock CONNECTION_REFERENCE_LOCK  before releasing PROTOCOL_REFERENCE_LOCK which causes a deadlock between the two threads.

The scenario is little complicated but basically, when the connection is broken, the following  might happen:
* When we call proxy.dispose(), main thread tries to clean the proxy, unregister the sessions, and remove them from listenerList
* Main thread gets delayed and execution is switched to USBTransportReader thread
* USBTransportReader thread tries to run the following loop: 

     > for (SdlSession session : listenerList) {
&nbsp;&nbsp;session.onTransportError(info, e);
}

In normal cases, The loop shouldn't run because main thread should have already cleaned listenerList. But because the main thread got delayed, the loop runs and causes a deadlock.

The PR prevents the rare case from happening by:
* adding a flag to SdlSession to check if one thread unregistered the session, the other threads don't do actions on it
* making USBTransport.disconnect() synchronized 

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlConnection.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlConnection.java
@@ -14,6 +14,7 @@ import com.smartdevicelink.protocol.ProtocolMessage;
 import com.smartdevicelink.protocol.SdlPacket;
 import com.smartdevicelink.protocol.WiProProtocol;
 import com.smartdevicelink.protocol.enums.SessionType;
+import com.smartdevicelink.proxy.SdlProxyBase;
 import com.smartdevicelink.transport.BTTransport;
 import com.smartdevicelink.transport.BTTransportConfig;
 import com.smartdevicelink.transport.BaseTransportConfig;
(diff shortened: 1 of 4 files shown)


joeygrover:
Closing as the USB transport has been rewritten and this version is deprecated.

PR #700, opened 2018-03-09
Bugfix/issue 687 - send intents explicitly

### LINKED ISSUE #687: Android O, not all SDL Router broadcasts are explicit.
### Bug Report
Upon code inspection, I noticed that there are several places where sendBroadcast() is called to send a message to other SDL apps that still send as implicit broadcasts.  These should be converted to send explicit broadcasts in the same way as was done for the "start" broadcast.

These lines seem to be the ones related to client broadcasts (action = "sdl.router.startservice"):

https://github.com/smartdevicelink/sdl_android/blob/3266a8147f360ccf72882eb90f92a062be19f61f/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java#L695

https://github.com/smartdevicelink/sdl_android/blob/3266a8147f360ccf72882eb90f92a062be19f61f/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java#L1255

https://github.com/smartdevicelink/sdl_android/blob/3266a8147f360ccf72882eb90f92a062be19f61f/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java#L2032

There are a few others, one at least seems to be targeted at other SdlRouterServices and maybe it also should be explicit?

https://github.com/smartdevicelink/sdl_android/blob/3266a8147f360ccf72882eb90f92a062be19f61f/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java#L277

I am not sure about the other two:

https://github.com/smartdevicelink/sdl_android/blob/3266a8147f360ccf72882eb90f92a062be19f61f/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java#L201

https://github.com/smartdevicelink/sdl_android/blob/3266a8147f360ccf72882eb90f92a062be19f61f/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java#L901

If these are just internal then maybe they should be sent using [LocalBroadcastManager](https://developer.android.com/reference/android/support/v4/content/LocalBroadcastManager.html)?

* Android Version: 8.x+ (O and above)
* SDL Android Version: 4.4.0

brettywhite:
Fixes #687 

### Risk

### Testing Plan
Tested new method on TDK using a local versions of Hello Sdl Android to make sure nothing bad happened. Tested on `targetSdkVersion`'s of 25 and 26

### Summary
- Some broadcast intents in the `SdlRouterService` were being sent implicitly, while android O+ requires them to be sent explicitly. Created a new private method to do just that and modified the broadcasts which need it to use the new method.

- Removed an unnecessary unregister intent in `onTransportDisconnected` in which a broadcast was sent out when there were no registered apps. 

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
@@ -166,6 +166,11 @@ public class SdlRouterService extends Service{
 	private boolean isPingingClients = false;
 	int pingCount = 0;
 
+	/**
+	 * List of cached SDL apps used for explicit broadcasts
+	 */
+	private List<ResolveInfo> sdlApps;
+
 	
 	/* **************************************************************************************************************************************
 	****************************************************************************************************************************************
@@ -637,7 +642,7 @@ public class SdlRouterService extends Service{
 	        			if(service.pingIntent == null){
 	        				service.initPingIntent();
 	        			}
-	        			service.getBaseContext().sendBroadcast(service.pingIntent); 
+	        			service.sendExplicitBroadcast(service.pingIntent, false);
 	        		}
 	        		break;
 	        	default:
@@ -1213,8 +1218,6 @@ public class SdlRouterService extends Service{
 		
 		Intent startService = new Intent();  
 		startService.setAction(TransportConstants.START_ROUTER_SERVICE_ACTION);
-		//Perform our query prior to adding any extras or flags
-		List<ResolveInfo> sdlApps = getPackageManager().queryBroadcastReceivers(startService, 0);
 
 		startService.putExtra(TransportConstants.START_ROUTER_SERVICE_SDL_ENABLED_EXTRA, true);
 		startService.putExtra(TransportConstants.FORCE_TRANSPORT_CONNECTED, true);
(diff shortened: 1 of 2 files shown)


sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -2044,6 +2034,28 @@ private void stopClientPings(){
...
+
+	/**
+	 * Iterate through all of the apps that we know are listening for this intent
+	 * with an explicit intent (necessary for Android O SDK 26+)
+	 *
+	 * @param intent - the intent to send explicitly
+	 * @param isRepeatingPings - If this is coming from a method that will repeat pings,
+	 * we will use a cached list instead of querying each ping
+	 */
+	private void sendExplicitBroadcast(Intent intent, Boolean isRepeatingPings) {
```
joeygrover: Instead of a boolean here, why don't we pass a list of ResolveInfo that we want to send the intents to. If the list is null, we do the normal query of broadcast receivers.  Then for the ping runnable you can cache the list in the runnable itself.

sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1269,14 +1266,7 @@ public void onTransportDisconnected(TransportType type){
...
 		cachedModuleVersion = -1; //Reset our cached version
-		if(registeredApps== null || registeredApps.isEmpty()){
-			Intent unregisterIntent = new Intent();
-			unregisterIntent.putExtra(HARDWARE_DISCONNECTED, type.name());
-			unregisterIntent.putExtra(TransportConstants.ENABLE_LEGACY_MODE_EXTRA, legacyModeEnabled);
-			unregisterIntent.setAction(TransportConstants.START_ROUTER_SERVICE_ACTION);
-			sendBroadcast(unregisterIntent);
-			//return;
-		}else{
+		if(registeredApps != null){
```
joeygrover: The check for empty list should not have been removed.

Issue #703, opened 2018-03-12
Getting library version from app


Tnnnguyen:
### Summary
We need a way to get which SDL library version an app is using to better troubleshoot and more effective application support purpose.

PR #704, opened 2018-03-12
#703 Logging library version

Tnnnguyen:
- Add log line in SdlProxyBase.performCommon()
- Add line in gradle file to generate xml file that has library version string

Fixes [#703](https://github.com/smartdevicelink/sdl_android/issues/703)

### Risk

### Testing Plan
Connect with head unit and verify library version is logged.
Verify that xml file is generated with library version string.

### CLA

### DIFF
--- a/sdl_android/build.gradle
+++ b/sdl_android/build.gradle
@@ -9,6 +9,7 @@ android {
         versionCode 2
         versionName "4.3.1"
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        resValue "string", "SDL_LIB_VERSION", '\"' + versionName + '\"'
     }
     buildTypes {
         release {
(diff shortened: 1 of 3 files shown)


joeygrover:
@Tnnnguyen With this change we should actually change [this version](https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/proxy/Version.java) variable. Then the library already uses some log methods to write out the version.

Tnnnguyen:
We can log using your variable above. But this is logging when connecting. 

I'd like to have the version string in an xml file so we can parse the apk. Strings aren't easily read when parsing the apk but xml file is. This method allows us another way to get the library version without having to connect.

joeygrover:
Just change the variable to this:
`public static final String VERSION =  BuildConfig.VERSION_NAME;`

There are already public methods that use that so it will be nice to have.

PR #709, opened 2018-03-14
Feature/issue 657, 783 - Remote Control - OnRCStatus notification

### LINKED ISSUE #657: [SDL 0106] Remote Control - OnRCStatus notification
Proposal: [Remote Control - OnRCStatus notification](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0106-remote-control-onRcStatus-notification.md)

> One simple resource allocation rule is that at any given time only one application controls a RC module. For example, if SDL allocates a climate control module to application ONE, SDL rejects other applications to change the settings of the same climate control module. The module is freed when application ONE exits, disconnects, goes to HMI level NONE or no longer has the right to control the module due to various reasons (for example, a policy update revokes the access right to the module from the app, or SDL re-allocates the module to another app). In the last case, the mobile application is not aware of the change.

> A new notification originated from SDL is nice to have. If SDL allocates a module to an application or it de-allocates a module from an application, SDL shall send OnRCStatus notifications to both the mobile application and the HMI. The notification shall contain an array of the RC modules allocated to the application, so that the mobile application can know it gains or loses control of a module before a failure request, and the HMI can update the status on which application controls which modules.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/315

mrapitis:
Fixes #657, #783 

## Risk

## Testing Plan

Added new unit tests and expanded existing ones

## Summary

Added RPC's, structs, enums as defined in the Remote Control OnRCStatus notification proposal.

## CLA

- [x] I have signed the CLA

### DIFF
--- a/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SdlProxyBaseTests.java
+++ b/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SdlProxyBaseTests.java
@@ -621,9 +621,9 @@ public class SdlProxyBaseTests extends AndroidTestCase{
 			Log.i(TAG, "SendHapticDataResponse response from SDL: " + response);
 		}
 
-        @Override
-        public void onOnRCStatus(OnRCStatus notification) {
+		@Override
+		public void onOnRCStatus(OnRCStatus notification) {
 
-        }
-    }
+		}
+	}
 }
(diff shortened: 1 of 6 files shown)


bilal-alsharifi:
@mrapitis can we add `allowed` parameter to `OnRCStatus` as described in [this](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0172-onRcStatus-allowed.md ) proposal. It is a different proposal but it will make it easier for testing if we do them in one PR.

mrapitis:
@bilal-alsharifi we have added the allowed parameter as described in #783 Can you please re-review when time permits? Thanks!

PR #788, opened 2018-06-11
[WIP] Adding in a MultiplexTcpTransport

askirk:


sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -63,7 +68,10 @@
 	private HashMap<SessionType, TransportType> activeTransports = new HashMap<SessionType,TransportType>();
 	private List<TransportType> availableTransports = new ArrayList<>();
 
-	List<TransportType> requestedPrimaryTransports, requestedSecondaryTransports;
+	List<TransportType> requestedPrimaryTransports;
+	List<TransportType> supportedSecondaryTransports;
+	boolean registeredHighBandwidth = false;
```
joeygrover: Is this necessary? Wouldn't it be safer to check against active transports?

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -567,19 +618,43 @@ else if (frameInfo == FrameDataControlFrameType.StartSession.getValue()) {
...
+							for(int a : audio){
+								Log.d(TAG, "Supported audio transports: " + a);
+							}
+						}
+						if(video != null){
+							for(int v : video){
+								Log.d(TAG, "Supported video transports: " + v);
+							}
+						}
+						handleEnableSecondaryTransport((byte) packet.getSessionId(), secondary, audio, video);
```
joeygrover: Why is it necessary to call up to the SdlSession/IProtocolListener? At this point I believe the protocol layer should only retain what is available and what service should be connected. Looking at the code in SdlSession it seems it just calls right back to WiProProtocol.

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -678,7 +753,23 @@ else if (frameInfo == FrameDataControlFrameType.StartSession.getValue()) {
 					int serviceDataAckSize = BitConverter.intFromByteArray(packet.getPayload(), 0);
 					handleProtocolServiceDataACK(serviceType, serviceDataAckSize,(byte)packet.getSessionId ());
 				}
-			} //TODO else if register transport frame
+			} else if (frameInfo == FrameDataControlFrameType.RegisterSecondaryTransportACK.getValue()) {
+				handleRegisterSecondaryTransportACK((byte) packet.sessionId);
```
joeygrover: Only circular logic exists from this call. WiProtocol -> IProtocolListener -> WiProProtocol

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -678,7 +753,23 @@ else if (frameInfo == FrameDataControlFrameType.StartSession.getValue()) {
 					int serviceDataAckSize = BitConverter.intFromByteArray(packet.getPayload(), 0);
 					handleProtocolServiceDataACK(serviceType, serviceDataAckSize,(byte)packet.getSessionId ());
 				}
-			} //TODO else if register transport frame
+			} else if (frameInfo == FrameDataControlFrameType.RegisterSecondaryTransportACK.getValue()) {
+				handleRegisterSecondaryTransportACK((byte) packet.sessionId);
+			} else if (frameInfo == FrameDataControlFrameType.RegisterSecondaryTransportNACK.getValue()) {
+				String reason = (String) packet.getTag(ControlFrameTags.RPC.RegisterSecondaryTransportNAK.REASON);
+				handleRegisterSecondaryTransportNAKed((byte) packet.sessionId, reason);
```
joeygrover: Only circular logic exists from this call. WiProtocol -> IProtocolListener -> WiProProtocol

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -753,13 +844,21 @@ public void StartProtocolService(SessionType sessionType, byte sessionID, boolea
 		handlePacketToSend(header);
 	}
 
+	@Override
+	public void RegisterSecondaryTransport(byte sessionId) {
```
joeygrover: Method should be correct camel case regardless of what exists in other methods. Signature should also take in the specific transport that should be registered on.

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -117,6 +125,36 @@ public void setPrimaryTransports(List<TransportType> primaryTransports){
...
+	}
+
+	public void setSupportedServices(SessionType serviceType, List<Integer> order){
+		if(supportedServicesMap == null){
+			supportedServicesMap = new HashMap<>();
+		}
+		this.supportedServicesMap.put(serviceType, order);
+	}
+
+	public void setRegisteredHighBandwidth(boolean registeredHighBandwidth){
```
joeygrover: This method seems to be confusing. Shouldn't we set our registered transports when we receive the control frame ACK?

joeygrover:
There is a an issue where the `WiProProtocol` class tries to immediately connect to a secondary transport after being informed of their existence. This should only happen after the protocol has been requested to start a service that can only take place on a secondary transport. Essentially when a sart service request method is called, `WiProProtocol` needs to check if the desired transport for that service type, if it is secondary and that transport is not yet connected, that is when the sequence should be kicked off.

sdl_android/src/main/java/com/smartdevicelink/SdlConnection/ISdlConnectionListener.java
```diff
@@ -3,8 +3,11 @@
 import com.smartdevicelink.protocol.ProtocolMessage;
 import com.smartdevicelink.protocol.enums.SessionType;
 import com.smartdevicelink.transport.MultiplexTransportConfig;
+import com.smartdevicelink.transport.enums.TransportType;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
```
joeygrover: I believe all these imports can be removed

sdl_android/src/main/java/com/smartdevicelink/SdlConnection/SdlConnection.java
```diff
@@ -1,7 +1,9 @@
 package com.smartdevicelink.SdlConnection;
 
 
+import java.util.ArrayList;
```
joeygrover: Unused import

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -12,8 +12,10 @@
 import java.net.MalformedURLException;
 import java.net.ProtocolException;
 import java.net.URL;
+import java.util.ArrayList;
```
joeygrover: Unused import

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -12,8 +12,10 @@
 import java.net.MalformedURLException;
 import java.net.ProtocolException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Hashtable;
 import java.util.List;
+import java.util.Map;
```
joeygrover: Unused import

sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java
```diff
@@ -117,36 +183,142 @@ public void setPrimaryTransports(List<TransportType> primaryTransports){
 		this.requestedPrimaryTransports = primaryTransports;
 	}
 
+	public void setTransortPriorityForService(SessionType serviceType, List<Integer> order){
```
askirk: Typo: setTransport*

PR #791, opened 2018-06-13
Feature/Deprecate RPCRequestFactory class

### LINKED ISSUE #792: [SDL 0174] Deprecate RPCRequestFactory
Proposal: [Deprecate RPCRequestFactory](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0174-deprecate-rpc-request-factory.md)

> This proposal is to deprecate the `RPCRequestFactory` class and start using RPC constructors directly to create new RPC instances.

Review: smartdevicelink/sdl_evolution#509

Steering Committee Decision:
> The Steering Committee voted to accept this proposal with revisions. The revisions will be clarifying that the proposal itself is not a major change, but instead, at the time of the next major change, the deprecated class will be removed.

> The proposal .md was updated on June 13, 2018 to include the agreed upon revisions.

bilal-alsharifi:
Fixes #792 

### Risk

### Summary
- Deprecate `RPCRequestFactory` class as described in the proposal.
- Replace all calls to `RPCRequestFactory` factory methods in the library with RPC constructor calls.

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
@@ -123,8 +123,7 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 	private static final int PROX_PROT_VER_ONE = 1;
 	private static final int RESPONSE_WAIT_TIME = 2000;
 
-	private static final int SDL_MSG_MAJOR_VERSION = 1;
-	private static final int SDL_MSG_MINOR_VERSION = 0;
+	private static final com.smartdevicelink.util.Version MAX_SUPPORTED_RPC_VERSION = new com.smartdevicelink.util.Version("4.5.0");
 
 	private SdlSession sdlSession = null;
 	private proxyListenerType _proxyListener = null;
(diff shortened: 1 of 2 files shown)


sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -4550,9 +4562,31 @@ public void addCommand(Integer commandID,
 						   String menuText, Integer parentID, Integer position,
 						   Vector<String> vrCommands, String IconValue, ImageType IconType, Integer correlationID)
 			throws SdlException {
-		
-		AddCommand msg = RPCRequestFactory.buildAddCommand(commandID, menuText, parentID, position,
-			vrCommands, IconValue, IconType, correlationID);
+
+
+		AddCommand msg = new AddCommand(commandID);
+		msg.setCorrelationID(correlationID);
```
brettywhite: we don't need to explicitly set correlation IDs for RPCs anymore, unless you have a reason to specifically track it for the response. https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/proxy/RPCRequest.java#L26-L28

PR #813, opened 2018-07-16
lockscreen manager

BrettyWhite:

### Summary
Add lockscreen manager. 

This consists of adding:
- `LockScreenConfig`
- `LockScreenManager`
- `SDLLockScreenActivity`
- `activity_sdllock_screen.xml`
- `strings.xml`
- `sdl_lockscreen_icon` drawable

This is meant to really simplify the complex nature of lockscreens. 

The configuration is set through `SdlManager`

```java
LockScreenConfig lsc = new LockScreenConfig();
lsc.setBackgroundColor(R.color.colorPrimary);

SdlManager.Builder builder = new SdlManager.Builder(this, appId, appName, listener);
builder.setAppTypes(appType);
builder.setLockScreenConfig(lsc);
builder.setTransportType(transport);
proxy = builder.build();
proxy.start();
```
There are more setters in `LockScreenConfig`, but that should give you an idea.

Use:
- `setEnabled(boolean)` - whether or not we show the lockscreen for you. If false, you must do it all yourself. This should almost always be true
- `setBackgroundColor(int backgroundColor)` - you can override the default color with a color of your choice from your `colors.xml` file
- `setAppIcon(int appIconInt)` - override the default SDL lockscreen icon with an icon of your own. Set this to a drawable you have. 
- `lsc.showDeviceLogo(boolean deviceLogo)` - If you'd like to show the device (OEM) logo below your own, set this to true. If the vehicles' policy table returns a URL for an icon, it will show it, otherwise it would be blank.
- setCustomView(int customViewInt) - pass in your own lockscreen layout. This layout should be static, and fully built using xml. If this is set, it will override the other setters with the exception of `setEnabled`.

In your app, you must also declare the activity in your manifest:
```xml
<activity android:name="com.smartdevicelink.api.lockscreen.SDLLockScreenActivity"
               android:launchMode="singleInstance"/>
```

### Tests
Unit tests and extensive testing on TDKs and a production vehicle were performed

### CLA


sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,285 @@
...
+			if (!driverDistStatus) {
+				return LockScreenStatus.OPTIONAL;
+			} else {
+				return LockScreenStatus.REQUIRED;
+			}
+		}
+		return LockScreenStatus.OFF;
+	}
+
+	private void downloadLockScreenIcon(final String url){
```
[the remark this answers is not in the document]
brettywhite: just to put down what we had discussed in person, the reason I chose to place this here rather than the activity is because there is no good way to hold onto this image as the activity is instantiated and killed frequently. If there was a race condition where `showOEMLogo` is true and the lockscreen is instantiated before the download completes, it will not show the logo the first time, but will show it every subsequent time for the duration of the session. This prevents unnecessary repeated downloads for what i believe is a small trade off. This alone with only downloading if wanted (your comment from above), I believe this is pretty good

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -65,11 +67,10 @@
 
 	private CompletionListener initListener;
 	private int state = -1;
-	//public LockScreenConfig lockScreenConfig;
-
+	public LockScreenConfig lockScreenConfig;
```
joeygrover: Does the SDL Manager need to hold onto this for it's lifetime? Also why does it need to be public? Won't at least protected suffice?

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -298,34 +302,50 @@ public Builder setContext(Context context){
...
+			sdlManager.hmiTypes = hmiTypesDefault;
+			sdlManager.isMediaApp = false;
+		}
 
-			sdlManager.state = BaseSubManager.SETTING_UP;
+		if (sdlManager.lockScreenConfig == null){
+			// if lock screen params are not set, use default
+			LockScreenConfig lsc = new LockScreenConfig();
+			lsc.setEnabled(true);
+			lsc.setShowOEMLogo(false);
```
joeygrover: Shouldn't these simply be set as default values instead of having to call set methods to create a default instantiation of the class?

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenConfig.java
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	public void setEnabled(boolean enableAutomaticLockScreen){
+		this.enableAutomaticLockScreen = enableAutomaticLockScreen;
+	}
+
+	/**
+	 * Gets whether the lock screen is being managed for you
+	 * @return boolean
+	 */
+	public boolean getEnabled() {
```
joeygrover: please change to `isEnabled()`

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenConfig.java
```diff
@@ -0,0 +1,108 @@
...
+	public LockScreenConfig(){}
+
+	/**
+	 * If set to true, SDL will manage the showing and dismissing of the lock screen for you. <br>
+	 *
+	 * If false, you must manage the lock screen
+	 * @param enableAutomaticLockScreen boolean
+	 */
+	public void setEnabled(boolean enableAutomaticLockScreen){
+		this.enableAutomaticLockScreen = enableAutomaticLockScreen;
```
joeygrover: This variable is extremely verbose, can we simply use `enabled` instead? The `get/set` methods should have javadoc that explains what `enabled` means. The variable itself could have javadoc itself, but I believe it's pretty clear what it will mean.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenConfig.java
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	public boolean getEnabled() {
+		return enableAutomaticLockScreen;
+	}
+
+	/**
+	 * Set the int of the background color. You should define colors in your Colors.xml file
+	 * @param backgroundColor int of the color you'd like
+	 */
+	public void setBackgroundColor(int backgroundColor){
```
joeygrover: I believe this is a little misleading, are they setting the resource int of the color or the actual color value?

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenConfig.java
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Gets whether the lock screen is being managed for you
+	 * @return boolean
+	 */
+	public boolean getEnabled() {
+		return enableAutomaticLockScreen;
+	}
+
+	/**
+	 * Set the int of the background color. You should define colors in your Colors.xml file
```
joeygrover: `You should define...` should be ` Colors should be defined...`

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenConfig.java
```diff
@@ -0,0 +1,108 @@
...
+	public int getCustomView() {
+		return customViewInt;
+	}
+
+	/**
+	 * Whether or not to show the OEM logo on the default lock screen <br>
+	 * If using a Custom View, this will be ignored.
+	 * @param showOEMLogo - boolean
+	 */
+	public void setShowOEMLogo(boolean showOEMLogo) {
```
joeygrover: This should be more generalized instead of OEM logo, device or  hardware logo or something similar. Also I think `display` is better than `show` here. Maybe this`setDisplayDeviceLogo(boolean display)` then include more javadoc to explain that the device logo comes from the connected hardware. (`get` method should also be updated to reflect the new name).

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+
+	////
+	// SETUP
+	////
+
+	/**
+	 * Reads configuration set in SDLManager Builder
+	 * @param lockScreenConfig - the configuration set.
+	 */
+	private void readConfiguration(LockScreenConfig lockScreenConfig){
```
joeygrover: Does this method ever get called beyond the constructor? Might be best to just include this parsing in the constructor.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+ */
+public class LockScreenManager extends BaseSubManager {
+
+	private static final String TAG = "LockScreenManager";
+	private WeakReference<Context> context;
+	private HMILevel hmiLevel;
+	private boolean driverDistStatus, lockScreenEnabled, showOEMLogo;
+	private int lockScreenIcon, lockScreenColor, customView;
+	private OnRPCNotificationListener systemRequestListener, ddListener, hmiListener;
+	private String OEMIconUrl;
```
joeygrover: `OEM` should be `device` per previous comment. The OEM reference should also be updated throughout the class.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+
+		// set up system request listener
+		if (showOEMLogo) {
+			systemRequestListener = new OnRPCNotificationListener() {
+				@Override
+				public void onNotified(RPCNotification notification) {
+					// do something with the status
+					final OnSystemRequest msg = (OnSystemRequest) notification;
+					try {
+						Log.i(TAG, "SYSTEM REQUEST: " + notification.serializeJSON().toString());
```
joeygrover: This should be removed for release

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+					if (ddState.getState() == DriverDistractionState.DD_ON){
+						Log.i(TAG, "DD On Notification");
+						// launch lock screen
+						driverDistStatus = true;
+						launchLockScreenActivity();
+					}else{
+						Log.i(TAG, "DD Off Notification");
+						// close lock screen
+						driverDistStatus = false;
+						context.get().sendBroadcast(new Intent(SDLLockScreenActivity.CLOSE_LOCK_SCREEN_ACTION));
```
joeygrover: This is a potential NPE. The weak reference object will always hold a value but the object it references can be cleared.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+				showLockScreenIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+
+				// Extra parameters for customization of the lock screen view
+				showLockScreenIntent.putExtra(SDLLockScreenActivity.LOCKSCREEN_ICON_EXTRA, lockScreenIcon);
+				showLockScreenIntent.putExtra(SDLLockScreenActivity.LOCKSCREEN_COLOR_EXTRA, lockScreenColor);
+				showLockScreenIntent.putExtra(SDLLockScreenActivity.LOCKSCREEN_CUSTOM_VIEW_EXTRA, customView);
+				showLockScreenIntent.putExtra(SDLLockScreenActivity.LOCKSCREEN_OEM_ICON_EXTRA, showOEMLogo);
+				showLockScreenIntent.putExtra(SDLLockScreenActivity.LOCKSCREEN_OEM_ICON_BITMAP, lockScreenOEMIcon);
+
+				context.get().startActivity(showLockScreenIntent);
```
joeygrover: Potential NPE

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+			} else if (status == LockScreenStatus.OFF) {
+				context.get().sendBroadcast(new Intent(SDLLockScreenActivity.CLOSE_LOCK_SCREEN_ACTION));
```
joeygrover: Potential NPE

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+	 * 3. Build intent and start the SDLLockScreenActivity
+	 *
+	 * X. If the status is set to OFF, Send broadcast to close lock screen if it is open
+	 */
+	private void launchLockScreenActivity(){
+		// intent to open SDLLockScreenActivity
+		// pass in icon, background color, and custom view
+		if (lockScreenEnabled) {
+			LockScreenStatus status = getLockScreenStatus();
+			if (hmiLevel == HMILevel.HMI_FULL && status == LockScreenStatus.REQUIRED) {
```
joeygrover: Checking for HMI_FULL here seems redundant. The `getLockScreenStatus` call already checks for hmi level, am I missing something?

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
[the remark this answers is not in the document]
joeygrover: Why not just add a ping to the Activity that the image has been dowloaded and is ready for retrieval? The activity already has a broadcast receiver so we should be able to simply notify that with an intent.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,274 @@
...
+
+	protected boolean getLockScreenEnabled(){
+		return lockScreenEnabled;
+	}
+
+	protected Bitmap getLockScreenOEMIcon(){
+		return lockScreenOEMIcon;
+	}
+
+	protected LockScreenStatus testGetLockScreenStatus(){
```
joeygrover: Why is this method included? Can't the getLockScreenStatus method just be `protected` instead of `private` since that's what this method essentially does?

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/SDLLockScreenActivity.java
```diff
@@ -0,0 +1,104 @@
...
+
+	@Override
+	protected void onDestroy() {
+		unregisterReceiver(closeLockScreenBroadcastReceiver);
+		super.onDestroy();
+	}
+
+	public void initializeActivity(Intent intent){
+		if (intent != null){
+			boolean showOEMLogo = intent.getBooleanExtra(LOCKSCREEN_OEM_ICON_EXTRA, true);
```
joeygrover: Is this really needed? Since this is only a local variable and only used in combination of checking if an icon was included, it depends on `lockScreenOEMIcon` holding a value anyway.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/SDLLockScreenActivity.java
```diff
@@ -0,0 +1,104 @@
...
+				}
+
+				if (showOEMLogo && lockScreenOEMIcon != null){
+					showOEMIcon();
+				}
+			}
+		}
+	}
+
+	private void changeBackgroundColor() {
```
joeygrover: Doesn't make more sense to have these methods take in the values they are setting and remove the values from global reference? After they get set is there a reason they are still needed?

sdl_android/src/main/res/layout/activity_sdllock_screen.xml
```diff
@@ -0,0 +1,44 @@
...
+            android:id="@+id/lockscreen_image"
+            android:layout_width="120dp"
+            android:layout_height="120dp"
+            android:scaleType="fitXY"
+            android:layout_gravity="center_horizontal"
+            android:contentDescription="@string/lockscreen_image_description"
+            android:background="@drawable/sdl_lockscreen_icon"/>
+
+        <ImageView
+            android:id="@+id/OEM_image"
```
joeygrover: id should be renamed as mentioned in previous comments but should also not include capitals letters in the future.

kshala-ford:
I just reviewed this PR briefly so please accept my apoligies if I skipped the part about the following issue:

In [LockScreenActivity.java](https://github.com/smartdevicelink/hello_sdl_android/blob/master/hello_sdl_android/app/src/main/java/com/sdl/hellosdlandroid/LockScreenActivity.java) the lock screen considers app's activity lifecycle. The lock screen should not appear if the app is not in foreground on the phone.  Are you planning something similar?

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
theRealBrett: the developer uses the LockScreenConfig to 'build' their lockscreen. It is essentially the instructions to the manager on how to behave:

From `SdlService.java`:

```java
			LockScreenConfig lsc = new LockScreenConfig();
			lsc.setEnabled(true);
                        lsc.setBackgroundColor(R.color.colorPrimary);
			lsc.setShowOEMLogo(true);

			SdlManager.Builder builder = new SdlManager.Builder();
			builder.setAppId(appId);
			builder.setAppName(appName);
			builder.setAppTypes(appType);
			builder.setContext(this);
			builder.setLockScreenConfig(lsc);
			builder.setTransportType(transport);
			proxy = builder.build();
```

sdl_android/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="lockscreen_text">Locked for your safety</string>
```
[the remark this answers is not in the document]
theRealBrett: Yes it is the same text. No there is no way to change it in iOS.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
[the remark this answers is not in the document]
theRealBrett: generalized the broadcast receiver and added this. done.

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
theRealBrett: you are right. set to private, and cleared on proxy start. done.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,276 @@
...
+public class LockScreenManager extends BaseSubManager {
+
+	private static final String TAG = "LockScreenManager";
+	private WeakReference<Context> context;
+	private HMILevel hmiLevel;
+	private boolean driverDistStatus, lockScreenEnabled, displayDeviceLogo;
+	private int lockScreenIcon, lockScreenColor, customView;
+	private OnRPCNotificationListener systemRequestListener, ddListener, hmiListener;
+	private String deviceIconUrl;
+	private Bitmap lockScreenOEMIcon;
```
joeygrover: Please refactor this variable to "lockScreenDeviceIcon" to be consistent.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenConfig.java
```diff
@@ -0,0 +1,114 @@
...
+		return customViewInt;
+	}
+
+	/**
+	 * Whether or not to show the Display Device's logo on the default lock screen <br>
+	 * The logo will come from the connected hardware, if set by the manufacturer <br>
+	 * If using a Custom View, this will be ignored.
+	 * @param displayDeviceLogo - boolean
+	 */
+	public void displayDeviceLogo(boolean displayDeviceLogo) {
```
joeygrover: Please refactor to `setDisplayDeviceLogo`

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/SDLLockScreenActivity.java
```diff
@@ -0,0 +1,115 @@
...
+public class SDLLockScreenActivity extends Activity {
+
+	public static final String LOCKSCREEN_COLOR_EXTRA = "LOCKSCREEN_COLOR_EXTRA";
+	public static final String LOCKSCREEN_ICON_EXTRA = "LOCKSCREEN_ICON_EXTRA";
+	public static final String LOCKSCREEN_DEVICE_ICON_EXTRA = "LOCKSCREEN_DEVICE_ICON_EXTRA";
+	public static final String LOCKSCREEN_DEVICE_ICON_BITMAP = "LOCKSCREEN_DEVICE_ICON_BITMAP";
+	public static final String LOCKSCREEN_CUSTOM_VIEW_EXTRA = "LOCKSCREEN_CUSTOM_VIEW_EXTRA";
+	public static final String LOCKSCREEN_ICON_DOWNLOADED = "LOCKSCREEN_ICON_DOWNLOADED";
+	public static final String CLOSE_LOCK_SCREEN_ACTION = "CLOSE_LOCK_SCREEN";
+	private static final String TAG = "SDLLockScreenActivity";
```
joeygrover: Please move this to the top of the class.

sdl_android/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,6 @@
...
+    <string name="lockscreen_image_description">SDL Icon</string>
+    <string name="lockscreen_device_image_description">Display Device Icon</string>
```
joeygrover: Please remove `Display`

sdl_android/src/main/res/layout/activity_sdllock_screen.xml
```diff
@@ -0,0 +1,44 @@
...
+            android:id="@+id/lockscreen_image"
+            android:layout_width="120dp"
+            android:layout_height="120dp"
+            android:scaleType="fitXY"
+            android:layout_gravity="center_horizontal"
+            android:contentDescription="@string/lockscreen_image_description"
+            android:background="@drawable/sdl_lockscreen_icon"/>
+
+        <ImageView
+            android:id="@+id/display_device_image"
```
joeygrover: Please remove `display_`

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/SDLLockScreenActivity.java
```diff
@@ -0,0 +1,115 @@
...
+		RelativeLayout layout = findViewById(R.id.lockscreen_relative_layout);
+		layout.setBackgroundColor(getResources().getColor(customColor));
+	}
+
+	private void changeIcon(int customIcon) {
+		ImageView lockscreen_iv = findViewById(R.id.lockscreen_image);
+		lockscreen_iv.setBackgroundResource(customIcon);
+	}
+
+	private void displayDeviceImage(Bitmap displayDeviceIcon) {
```
joeygrover: Please refactor to `setDeviceIcon(Bitmap deviceIcon)`

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/SDLLockScreenActivity.java
```diff
@@ -0,0 +1,115 @@
...
+
+	@Override
+	protected void onDestroy() {
+		unregisterReceiver(lockScreenBroadcastReceiver);
+		super.onDestroy();
+	}
+
+	public void initializeActivity(Intent intent){
+		if (intent != null){
+			boolean displayDeviceImage = intent.getBooleanExtra(LOCKSCREEN_DEVICE_ICON_EXTRA, true);
```
joeygrover: Please refactor to `deviceImage` and all other references please.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/LockScreenManager.java
```diff
@@ -0,0 +1,252 @@
...
+
+		new Thread(new Runnable(){
+			@Override
+			public void run(){
+				try{
+					deviceLogo = HttpUtils.downloadImage(url);
+					Intent intent = new Intent(SDLLockScreenActivity.LOCKSCREEN_DEVICE_LOGO_DOWNLOADED);
+					intent.putExtra(SDLLockScreenActivity.LOCKSCREEN_DEVICE_LOGO_EXTRA, deviceLogoEnabled);
+					intent.putExtra(SDLLockScreenActivity.LOCKSCREEN_DEVICE_LOGO_BITMAP, deviceLogo);
+					context.get().sendBroadcast(intent);
```
joeygrover: Please add a null check for `context.get()` since this executes in a new thread a later time the previous check is not adequate.

sdl_android/src/main/java/com/smartdevicelink/api/lockscreen/SDLLockScreenActivity.java
```diff
@@ -0,0 +1,124 @@
...
+		}
+	};
+
+	@Override
+	protected void onCreate(Bundle savedInstanceState) {
+		super.onCreate(savedInstanceState);
+		this.requestWindowFeature(Window.FEATURE_NO_TITLE);
+
+		// set any parameters that came from the lock screen manager
+		initializeActivity(getIntent());
```
joeygrover: This call to getIntent() will only return the first intent used to start the activity. You must also override the activity's `onNewIntent` method. Add the following block:
```
@Override
protected void onNewIntent(Intent intent) {
    super.onNewIntent(intent);
    setIntent(intent);
}
```

PR #819, opened 2018-07-18
WIP: Feature/manager ready callback

brettywhite:
Fixes 2 issues:

1. Allows developer to know when sdlmanager is ready to send requests
2. doesnt start submanagers until active proxy connection

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
@@ -298,34 +298,28 @@ public class SdlManager{
 			return this;
 		}
 
-		@SuppressWarnings("unchecked")
 		public SdlManager build() {
-			try {
-
-				if (sdlManager.appName == null) {
-					throw new IllegalArgumentException("You must specify an app name by calling setAppName");
-				}
-
-				if (sdlManager.appId == null) {
-					throw new IllegalArgumentException("You must specify an app ID by calling setAppId");
-				}
-
-				if (sdlManager.hmiTypes == null) {
-					Vector<AppHMIType> hmiTypesDefault = new Vector<>();
-					hmiTypesDefault.add(AppHMIType.DEFAULT);
-					sdlManager.hmiTypes = hmiTypesDefault;
-					sdlManager.isMediaApp = false;
-				}
-
-				if (sdlManager.hmiLanguage == null){
-					sdlManager.hmiLanguage = Language.EN_US;
-				}
-
-				sdlManager.proxy = new SdlProxyBase(sdlManager.proxyBridge, sdlManager.appName, sdlManager.shortAppName, sdlManager.isMediaApp, sdlManager.hmiLanguage, sdlManager.hmiLanguage, sdlManager.hmiTypes, sdlManager.appId, sdlManager.transport, sdlManager.vrSynonyms, sdlManager.ttsChunks, sdlManager.dayColorScheme, sdlManager.nightColorScheme) {};
-				sdlManager.state = BaseSubManager.SETTING_UP;
-			} catch (SdlException e) {
-				e.printStackTrace();
+			if (sdlManager.appName == null) {
+				throw new IllegalArgumentException("You must specify an app name by calling setAppName");
 			}
+
+			if (sdlManager.appId == null) {
+				throw new IllegalArgumentException("You must specify an app ID by calling setAppId");
+			}
+
+			if (sdlManager.hmiTypes == null) {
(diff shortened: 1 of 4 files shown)


sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -380,6 +404,20 @@ public void sendRPCs(List<? extends RPCMessage> rpcs, final OnMultipleRequestLis
 
 	// LIFECYCLE / OTHER
 
+	// STARTUP
+	public void start(CompletionListener listener){
+		initListener = listener;
+		if(initialized){
+			listener.onComplete(true);
```
joeygrover: Potential NPE

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -380,6 +404,20 @@ public void sendRPCs(List<? extends RPCMessage> rpcs, final OnMultipleRequestLis
 
 	// LIFECYCLE / OTHER
 
+	// STARTUP
+	public void start(CompletionListener listener){
+		initListener = listener;
+		if(initialized){
```
joeygrover: Shouldn't this be kept track of by a state rather than a flag?

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -75,6 +79,27 @@
     private PermissionManager permissionManager;
     */
 
+	CompletionListener subManagerListener = new CompletionListener() {
+		boolean allSucceeded = true;
+		int subManagerCount = 1; // Update per amount of sub managers implemented
+		@Override
+		public void onComplete(boolean success) {
```
joeygrover: This needs a synchronized lock. Submanagers will complete asynchronously and therefore possibly try to alter the same variables at the same time.

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -380,6 +404,20 @@ public void sendRPCs(List<? extends RPCMessage> rpcs, final OnMultipleRequestLis
...
+	// STARTUP
+	public void start(CompletionListener listener){
+		initListener = listener;
+		if(initialized){
+			listener.onComplete(true);
+		}
+	}
+
+	@Override
+	public void onProxyConnected() {
```
joeygrover: After further review, I believe the SdlManager class should create an anonymous class of the Lifecycle Listener interface instead of implement. These methods should not be exposed to the developer.

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -380,6 +404,20 @@ public void sendRPCs(List<? extends RPCMessage> rpcs, final OnMultipleRequestLis
 
 	// LIFECYCLE / OTHER
 
+	// STARTUP
+	public void start(CompletionListener listener){
```
joeygrover: Please add javadoc to this method to inform the developer of what they can expect by calling it.

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -75,6 +79,27 @@
     private PermissionManager permissionManager;
     */
 
+	CompletionListener subManagerListener = new CompletionListener() {
+		boolean allSucceeded = true;
+		int subManagerCount = 1; // Update per amount of sub managers implemented
```
joeygrover: This seems like an error waiting to happen. We should find a less brute force way.

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -75,6 +79,27 @@
     private PermissionManager permissionManager;
     */
 
+	CompletionListener subManagerListener = new CompletionListener() {
+		boolean allSucceeded = true;
```
joeygrover: If a single sub manager fails do we even continue? Or should the SdlManager shutdown and throw an error?

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -75,6 +79,27 @@
     private PermissionManager permissionManager;
     */
 
+	CompletionListener subManagerListener = new CompletionListener() {
```
joeygrover: Can be final

PR #820, opened 2018-07-20
Feature/issue 654 AudioStreamManager

### LINKED ISSUE #654: [SDL 0113] SDLAudioStreamManager
Proposal: [SDLAudioStreamManager](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0113-audiostreammanager.md)

> This proposal provides an interface for automatically transcoding audio files – such as mp3 – to the SDL compatible PCM format, queueing these files, and playing them using the PCM audio stream available to `NAVIGATION` apps.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/347

Steering Committee Decision:

> When reviewing this proposal, there was a question about a use case for audio streaming. The author of this proposal advised that audio streaming would be used for navigation or projection apps only, and that navigation apps that are playing audio (voice of navigation directions, for example) often don't have directions available in PCM format, or the correct PCM format. The AudioStreamManager will take the format that the app has (MP3, for example), convert it into PCM format and play. This allows the app to use the manager instead of having to convert the file themselves.  After this discussion, the Steering Committee voted to accept this proposal.

kshala-ford:
Fixes #654 

### Risk

### Testing Plan
On functional side the resample method is challenged with test samples.
The audio manager should be tested with test audio that is valid (happy path).
A test audio file (digital signal, 80% amp, 200hz) should be used to validate correct decoding and resampling.

### Summary
This PR implements the AudioStreamManager which comes with a AudioDecoder to decode audio files. The decoder is based on [MediaCodec](https://developer.android.com/reference/android/media/MediaCodec) and linear interpolation for resampling. It includes sample type conversion between unsigned 8 bit, signed 16 bit and float samples.

##### Breaking Changes
* No breaking changes are expected with this code

##### Enhancements
* Adds audio decoding and an audio streaming manager.

### Tasks Remaining:
- [x] [Write AudioDecoder class]
- [x] [Complete AudioStreamingManager]
- [x] [Add code to support api 16-20] 
- [x] [Add more test cases to test the manager and decoded audio data]

### CLA


kshala-ford:
@askirk The code is ready for review. During the next day I want to add two more test cases to test the actual audio manager (integration test).

sdl_android/src/androidTest/java/com/smartdevicelink/test/api/audio/AudioStreamManagerTest.java
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        return file;
+    }
+
+    public interface StreamManagerTestCallback {
+        void onTestResult(boolean isPassed, Exception e);
+    }
+
+    ISdl sdlInterface = new ISdl() {
```
askirk: You can use mockito to mock this and add specific "answers" when certain methods are called. See here https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/androidTest/java/com/smartdevicelink/api/FileManagerTests.java#L142-L144

sdl_android/src/androidTest/java/com/smartdevicelink/test/api/audio/AudioStreamManagerTest.java
```diff
@@ -0,0 +1,313 @@
...
+            method.setAccessible(true);
+        } catch (NoSuchMethodException e) {
+            e.printStackTrace();
+            fail();
+        }
+        return method;
+    }
+
+    private BaseAudioDecoder createMockBaseAudioDecoder() {
+        AudioDecoderListener listener = new AudioDecoderListener() {
```
askirk: Can use mockito to mock,
`AudioDecoderListener listener = mock(AudioDecoderListener.class);`

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,192 @@
...
+import com.smartdevicelink.proxy.rpc.enums.SamplingRate;
+
+import java.io.File;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.util.LinkedList;
+import java.util.Queue;
+
+@RequiresApi(api = Build.VERSION_CODES.JELLY_BEAN)
+public class AudioStreamManager implements ISdlServiceListener {
```
askirk: Instead of implementing `ISdlServiceListener` which exposes the public methods to developers, you can make a private instance an add that in L71. See https://github.com/smartdevicelink/sdl_android/blob/d0097ac58cc405746f924cf57e3d596d5924c8fa/sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java#L60-L80 for reference

askirk: This class should extend `BaseSubManager`, which provides a way for the encapsulating `SdlManager` to start each submanager and be notified when they are ready or error out.

https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/BaseSubManager.java

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        switch (sampleType) {
+            case BitsPerSample.EIGHT_BIT:
+                sdlSampleType = SampleType.UNSIGNED_8_BIT;
+                break;
+            case BitsPerSample.SIXTEEN_BIT:
+                sdlSampleType = SampleType.SIGNED_16_BIT;
+                break;
+        }
```
askirk: This is where you would use the `BaseSubManager` method to set the state to `READY`:

`transitionToState(BaseSubManager.READY);`

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    public void startAudioService(boolean encrypted) {
```
askirk: Change name to `start()` based on proposal https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0113-audiostreammanager.md#android

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,192 @@
...
+        if (sdlInterface != null && sdlInterface.isConnected()) {
+            sdlInterface.addServiceListener(SessionType.PCM, this);
+            sdlInterface.startAudioService(encrypted);
+        }
+    }
+
+    public void stopAudioService() {
```
askirk: Change name to stop() based on proposal https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0113-audiostreammanager.md#android

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,192 @@
...
+        if (sdlInterface != null && sdlInterface.isConnected()) {
+            didRequestShutdown = true;
+            sdlInterface.stopAudioService();
+        }
+    }
+
+    public void pushAudioFile(File audioFile) {
```
askirk: Change name to `play()` based on proposal https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0113-audiostreammanager.md#android

askirk:
Could you also update `SdlManager` in this PR to accomidate the audio streaming manager? Uncomment these lines:
https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java#L78
https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java#L124
https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java#L153-L154
https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java#L166
https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java#L379-L384

And this line in the testing file:
https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/androidTest/java/com/smartdevicelink/api/SdlManagerTests.java#L161

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
kshala-ford: I don't think this is a good idea because "start" is too confusing. Does it start the audio service or the stream? the decoder? I decided to give it a more specific name (just like the video manager is doing).

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
kshala-ford: I don't think this is a good idea because "stop" is too confusing. Does it start the audio service or the stream? the decoder? I decided to give it a more specific name (just like the video manager is doing).

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
kshala-ford: The play method (with playImmediately) was proposed based on the experience with iOS that audio decoding only works through a file. The idea was to pre-decode and avoid delays. Fortunately we can live stream from an audio file which makes it irrelevant. Still the manager could benefint from some optimization (buffer limit)... working on it

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
kshala-ford: how is the lifecycle implemented for the video streaming manager? Is the manager starting the video service (and stream)?

sdl_android/src/main/java/com/smartdevicelink/api/audio/SampleBuffer.java
```diff
@@ -0,0 +1,233 @@
...
+                return avg;
+            }
+            default: {
+                Log.e(TAG, "SampleBuffer.get(int): The sample type is not known: " + sampleType);
+                return 0.0;
+            }
+        }
+    }
+
+    public void put(double sample) {
```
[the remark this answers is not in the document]
kshala-ford: wilco

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
askirk: The `SdlManager` calls the `start()` method that the video streaming manager inherits from `BaseSubManager`, providing a callback that is called when the video streaming manager changes its `BaseSubManager` state to `READY` or `ERROR` : https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/BaseSubManager.java#L53-L59

If you look at the video streaming manager itself, it uses an internal state machine  to keep track of its _streaming_ state: https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_video_streaming_manager/sdl_android/src/main/java/com/smartdevicelink/api/StreamingStateMachine.java
This gives the ability to check the current status of the stream.

The video service is only started when `startRemoteDisplayStream()`, `createOpenGLInputSurface`, or `startVideoStream` methods are used. The video streaming manager changes its `BaseSubManager` state to `READY` after the encoder is created and the necessary listeners are added:

https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_video_streaming_manager/sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java#L117

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
askirk: Understood, it probably should also have javadocs explaining how to use it.

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
askirk: Add javadoc to clarify

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
askirk: Ah, makes sense

kshala-ford:
@askirk I changed quite a bit and added more test cases. The manager does not look more like the video manager. Could you please review again? Thanks

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,271 @@
...
+     * @param encrypted Specify whether or not the audio stream should be encrypted.
+     */
+    public void startAudioStream(boolean encrypted) {
+        // audio stream cannot be started without a connected internal interface
+        if (internalInterface == null || !internalInterface.isConnected()) {
+            return;
+        }
+
+        // streaming state must be NONE (starting the service is ready. starting stream is started)
+        if (streamingStateMachine.getState() != StreamingStateMachine.NONE) {
```
askirk: I believe this logic is already built into the streamingStateMachine. It will only change state on a valid transition (NONE->READY or NONE->ERROR) so this check probably isn't necessary

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,271 @@
...
+     * Stops the audio service and audio stream to the connected device.
+     * The method is non-blocking.
+     */
+    public void stopAudioStream() {
+        if (internalInterface == null || !internalInterface.isConnected()) {
+            return;
+        }
+
+        // streaming state must be STARTED (starting the service is ready. starting stream is started)
+        if (streamingStateMachine.getState() != StreamingStateMachine.STARTED) {
```
askirk: I believe this logic is already built into the streamingStateMachine. It will only change state on a valid transition (STARTED->STOPPED) so this check isn't necessary

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -143,24 +147,30 @@ protected void initialize(){
...
-		this.audioStreamManager = new AudioStreamManager(_internalInterface);
-		this.audioStreamManager.start(subManagerListener);
+
 		*/
 	}
 
 	private void dispose() {
 		this.permissionManager.dispose();
 		this.fileManager.dispose();
+		this.audioStreamManager.dispose();
```
askirk: This needs an `if` block similar to Line 150

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,271 @@
...
+
+        streamingStateMachine.transitionToState(StreamingStateMachine.NONE);
+    }
+
+    /**
+     * Starts the audio service and audio stream to the connected device.
+     * The method is non-blocking.
+     * @param encrypted Specify whether or not the audio stream should be encrypted.
+     */
+    public void startAudioStream(boolean encrypted) {
```
theRealBrett: there should be a callback on this. Currently if theres an issue, the developer would be none the wiser, and it would negatively impact the rest of the session. Even if its the simple `CompletionListener` that we are using everywhere, it would be better it currently is

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * Pushes the specified audio file to the playback queue.
+     * The audio file will be played immediately. If another audio file is currently playing
+     * the specified file will stay queued and automatically played when ready.
+     * @param audioFile The specified audio file to be played.
+     * @param completionListener A completion listener that informs when the audio file is played.
+     */
+    public void pushAudioFile(File audioFile, final CompletionListener completionListener) {
```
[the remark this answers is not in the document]
kshala-ford: @theRealBrett Not sure if a class like SdlFile is needed here. The purpose of that class is to upload a file using PutFile. Only Uri or Resource (as an AssetFileDescriptor) are useful but all other variables are ignored. For both useful variables I need a context. Does it make sense to confuse the developer with ignored stuff? Which one of the variables should I prefer if both are set? How should I get the context object?

I think it makes more sense to method overloads (one for for resource, another for Uri). This looks to me more clear regarding API design.

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
kshala-ford: Could we instead do https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0193-update-android-min-sdk.md ? My additions to l150 were to solve the issue of min-sdk 8 but I want to avoid a nullable manager though...

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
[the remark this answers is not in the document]
theRealBrett: I guess the one thing that I liked about `SdlFile`, or more what I was going for was letting the developer know the formats we supported and trying to keep some consistency since for pretty much every other file we use in the managers. Upon inspection of iOS, I see that they simply pass in the URL of the file to be played. Do you think that would be better here? Instead of having the developer implement the methods to obtain the byte arrays of file data? I guess that will need you to obtain context as well though. We can pass context in from the `SdlManager` to obtain it though like we did in the `FileManager`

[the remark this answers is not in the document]
kshala-ford: > Instead of having the developer implement the methods to obtain the byte arrays of file data?

I don't understand this...

Because of iOS I proposed the overload with Uri. Do you consider using Uri also for resources? I would still do the method overload for a resource Id and try https://stackoverflow.com/questions/7976141/get-uri-of-mp3-file-stored-in-res-raw-folder-in-android

Anyway I need a context. I can take a look at the file manager as you suggest. I kind of prefer to ask for a context using a parameter (due to context lifecycle)... any downsides from your point of view?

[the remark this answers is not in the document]
askirk: I think Brett was suggesting that the AudioStreamManager could implement methods similar to these 

https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/FileManager.java#L360

https://github.com/smartdevicelink/sdl_android/blob/feature/issue_782_Android_Managers/sdl_android/src/main/java/com/smartdevicelink/api/FileManager.java#L384

so that the developer can simply provide the resource id or Uri of the desired file and the manager converts it to a File object. I was using this method when testing this PR:

```
private File getSampleFile(int id) {
		File file = new File(getCacheDir() + "/" + "sample");

		try {
			InputStream is = getResources().openRawResource(id);
			FileOutputStream fos = new FileOutputStream(file);
			byte[] buffer = new byte[1 << 12];
			int length;
			do {
				length = is.read(buffer);
				if (length > 0) {
					fos.write(buffer, 0, length);
				}
			} while (length > 0);

			fos.flush();
			fos.close();
			is.close();
		} catch (IOException e) {
			e.printStackTrace();
		}

		return file;
	}
```

`FileManager` takes in a `Context` in its constructor, and holds a `WeakReference` to prevent a memory leak.

kshala-ford:
@askirk @theRealBrett done. Starting and stopping audio stream works with a completion handler. Uri replaced File. A method overload also accepts resource files (int).

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -143,27 +147,35 @@ protected void initialize(){
...
-		this.audioStreamManager.start(subManagerListener);
+
 		*/
 	}
 
 	private void dispose() {
 		this.permissionManager.dispose();
 		this.fileManager.dispose();
+
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
askirk: Should this also have a null check?

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    @StringDef({SamplingRate.EIGHT_KHZ, SamplingRate.SIXTEEN_KHZ, SamplingRate.TWENTY_TWO_KHZ, SamplingRate.FOURTY_FOUR_KHX})
+    @Retention(RetentionPolicy.SOURCE)
+    public @interface SamplingRate {
+        String EIGHT_KHZ = "8KHZ";
+        String SIXTEEN_KHZ = "16KHZ";
+        String TWENTY_TWO_KHZ = "22KHZ";
+        String FOURTY_FOUR_KHX = "44KHZ";
+    }*/
```
joeygrover: Is there a reason to keep these comment blocks?

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,358 @@
...
+            Log.w(TAG, "startAudioStream called but streamingStateMachine is not in state NONE (current: " + streamingStateMachine.getState() + ")");
+            if (completionListener != null) {
+                completionListener.onComplete(false);
+            }
+            return;
+        }
+
+        AudioPassThruCapabilities capabilities = (AudioPassThruCapabilities) internalInterface.getCapability(SystemCapabilityType.PCM_STREAMING);
+
+        switch (capabilities.getSamplingRate()) {
```
joeygrover: Would probably be good to add a `null` check to `capabilities`

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
```diff
@@ -0,0 +1,226 @@
...
+    private double lastOutputSample = 0;
+
+    private double lastOutputPresentationTimeUs = 0;
+    private double lastTargetPresentationTimeUs = 0;
+
+    protected MediaExtractor extractor;
+    protected MediaCodec decoder;
+
+    protected Uri audioSource;
+    protected Context context;
```
joeygrover: private final WeakReference<Context>

joeygrover: This should be `private final WeakReference<Context>`. Even though this class should only have a short runtime, it's important to note its use in a queue which means it could hang around for an extended time and therefore shouldn't hold a direct reference to a context.

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
```diff
@@ -0,0 +1,226 @@
...
+    protected AudioDecoderListener listener;
```
joeygrover: This can be `final`

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
```diff
@@ -0,0 +1,226 @@
...
+
+    public BaseAudioDecoder(Uri audioSource, Context context, int sampleRate, @SampleType int sampleType, AudioDecoderListener listener) {
```
joeygrover: `@NonNull` should at least be added to `audioSource`, `context`

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
```diff
@@ -0,0 +1,226 @@
...
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+@SuppressWarnings("WeakerAccess")
+@RequiresApi(api = Build.VERSION_CODES.JELLY_BEAN)
+public abstract class BaseAudioDecoder {
+    private static final String TAG = AudioDecoder.class.getSimpleName();
+
+    protected int targetSampleRate;
```
joeygrover: This can be `final`

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
```diff
@@ -0,0 +1,226 @@
...
+    protected @SampleType int targetSampleType;
```
joeygrover: This can be `final`

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
```diff
@@ -0,0 +1,226 @@
...
+    protected Context context;
+    protected AudioDecoderListener listener;
+
+    public BaseAudioDecoder(Uri audioSource, Context context, int sampleRate, @SampleType int sampleType, AudioDecoderListener listener) {
+        this.audioSource = audioSource;
+        this.context = context;
+        this.listener = listener;
+
+        targetSampleRate = sampleRate;
+        targetSampleType = sampleType;
```
joeygrover: See previous comment. Javadoc should be added to address possible values as well as some error checking at this point to ensure an error does not persist downstream (eg `SampleBuffer`).

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
```diff
@@ -0,0 +1,226 @@
...
+    protected Uri audioSource;
+    protected Context context;
+    protected AudioDecoderListener listener;
+
+    public BaseAudioDecoder(Uri audioSource, Context context, int sampleRate, @SampleType int sampleType, AudioDecoderListener listener) {
+        this.audioSource = audioSource;
+        this.context = context;
+        this.listener = listener;
+
+        targetSampleRate = sampleRate;
```
joeygrover: This class essentially assumes this value will never be 0, however, that value can certainly be supplied. We should add javadoc to the constructor to inform of the issue. I would also consider adding a default value of the supplied `sampleRate` was `<=0`

sdl_android/src/main/java/com/smartdevicelink/api/audio/SampleBuffer.java
```diff
@@ -0,0 +1,280 @@
...
+ */
+public class SampleBuffer {
+    private static final String TAG = SampleBuffer.class.getSimpleName();
+
+    @SuppressWarnings({"unused", "FieldCanBeLocal"})
+    private @SampleType int sampleType;
+    private ByteBuffer byteBuffer;
+    private int channelCount;
+    private long presentationTimeUs;
+
```
joeygrover: All local vars here can be `final`

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioDecoderCompat.java
```diff
@@ -0,0 +1,131 @@
...
+            new DecodeAsync(this).execute();
+        } catch (Exception e) {
+            e.printStackTrace();
+            this.listener.onDecoderError(e);
+            this.listener.onDecoderFinish(false);
+            stop();
+        }
+    }
+
+    private static class DecodeAsync extends AsyncTask<Void, Void, Void> {
```
joeygrover: Is there a reason this is an `AsyncTask` vs a Thread and Runnable? It limits this class to be required to be instantiated on the UI thread. (See [Threading rules for AsyncTask](https://developer.android.com/reference/android/os/AsyncTask) )

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioDecoderCompat.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Creates a new object of AudioDecoder.
+     * @param audioSource The audio source to decode.
+     * @param context The context object to use to open the audio source.
+     * @param sampleRate The desired sample rate for decoded audio data.
+     * @param sampleType The desired sample type (8bit, 16bit, float).
+     * @param listener A listener who receives the decoded audio.
+     */
+    AudioDecoderCompat(Uri audioSource, Context context, int sampleRate, @SampleType int sampleType, AudioDecoderListener listener) {
```
joeygrover: `@NonNull` should at least be added to `audioSource`, `context`

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioDecoderCompat.java
```diff
@@ -0,0 +1,131 @@
...
+        protected void onPreExecute() {
+            super.onPreExecute();
+        }
+
+        @Override
+        protected void onPostExecute(Void aVoid) {
+            super.onPostExecute(aVoid);
+            AudioDecoderCompat reference = weakReference.get();
+            if (reference == null) {
+                Log.w(TAG, "Weak reference actually got null...");
```
joeygrover: Please update log string to `AudioDecorder reference was null`

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioDecoderCompat.java
```diff
@@ -0,0 +1,131 @@
...
+
+            reference.listener.onDecoderFinish(true);
+            reference.stop();
+        }
+
+        @Override
+        protected Void doInBackground(Void... voids) {
+            AudioDecoderCompat reference = weakReference.get();
+            if (reference == null) {
+                Log.w(TAG, "Weak reference actually got null...");
```
joeygrover: Please update log string to `AudioDecorder reference was null`

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,358 @@
...
+public class AudioStreamManager extends BaseSubManager {
+    private static final String TAG = AudioStreamManager.class.getSimpleName();
+    private static final int COMPLETION_TIMEOUT = 2000;
+
+    private IAudioStreamListener sdlAudioStream;
+    private int sdlSampleRate;
+    private @SampleType int sdlSampleType;
+    private final Queue<BaseAudioDecoder> queue;
+    private final WeakReference<Context> context;
+    private StreamingStateMachine streamingStateMachine;
```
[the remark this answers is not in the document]
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,358 @@
...
+
+    // This completion listener is used as a callback to the app developer when starting/stopping audio service
+    private CompletionListener serviceCompletionListener;
+    // As the internal interface does not provide timeout we need to use a future task
+    private Handler serviceCompletionHandler;
+
+    private Runnable serviceCompletionTimeoutCallback = new Runnable() {
```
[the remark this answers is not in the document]
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
```diff
@@ -0,0 +1,358 @@
...
+    private int sdlSampleRate;
+    private @SampleType int sdlSampleType;
+    private final Queue<BaseAudioDecoder> queue;
+    private final WeakReference<Context> context;
+    private StreamingStateMachine streamingStateMachine;
+
+    // This completion listener is used as a callback to the app developer when starting/stopping audio service
+    private CompletionListener serviceCompletionListener;
+    // As the internal interface does not provide timeout we need to use a future task
+    private Handler serviceCompletionHandler;
```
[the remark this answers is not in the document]
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioStreamManager.java
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/BaseAudioDecoder.java
Tnnnguyen: Done, this value will be checked in AudioStreamManager.

sdl_android/src/main/java/com/smartdevicelink/api/audio/SampleBuffer.java
Tnnnguyen: Done

sdl_android/src/main/java/com/smartdevicelink/api/audio/AudioDecoderCompat.java
Tnnnguyen: Done

joeygrover:
Closed in favor #877 (Contains fixes to comments on this branch)

PR #824, opened 2018-07-24
Screen manager

### LINKED ISSUE #782: [SDL 0171] Android Manager APIs
Proposal: [Android Manager APIs](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0171-android-manager-apis.md)

> This proposal is for syncing up the Android SDL library with the iOS library in terms of developer facing APIs. It accomplishes this by introducing a high level developer interface for the Android library through managers. It handles many complex aspects of SDL development, including managing files and setting information on the screen. It also creates a system in which other managers can be easily added in the future to handle tasks. Many aspects of this system are mirrored from the iOS manager system and are designed to be as close as possible to that API while maintaining common Java and Android paradigms.

Review: smartdevicelink/sdl_evolution#500

Steering Committee Decision:
> The Steering Committee voted to accept this proposal with revisions.  The revisions will include those outlined in [this comment](https://github.com/smartdevicelink/sdl_evolution/issues/500#issuecomment-394739127) from one of the authors of the proposal.

> The proposal .md was updated on June 6, 2018 to include the agreed upon revisions.

BrettyWhite:
Fixes #782 

### Risk

### Testing Plan
* This code snippet can be used to test the manager:
```
// Begin update transaction
proxy.getScreenManager().beginTransaction();

//proxy.getScreenManager().setDisplayLayout(PredefinedLayout.DOUBLE_GRAPHIC_WITH_SOFTBUTTONS);

// Set text fields
proxy.getScreenManager().setTextField1("It is");
proxy.getScreenManager().setTextField2("Wednesday");
proxy.getScreenManager().setTextField3("My");
proxy.getScreenManager().setTextField4("Dudes");


// Set primary graphic
proxy.getScreenManager().setPrimaryGraphic(new SdlArtwork("image1", FileType.GRAPHIC_PNG, R.drawable.image1, true));


// Create softButtonObject1
SoftButtonState softButtonState1 = new SoftButtonState("object1-state1", "it is", new SdlArtwork(ICON_FILENAME, FileType.GRAPHIC_PNG, R.mipmap.ic_launcher, false));
SoftButtonState softButtonState2 = new SoftButtonState("object1-state2", "Wed", new SdlArtwork(IMAGE1_FILENAME, FileType.GRAPHIC_PNG, R.drawable.image1, false));
SoftButtonObject softButtonObject1 = new SoftButtonObject("object1", Arrays.asList(softButtonState1, softButtonState2), softButtonState1.getName(), new SoftButtonObject.OnEventListener() {
    @Override
    public void onPress(SoftButtonObject softButtonObject, OnButtonPress onButtonPress) {
        softButtonObject.transitionToNextState();
    }

    @Override
    public void onEvent(SoftButtonObject softButtonObject, OnButtonEvent onButtonEvent) {

    }
});


// Create softButtonObject2
SoftButtonState softButtonState3 = new SoftButtonState("object2-state1", "my", new SdlArtwork(IMAGE2_FILENAME, FileType.GRAPHIC_PNG, R.drawable.image2, false));
SoftButtonState softButtonState4 = new SoftButtonState("object2-state2", "dudes!", null);
SoftButtonObject softButtonObject2 = new SoftButtonObject("object2", Arrays.asList(softButtonState3, softButtonState4), softButtonState3.getName(), null);
softButtonObject2.setOnEventListener(new SoftButtonObject.OnEventListener() {
    @Override
    public void onPress(SoftButtonObject softButtonObject, OnButtonPress onButtonPress) {
        softButtonObject.transitionToNextState();
    }

    @Override
    public void onEvent(SoftButtonObject softButtonObject, OnButtonEvent onButtonEvent) {

    }
});


// Set the soft button objects
proxy.getScreenManager().setSoftButtonObjects(Arrays.asList(softButtonObject1, softButtonObject2));


// Commit update transaction
proxy.getScreenManager().commit(new CompletionListener() {
    @Override
    public void onComplete(boolean success) {
        Log.i(TAG, "UPDATES COMPLETE: " + success);

    }
});
```

### Summary
This PR implements the `ScreenManager` as described in the proposal.
* Implemented `SoftButtonManager`
* Implemented `TextAndGraphicManager`
* Implemented `ScreenManager`
* Added test classes to test the functionalties for the managers

### CLA


sdl_android/src/main/java/com/smartdevicelink/api/datatypes/SdlFile.java
```diff
@@ -16,6 +16,29 @@
...
+        setType(fileType);
+        setUri(uri);
+        setPersistent(persistentFile);
+    }
+
+    public SdlFile(@NonNull String fileName, @NonNull FileType fileType, byte[] data, boolean persistentFile){
+        setName(fileName);
+        setType(fileType);
+        setFileData(data);
+        setPersistent(persistentFile);
```
joeygrover: Since there is no synchronization in the set methods it doesn't make sense to call them here instead of setting the value directly. The method call is an unnecessary step.

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * <strong>ScreenManager</strong> <br>
+ *
+ * Note: This class must be accessed through the SdlManager. Do not instantiate it by itself. <br>
+*/
+public class ScreenManager extends BaseSubManager {
+
+	private static String TAG = "ScreenManager";
+	private FileManager fileManager;
```
joeygrover: This should be a `WeakReference` since the original lives in the `SdlManager` and the screen manager shouldn't keep the `FileManager` from being GC'ed.

joeygrover: The reference can also be `final`

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
```diff
@@ -0,0 +1,343 @@
...
+		this.softButtonManager = new SoftButtonManager(internalInterface, fileManager);
+		this.softButtonManager.start(subManagerListener);
+		this.textAndGraphicManager = new TextAndGraphicManager(internalInterface, fileManager, softButtonManager);
+		this.textAndGraphicManager.start(subManagerListener);
+	}
+
+	/**
+	 * <p>Called when manager is being torn down</p>
+	 */
+	public void dispose(){
```
joeygrover: Shouldn't this call the submanagers' dispose methods?

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
```diff
@@ -0,0 +1,343 @@
...
+		transitionToState(SHUTDOWN);
+	}
+
+	/**
+	 * Set the textField1 on the head unit screen
+	 * @param textField1 String value represents the textField1
+	 */
+	public void setTextField1(@NonNull String textField1) {
```
joeygrover: Is there a reason this has the `@NonNull` tag while all the other fields do not? If so, it would be good to include the reason in the javadoc. Also including a note that empty set`""` will clear the field.

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
```diff
@@ -0,0 +1,343 @@
...
+		softButtonManager.setBatchUpdates(true);
+		textAndGraphicManager.setBatchUpdates(true);
+	}
+
+	/**
+	 * Send the updates that were started after beginning the transaction
+	 * @param listener a CompletionListener that has a callback that will be called when the updates are finished
+	 */
+	public void commit(final CompletionListener listener){
+		allSubManagersFinishedUpdatingSuccesslly = true;
```
joeygrover: Is there a reason this isn't a local variable? Just add `boolean allSubManagersFinishedUpdatingSuccesslly = true;` in the global scope of the new anonymous `CompletionListener` instantiation. I would also recommend changing the name as this is extremely verbose, perhaps `updateSuccessful`.

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+
+/**
+ * <strong>SoftButtonManager</strong> <br>
+ * SoftButtonManager gives the developer the ability to control how soft buttons are displayed on the head unit.<br>
+ * Note: This class must be accessed through the SdlManager->ScreenManager. Do not instantiate it by itself.<br>
+ */
+class SoftButtonManager extends BaseSubManager {
+
+    private static final String TAG = "SoftButtonManager";
+    private FileManager fileManager;
```
joeygrover: This should be a `WeakReference` since the original lives in the `SdlManager` and the soft button  manager shouldn't keep the `FileManager` from being GC'ed.

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+     */
+    private String currentMainField1;
+
+
+    /**
+     * Creates a new instance of the SoftButtonManager
+     * @param internalInterface
+     * @param fileManager
+     */
+    SoftButtonManager(ISdl internalInterface, FileManager fileManager) {
```
joeygrover: Both vars should have `@NonNull` tags

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+    private FileManager fileManager;
+    private DisplayCapabilities displayCapabilities;
+    private SoftButtonCapabilities softButtonCapabilities;
+    private List<SoftButtonObject> softButtonObjects;
+    private HMILevel currentHMILevel;
+    private Show inProgressShowRPC;
+    private CompletionListener inProgressListener, queuedUpdateListener, cachedListener;
+    private boolean hasQueuedUpdate, batchUpdates, waitingOnHMILevelUpdateToSetButtons;
+    private OnSystemCapabilityListener onSoftButtonCapabilitiesListener, onDisplayCapabilitiesListener;
+    private OnRPCNotificationListener onHMIStatusListener, onButtonPressListener, onButtonEventListener;
```
joeygrover: I believe these can all be `final`

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+    private static final String TAG = "SoftButtonManager";
+    private FileManager fileManager;
+    private DisplayCapabilities displayCapabilities;
+    private SoftButtonCapabilities softButtonCapabilities;
+    private List<SoftButtonObject> softButtonObjects;
+    private HMILevel currentHMILevel;
+    private Show inProgressShowRPC;
+    private CompletionListener inProgressListener, queuedUpdateListener, cachedListener;
+    private boolean hasQueuedUpdate, batchUpdates, waitingOnHMILevelUpdateToSetButtons;
+    private OnSystemCapabilityListener onSoftButtonCapabilitiesListener, onDisplayCapabilitiesListener;
```
joeygrover: I believe these (OnSystemCapabilityListener) can all be final

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+ * SoftButtonManager gives the developer the ability to control how soft buttons are displayed on the head unit.<br>
+ * Note: This class must be accessed through the SdlManager->ScreenManager. Do not instantiate it by itself.<br>
+ */
+class SoftButtonManager extends BaseSubManager {
+
+    private static final String TAG = "SoftButtonManager";
+    private FileManager fileManager;
+    private DisplayCapabilities displayCapabilities;
+    private SoftButtonCapabilities softButtonCapabilities;
+    private List<SoftButtonObject> softButtonObjects;
```
joeygrover: This list is currently not thread safe. If a soft button object gets removed/added during one of the iteration processes it should through a ConcurrentModificationException. I would suggest using a `CopyOnWriteArrayList`

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+        // This is necessary because there may be no images needed to be uploaded
+        update(cachedListener);
+    }
+
+    /**
+     * Update the SoftButtonManger by sending a new Show RPC to reflect the changes
+     * @param listener a CompletionListener
+     */
+    protected void update(CompletionListener listener) {
+        cachedListener = listener;
```
joeygrover: This will get set to `null` sometimes, is that acceptable? For instance, out of the SoftButonObject.

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+
+
+        // Send Show RPC with soft buttons representing the current state for the soft button objects
+        inProgressListener = listener;
+        inProgressShowRPC = new Show();
+        inProgressShowRPC.setMainField1(getCurrentMainField1());
+        if (softButtonObjects == null) {
+            Log.d(TAG, "Soft button objects are null, sending an empty array");
+            inProgressShowRPC.setSoftButtons(new ArrayList<SoftButton>());
+        } else if ((currentStateHasImages() && !allCurrentStateImagesAreUploaded()) && (softButtonCapabilities == null || !softButtonCapabilities.getImageSupported())) {
```
joeygrover: Shouldn't this if be an OR instead of an AND?
`else if ((currentStateHasImages() && !allCurrentStateImagesAreUploaded()) || (softButtonCapabilities == null || !softButtonCapabilities.getImageSupported())) { `

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+            inProgressShowRPC.setSoftButtons(createSoftButtonsForCurrentState());
+        }
+
+        inProgressShowRPC.setOnRPCResponseListener(new OnRPCResponseListener() {
+            @Override
+            public void onResponse(int correlationId, RPCResponse response) {
+                Log.i(TAG, "Soft button update completed");
+
+                inProgressShowRPC = null;
+                CompletionListener currentListener;
```
joeygrover: Move this into the next `if` statement

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+                }
+
+
+                if (hasQueuedUpdate) {
+                    Log.d(TAG, "Queued update exists, sending another update");
+                    currentListener = queuedUpdateListener;
+                    queuedUpdateListener = null;
+                    hasQueuedUpdate = false;
+                    update(currentListener);
+                }
```
joeygrover: Code can be condensed into: 

```
        inProgressShowRPC.setOnRPCResponseListener(new OnRPCResponseListener() {
            @Override
                Log.i(TAG, "Soft button update completed");
                handleResponse(true);
            }

            @Override
            public void onError(int correlationId, Result resultCode, String info) {
                super.onError(correlationId, resultCode, info);

                Log.e(TAG, "Soft button update error");
                handleResponse(false);

            }
            
            public void handleResponse(boolean success){

                inProgressShowRPC = null;
                CompletionListener currentListener;
                if (inProgressListener != null) {
                    currentListener = inProgressListener;
                    inProgressListener = null;
                    currentListener.onComplete(success);
                }


                if (hasQueuedUpdate) {
                    Log.d(TAG, "Queued update exists, sending another update");
                    currentListener = queuedUpdateListener;
                    queuedUpdateListener = null;
                    hasQueuedUpdate = false;
                    update(currentListener);
                }
            }
        });
```

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonObject.java
```diff
@@ -0,0 +1,261 @@
...
+ */
+public class SoftButtonObject {
+
+    private static final String TAG = "SoftButtonObject";
+    private String name;
+    private List<SoftButtonState> states;
+    private String currentStateName;
+    private int buttonId;
+    private OnEventListener onEventListener;
+    private SoftButtonManager softButtonManager;
```
joeygrover: This seems excessive for a single method called and creates a circular logic. I would propose introducing an `UpdateListener` interface in this object, making a a `final` instantiation in the SoftButtonManager and passing that into each object instead.

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonState.java
```diff
@@ -0,0 +1,96 @@
...
+ * The states of SoftButtonObject allow the developer to not have to manage multiple SoftButtons that have very similar functionality.<br>
+ * For example, a repeat button in a music app can be thought of as one SoftButtonObject with three typical states: repeat off, repeat 1, and repeat on.<br>
+ * @see SoftButtonObject
+ */
+public class SoftButtonState {
+
+    private static final String TAG = "SoftButtonState";
+    private String name;
+    private SdlArtwork artwork;
+    private SoftButton softButton;
```
joeygrover: This can be `final`

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+ * Note: This class must be accessed through the SdlManager. Do not instantiate it by itself. <br>
+ *
+ */
+class TextAndGraphicManager extends BaseSubManager {
+
+	private static final String TAG = "TextAndGraphicManager";
+	private static final String MAIN_FIELD_1 = "mainField1";
+	private static final String MAIN_FIELD_2 = "mainField2";
+	private static final String MAIN_FIELD_3 = "mainField3";
+	private static final String MAIN_FIELD_4 = "mainField4";
```
joeygrover: These are redundant. The enum `TextFieldName` should be used in their place.

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+			return;
+		}
+
+		if (isDirty){
+			isDirty = false;
+			sdl_update(listener);
+		}
+	}
+	
+	private void sdl_update(CompletionListener listener){
```
joeygrover: Please rename to use proper naming conventions. `sdlUpdate` or `internalUpdate`

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+		}
+
+		Log.v(TAG, "Updating Text and Graphics");
+		if (inProgressUpdate != null){
+
+			Log.v(TAG, "In progress update exists, queueing update");
+			if (queuedUpdateListener != null){
+
+				Log.v(TAG, "Queued update already exists, superseding previous queued update");
+				queuedUpdateListener.onComplete(false);
```
joeygrover: Please comment out these logs or change them to simple comments.

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+			Log.v(TAG, "Trying to send show on HMI_NONE, waiting for full");
+			pendingHMIFull = true;
+			if (listener != null){
+				pendingHMIListener = listener;
+			}
+			return;
+		}
+
+		Log.v(TAG, "Updating Text and Graphics");
+		if (inProgressUpdate != null){
```
joeygrover: This variable seems to be updated from a few different methods. I would recommend some type of synchronization; adding the `volatile` keyword to the variable declaration should suffice.

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+        }
+
+        // Don't send if we're in HMI NONE
+        if (currentHMILevel == null || currentHMILevel == HMILevel.HMI_NONE) {
+            return;
+        }
+
+        Log.v(TAG, "Updating soft buttons");
+
+        cachedListener = null;
```
joeygrover: So this actually always gets set to null, what's the point in setting it to begin with in this method?

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
```diff
@@ -0,0 +1,549 @@
...
+    protected void update(CompletionListener listener) {
+        cachedListener = listener;
+
+        if (batchUpdates) {
+            return;
+        }
+
+        // Don't send if we're in HMI NONE
+        if (currentHMILevel == null || currentHMILevel == HMILevel.HMI_NONE) {
+            return;
```
joeygrover: Shouldn't the callback listener be called at this point?

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
joeygrover: I believe this should also only be set before the next two `return` statements.

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+					inProgressListener.onComplete(true);
+					inProgressListener = null;
+				}
+
+				if (hasQueuedUpdate){
+					Log.v(TAG, "Queued update exists, sending another update");
+					sdl_update(queuedUpdateListener);
+					queuedUpdateListener = null;
+					hasQueuedUpdate = false;
+				}
```
joeygrover: These two value writes will clear out whatever was set in the sdl_update method. I would suggest moving these into the `sdl_update` method instead after the first few checks or create a temp value for the queuedUpdateListener and moving the write logic before the `sdl_update` call. 

```
hasQueuedUpdate = false;
CompletionListener temp = queuedUpdateListener;
queuedUpdateListener = null;
sdl_update(temp);
```

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+
+	private Show assembleShowImages(Show show){
+
+		if (!shouldUpdatePrimaryImage() && ! shouldUpdateSecondaryImage()){
+			return show;
+		}
+
+		if (shouldUpdatePrimaryImage()){
+			Image primaryImage = new Image();
+			primaryImage.setImageType(ImageType.DYNAMIC);
```
joeygrover: How does a developer use a STATIC image?

sdl_android/src/main/java/com/smartdevicelink/api/datatypes/SdlArtwork.java
```diff
@@ -8,6 +9,20 @@
  * A class that extends SdlFile, representing artwork (JPEG, PNG, or BMP) to be uploaded to core
  */
 public class SdlArtwork extends SdlFile {
```
joeygrover: There is going to need to be a way to support static icons

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+			}
+
+			if (listener != null){
+				queuedUpdateListener = listener;
+			}else{
+				hasQueuedUpdate = true;
+			}
+			return;
+		}
+
```
joeygrover: This is the point where the synchronization block should start until the end of the method

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+				@Override
+				public void onComplete(boolean success) {
+					if (!success){
+						Log.e(TAG, "Error uploading image");
+						inProgressUpdate = extractTextFromShow(inProgressUpdate);
+						sendShow();
+					}
+					// Check if queued image update still matches our images (there could have been a new Show in the meantime)
+					// and send a new update if it does. Since the images will already be on the head unit, the whole show will be sent
+					if ((thisUpdate.getGraphic() != null && queuedImageUpdate.getGraphic() != null) && thisUpdate.getGraphic().equals(queuedImageUpdate.getGraphic()) ||
```
joeygrover: No need for second null check. Logic check can be as followed: 
`thisUpdate.getGraphic() != null && thisUpdate.getGraphic().equals(queuedImageUpdate.getGraphic())`

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+							(thisUpdate.getSecondaryGraphic() != null && queuedImageUpdate.getSecondaryGraphic() != null) && thisUpdate.getSecondaryGraphic().equals(queuedImageUpdate.getSecondaryGraphic())){
+						Log.v(TAG, "Queued image update matches the images we need, sending update");
+						sendShow();
+					} else {
```
joeygrover: else can be removed in favor of a comment

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+					Log.v(TAG, "Successfully uploaded Artworks");
+					listener.onComplete(true);
+				}
+			}
+		});
+	}
+
+	private Show assembleShowImages(Show show){
+
+		if (!shouldUpdatePrimaryImage() && ! shouldUpdateSecondaryImage()){
```
joeygrover: Remove unnecessary if statement. The next two if statements will take care of adding the image to the show or not and if not the original `show` will be returned

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+		}
+
+		List<String> nonNullFields = findNonNullMainTextFields();
+		if (nonNullFields.size() == 0){
+			return show;
+		}
+
+		int numberOfLines = getNumberOfLines();
+
+		if (numberOfLines == 1){
```
joeygrover: This should be a switch statement

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+	// Text
+
+	protected Show assembleShowText(Show show){
+
+		show = setBlankTextFields(show);
+
+		if (mediaTrackTextField != null){
+			show.setMediaTrack(mediaTrackTextField);
+		} else {
+			show.setMediaTrack("");
```
joeygrover: This is unnecessary as the previous method call will sets this field to empty set

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+		newShow.setMainField1("");
+		newShow.setMainField2("");
+		newShow.setMainField3("");
+		newShow.setMainField4("");
+		newShow.setMediaTrack("");
+
+		return newShow;
+	}
+
+	private void updateCurrentScreenDataFromShow(Show show){
```
joeygrover: Please refactor to `updateCurrentScreenDataState`

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,893 @@
...
+			currentScreenData.setGraphic(show.getGraphic());
+		}
+		if (show.getSecondaryGraphic() != null){
+			currentScreenData.setSecondaryGraphic(show.getSecondaryGraphic());
+		}
+	}
+
+	// Helpers
+
+	private List<String> findNonNullMainTextFields(){
```
joeygrover: Since this actually returns if the values are not null along with if they are beyond empty set I would recommend changing the name to `findValidMainTextFields`

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
```diff
@@ -0,0 +1,343 @@
...
+		this.softButtonManager.start(subManagerListener);
+		this.textAndGraphicManager = new TextAndGraphicManager(internalInterface, fileManager, softButtonManager);
+		this.textAndGraphicManager.start(subManagerListener);
+	}
+
+	/**
+	 * <p>Called when manager is being torn down</p>
+	 */
+	public void dispose(){
+		transitionToState(SHUTDOWN);
```
joeygrover: if super.dispose() is called this will automatically be handled by the parent class

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
bilal-alsharifi: done

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
bilal-alsharifi: done

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
BrettyWhite: I added the `synchronized` modifier to the entire method as described in a previous comment

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
bilal-alsharifi: I don't see why this could be an issue. When the SoftButtonObject calls update(null), that shouldn't trigger the callback (same behavior in iOS)

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonManager.java
bilal-alsharifi: Yeah, that is right. Fixed that and also created a new private method to check if softButtonImages are supported

sdl_android/src/main/java/com/smartdevicelink/api/screen/SoftButtonObject.java
bilal-alsharifi: Agree. I made the changes. Let me know if the naming is good or need to be changed.

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
```diff
@@ -0,0 +1,343 @@
...
+	private SoftButtonManager softButtonManager;
+	private TextAndGraphicManager textAndGraphicManager;
+	private boolean allSubManagersFinishedUpdatingSuccesslly;
+
+	// Sub manager listener
+	private final CompletionListener subManagerListener = new CompletionListener() {
+		@Override
+		public synchronized void onComplete(boolean success) {
+			if(!success){
+				Log.d(TAG, "Sub manager failed to initialize");
```
joeygrover: Should there be a check against if the submanagers failed to initialize so that the screen manager transitions to error state?

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,890 @@
...
+	private OnRPCNotificationListener hmiListener;
+	private OnSystemCapabilityListener onDisplayCapabilitiesListener;
+	private SdlArtwork primaryGraphic, secondaryGraphic;
+	private TextAlignment textAlignment;
+	private String textField1, textField2, textField3, textField4, mediaTrackTextField;
+	private MetadataType textField1Type, textField2Type, textField3Type, textField4Type;
+
+	//Constructors
+
+	TextAndGraphicManager(ISdl internalInterface, FileManager fileManager, SoftButtonManager softButtonManager) {
```
joeygrover: All vars should be considered @nonnull right?

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
```diff
@@ -0,0 +1,890 @@
...
+			return 4;
+		}
+
+		int linesFound = 0;
+
+		List<TextField> textFields = displayCapabilities.getTextFields();
+
+		for (TextField field : textFields) {
+			if (field.getName() != null) {
+				TextFieldName name = field.getName();
```
joeygrover: Move the declaration for `TexFieldName name;` before the `for` loop and just set the name through each iteration rather than create a new reference.

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
```diff
@@ -0,0 +1,343 @@
...
+			}
+			if(
+					softButtonManager != null && softButtonManager.getState() != BaseSubManager.SETTING_UP &&
+					textAndGraphicManager != null && textAndGraphicManager.getState() != BaseSubManager.SETTING_UP
+					){
+				transitionToState(READY);
```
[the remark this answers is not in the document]
BrettyWhite: fixed

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
BrettyWhite: fixed

sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java
BrettyWhite: now it sets the screen manager state to ERROR

BrettyWhite: The start methods are used by both:
https://github.com/smartdevicelink/sdl_android/blob/feature/screen_manager/sdl_android/src/main/java/com/smartdevicelink/api/screen/ScreenManager.java#L55

bilal-alsharifi: The only place that I can think of for switching to state ERROR is that when they cannot retrieve capabilities.

joeygrover: The start method is called, but it is not overridden (not used) in either submanager so they will always return immediately.

sdl_android/src/main/java/com/smartdevicelink/api/BaseSubManager.java
```diff
@@ -18,7 +18,7 @@
 	// states - if this gets more complicated we can move elsewhere
 	private int state;
 	private final Object STATE_LOCK = new Object();
-	public static final int SETTING_UP = 0x00, READY = 0x30, SHUTDOWN = 0x60, ERROR = 0x90;
+	public static final int SETTING_UP = 0x00, READY = 0x30, SHUTDOWN = 0x60, ERROR = 0x90, LIMITED = 0xC0;
```
joeygrover: Let's change this as follows:
```
public static final int SETTING_UP = 0x00, READY = 0x30, LIMITED = 0x50, SHUTDOWN = 0x80, ERROR = 0xC0;
```

sdl_android/src/main/java/com/smartdevicelink/api/datatypes/SdlArtwork.java
joeygrover: Per internal discussions, it is deemed this will be added in a future release

sdl_android/src/main/java/com/smartdevicelink/api/screen/TextAndGraphicManager.java
joeygrover: Per internal discussions, it is deemed this will be added in a future release

PR #826, opened 2018-07-25
Video streaming manager implementation

askirk:
Partial fix to  #782 

### Risk

### Testing Plan
Unit tests in VideoStreamingManagerTests file, also via manual testing internally.

### Summary
Adds described functionality of VideoStreamingManager from Android Manager API proposal.


sdl_android/src/main/java/com/smartdevicelink/proxy/interfaces/ISdl.java
```diff
@@ -176,4 +176,10 @@
      */
     SdlMsgVersion getSdlMsgVersion();
 
+    /**
+     * Get wiProVersion
+     * @return byte representing wiProVersion
+     */
+    byte getWiProVersion();
```
joeygrover: We should not continue the misleading naming of the protocol layer. This method should be `Version getProtocolVersion`

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -141,28 +141,29 @@ protected void initialize(){
 		this.permissionManager = new PermissionManager(_internalInterface);
 		this.permissionManager.start(subManagerListener);
 
+		this.videoStreamingManager = new VideoStreamingManager(_internalInterface);
+		this.videoStreamingManager.start(subManagerListener);
```
joeygrover: I'm wondering if it makes sense to create the video streaming manager for every app. It seems as if the most responsible way to handle this would be to only create and start the manager when requested by the developer.

theRealBrett: Or we could do it based on AppType. If its nav or proj, then we could start it. same w/ the audiostreammanager

joeygrover: Yea that's a good idea. Though I'm not sure if audio streaming is limited to those two app types, something need to check into

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+import java.lang.ref.WeakReference;
+import java.util.List;
+import java.util.concurrent.FutureTask;
+
+@TargetApi(19)
+public class VideoStreamingManager extends BaseSubManager{
+	private static String TAG = "VideoStreamingManager";
+
+	private WeakReference<Context> context;
+	private volatile VirtualDisplayEncoder encoder;
```
joeygrover: Since there are two encoders of different types in this class, the names should be more specific to their types.

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+			stateMachine.transitionToState(StreamingStateMachine.ERROR);
+		}
+		return surface;
+	}
+
+	/**
+	 * Starts the MediaCodec encoder utilized in conjunction with the Surface returned via the createOpenGLInputSurface method
+	 * @see #createOpenGLInputSurface(int, int, int, int, int, boolean)
+	 */
+	public void startEncoder(){
```
joeygrover: It seems that these are exposed to the developer and they are responsible for handling them. I would assume the manager would take care of this instead. I'm wondering if allowing this is even useful for developers since they could create an `SdlRemoteDisplay` that contains a surface view on which they could draw.

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+			if(SessionType.NAV.equals(type)){
+				stateMachine.transitionToState(StreamingStateMachine.NONE);
+				if(remoteDisplay!=null){
+					stopStreaming();
+				}
+			}
+		}
+
+		@Override
+		public void onServiceError(SdlSession session, SessionType type, String reason) {}
```
joeygrover: Should transition to ERROR state

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+	};
+
+	private final OnRPCNotificationListener hmiListener = new OnRPCNotificationListener() {
+		@Override
+		public void onNotified(RPCNotification notification) {
+			hmiLevel = ((OnHMIStatus)notification).getHmiLevel();
```
joeygrover: Potential NPE

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+	 * that will be used to negotiate
+	 *
+	 * @param parameters  Video streaming parameters including: codec which will be used for streaming (currently, only
+	 *                    VideoStreamingCodec.H264 is accepted), height and width of the video in pixels.
+	 * @param encrypted Specify true if packets on this service have to be encrypted
+	 *
+	 * @return IVideoStreamListener interface if service is opened successfully and streaming is
+	 *         started, null otherwise
+	 */
+	public IVideoStreamListener startVideoStream(VideoStreamingParameters parameters, boolean encrypted){
```
joeygrover: Is there a reason this is `public`? I believe this is better `protected` at least.

joeygrover: The naming between `startVideoStream` and `startVideoStreaming` is very confusing. We should provide clearer naming conventions here.

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+			stateMachine.transitionToState(StreamingStateMachine.ERROR);
+			e.printStackTrace();
+		}
+		Log.d(TAG, parameters.toString());
+	}
+
+	/**
+	 * Stops streaming service and remote display encoder if applicable.
+	 */
+	public void stopStreaming(){
```
joeygrover: Is there a reason this is `public`?

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+		stateMachine.transitionToState(StreamingStateMachine.NONE);
+	}
+
+	// PUBLIC METHODS FOR CHECKING STATE
+
+	/**
+	 * Check if a video service is currently active
+	 * @return boolean (true = active, false = inactive)
+	 */
+	public boolean isVideoConnected(){
```
joeygrover: I'm confused by the return here since it will return true even if the state is `STOPPED`. I think the naming could be better here as well, since this is already the video streaming manager we don't need to preface everything with video, could just be `isStreaming` or `isActive` but it will depend on what info it's intending to return.

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
```diff
@@ -0,0 +1,571 @@
...
+		return (stateMachine.getState() == StreamingStateMachine.READY) ||
+				(stateMachine.getState() == StreamingStateMachine.STARTED) ||
+				(stateMachine.getState() == StreamingStateMachine.STOPPED);
+	}
+
+	/**
+	 * Check if video streaming has been paused due to app moving to background
+	 * @return boolean (true = not paused, false = paused)
+	 */
+	public boolean isVideoStreamingPaused(){
```
joeygrover: See previous comment on naming; `isPaused`, `isStreamingPaused`, etc

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
askirk: Maybe I should change the name to `isActive()`. I intended this for this to give developer an idea of whether the video service was started / is running successfully

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
askirk: Should I just call this internally in `createOpenGLInputSurface` after creating the surface then?

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
askirk: To give a way to stop streaming, I suppose `dispose` does the same thing, so this could be `private`

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
askirk: Will change to `isPaused`

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
askirk: Will change to `protected`, make the other method `startRemoteDisplayStreaming`

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
joeygrover: As discussed offline, if we remove the `createOpenGLInputSurface` functionality for this release these methods can also be removed. We will instruct developers to create a remote display with a surface view for now.

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
joeygrover: I believe `startVideoStream` is more related to starting the video service so I think maybe it's better named according to that. `startVideoService`. Correct me if I'm wrong though.

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
joeygrover: Ok, no I think that is smart. If stopped is called, the manager has to be able to restart properly without being recreated and as long as that is possible a `stopStreaming` method would be useful for developers. I would add javadoc to this method that explains what steps (at a high level) would need to be done to restart the stream.

sdl_android/src/main/java/com/smartdevicelink/api/VideoStreamingManager.java
joeygrover: So there are a few different states and I think we just need to define what is important for the developer to know: 

1. Video manager is ready to be used
2. Video service has been started
3. Streaming has started and is active

So what does active mean? Maybe we have different methods to define those states. I'm open for suggestions here.

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -1713,7 +1718,7 @@ private void dispatchIncomingMessage(ProtocolMessage message) {
 		}
 	}
 	
-	private byte getWiProVersion() {
+	public byte getWiProVersion() {
```
joeygrover: This should be changed to protocol version as well

sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -354,6 +354,11 @@ public SdlMsgVersion getSdlMsgVersion(){
 			return null;
 		}
 
+		@Override
+		public com.smartdevicelink.util.Version getProtocolVersion() {
+			return new com.smartdevicelink.util.Version(Byte.toString(SdlProxyBase.this.getWiProVersion()) + ".0.0");
```
joeygrover: I believe we can access this information directly from wiproprotocol and should not just use the major byte as previously used. https://github.com/smartdevicelink/sdl_android/blob/master/sdl_android/src/main/java/com/smartdevicelink/protocol/WiProProtocol.java#L96

PR #837, opened 2018-08-16
Add ManagerListener with way to notify of disposal

askirk:
- SdlManager now takes in a ManagerListener with callbacks for start up and destroy
- Destroy callback is called in dispose() method
- Gives a way for SdlManager to “restart”

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/api/ManagerListener.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/api/ManagerListener.java
@@ -1,0 +1,15 @@
+package com.smartdevicelink.api;
+
+public interface ManagerListener {
+
+	/**
+	 * Called when a manager is ready for use or failed setup
+	 * @param success - success or fail
+	 */
+	void onStart(boolean success);
+
+	/**
+	 * Called when the manager is destroyed
+	 */
+	void onDestroy();
+}
--- a/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
@@ -64,7 +64,7 @@ public class SdlManager{
 	private Vector<TTSChunk> ttsChunks;
 	private TemplateColorScheme dayColorScheme, nightColorScheme;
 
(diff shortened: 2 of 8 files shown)


sdl_android/src/main/java/com/smartdevicelink/api/ManagerListener.java
```diff
@@ -0,0 +1,15 @@
+package com.smartdevicelink.api;
+
+public interface ManagerListener {
```
joeygrover: This is a bit ambiguous in terms of naming convention. Is the intention to use this in other managers as well? If not, it should either be renamed to `SdlManagerListener` / `ISdlManager`, or be moved into the `SdlManager` class and just be titled `Listener`/`Callback` etc.

sdl_android/src/main/java/com/smartdevicelink/api/ManagerListener.java
```diff
@@ -0,0 +1,15 @@
...
+
+	/**
+	 * Called when a manager is ready for use or failed setup
+	 * @param success - success or fail
+	 */
+	void onStart(boolean success);
+
```
joeygrover: Might as well add an `onError(String info, Exception e)` method

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -497,26 +500,41 @@ public void sendRPCs(List<? extends RPCMessage> rpcs, final OnMultipleRequestLis
...
+		proxy.addOnRPCNotificationListener(FunctionID.ON_HMI_STATUS,listener);
+	}
+
+	/**
+	 * Remove an OnRPCNotificationListener for HMI status notifications
+	 * @param listener listener that was previously added for the HMI status notifications
+	 */
+	public void removeOnHmiStatusListener(OnRPCNotificationListener listener){
+		proxy.removeOnRPCNotificationListener(FunctionID.ON_HMI_STATUS, listener);
+	}
```
joeygrover: Are these supposed to be here? They seem unrelated.

PR #838, opened 2018-08-16
Add ManagerListener with way to notify of disposal

askirk:
-SdlManager now takes in a ManagerListener with callbacks for start up and destroy
-Destroy callback is called in dispose() method
-Gives a way for SdlManager to “restart”
-Adds HMI listener methods to let developer know when to start video or audio stream

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
@@ -64,7 +64,7 @@ public class SdlManager{
 	private Vector<TTSChunk> ttsChunks;
 	private TemplateColorScheme dayColorScheme, nightColorScheme;
 
-	private CompletionListener initListener;
+	private ManagerListener managerListener;
 	private int state = -1;
 	//public LockScreenConfig lockScreenConfig;
 
@@ -109,7 +109,7 @@ public class SdlManager{
 		}
 	});
 
-    // Sub manager listener
+	// Sub manager listener
 	private final CompletionListener subManagerListener = new CompletionListener() {
 		@Override
 		public synchronized void onComplete(boolean success) {
@@ -118,7 +118,7 @@ public class SdlManager{
 			}
 			if(
 					permissionManager != null && permissionManager.getState() != BaseSubManager.SETTING_UP &&
-					fileManager != null && fileManager.getState() != BaseSubManager.SETTING_UP
+							fileManager != null && fileManager.getState() != BaseSubManager.SETTING_UP
 					/*
 					videoStreamingManager != null && videoStreamingManager.getState() != BaseSubManager.SETTING_UP &&
 					audioStreamManager != null && audioStreamManager.getState() != BaseSubManager.SETTING_UP &&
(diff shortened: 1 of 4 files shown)


sdl_android/src/main/java/com/smartdevicelink/api/SdlManagerListener.java
```diff
@@ -0,0 +1,22 @@
+package com.smartdevicelink.api;
+
+public interface SdlManagerListener {
+
+	/**
+	 * Called when a manager is ready for use or failed setup
+	 * @param success - success or fail
+	 */
+	void onStart(boolean success);
```
joeygrover: I think we should leave onStart as always a successful start. If there is an error use the `onError` method. I believe this param was just a leftover from using the generic completion listener.

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -503,20 +506,19 @@ public void sendRPCs(List<? extends RPCMessage> rpcs, final OnMultipleRequestLis
...
+		managerListener = listener;
 		if (proxy == null) {
 			try {
 				proxy = new SdlProxyBase(proxyBridge, appName, shortAppName, isMediaApp, hmiLanguage,
 						hmiLanguage, hmiTypes, appId, transport, vrSynonyms, ttsChunks, dayColorScheme,
 						nightColorScheme) {
 				};
 			} catch (SdlException e) {
-				listener.onComplete(false);
+				listener.onStart(false);
```
joeygrover: I believe this should be `listener.onError("Unable to start manager", e)`

sdl_android/src/main/java/com/smartdevicelink/api/SdlManager.java
```diff
@@ -503,20 +506,19 @@ public void sendRPCs(List<? extends RPCMessage> rpcs, final OnMultipleRequestLis
...
 	/**
 	 * Starts up a SdlManager, and calls provided callback called once all BaseSubManagers are done setting up
-	 * @param listener CompletionListener that is called once the SdlManager state transitions
-	 * from SETTING_UP to READY or ERROR
+	 * @param listener ManagerListener that is called when the SdlManager is ready / failed to start, or is destroyed
 	 */
 	@SuppressWarnings("unchecked")
-	public void start(@NonNull CompletionListener listener){
-		initListener = listener;
+	public void start(@NonNull SdlManagerListener listener){
```
joeygrover: Since this listener is more than a simple completion listener should we add this to the builder instead?

theRealBrett: We could additionally have `onReady` and `onStart`

joeygrover: I understand the original purpose of putting this listener in the `start` method, but we have expanded the role of the listener so it no longer makes sense to be added during the singular process of start. The developer does not call `proxy.start()` so there is no difference between started and ready in the listeners context, the developer should only be notified when the manager can be used.  The `onStart` naming would be to follow closer to Android callbacks (Activity, Service, etc), but I'm fine with `onReady` instead.

PR #855, opened 2018-09-04
Issue 854 - Fix an issue in isEmulator

### LINKED ISSUE #854: DeviceUtil.isEmulator() returns false on Android Emulator
### Bug Report
Some unit tests fail when they are run locally on newer Android Emulators. The issue is happening because `com.smartdevicelink.test.util.DeviceUtil.isEmulator()` is returning `false` when it is supposed to return `true` on the emulator.

##### Reproduction Steps
1. Create a new Android Emulator
2. Run `com.smartdevicelink.SdlConnection.SdlConnectionTest` 

##### Expected Behavior
Tests should pass

##### Observed Behavior
Tests fail

bilal-alsharifi:
Fixes #854 

### Risk

### Testing Plan
1. Create a new Android Emulator
2. Run `com.smartdevicelink.SdlConnection.SdlConnectionTest` on the emulator
3. All tests should pass

### Summary
Fixed an issue in `isEmulator()` method where it returns false on new emulators while it is supposed to return `true`

### CLA

### DIFF
--- a/sdl_android/src/androidTest/java/com/smartdevicelink/test/util/DeviceUtil.java
+++ b/sdl_android/src/androidTest/java/com/smartdevicelink/test/util/DeviceUtil.java
@@ -9,6 +9,7 @@ public class DeviceUtil {
                 || Build.MODEL.contains("google_sdk")
                 || Build.MODEL.contains("Emulator")
                 || Build.MODEL.contains("Android SDK built for")
+                || Build.MANUFACTURER.contains("Genymotion")
                 || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic"))
                 || (Build.BRAND.startsWith("Android") && Build.DEVICE.startsWith("generic"))
                 || (Build.PRODUCT != null && Build.PRODUCT.startsWith("sdk_google_phone"))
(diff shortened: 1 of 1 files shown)


Tnnnguyen:
@bilal-alsharifi 
Also this line will never return true:
```java
Build.FINGERPRINT.startsWith("generic")
```
All the Build.FINGERPRINT's I have seen are in the form of
`
"google/sdk_gphone_.../generic_..."
`

It should be
```java
Build.FINGERPRINT.contains("generic")
```

The same is probably true for this line: (use "contains" instead of "startsWith")
```java
Build.FINGERPRINT.startsWith("unknown")
``` 
But I am not sure about "unknown"

bilal-alsharifi:
@Tnnnguyen In that case, this condition will return true:
`Build.MODEL.contains("Android SDK built for")`

So for example, when I test it with my emulator, I get this output:
```
Build.FINGERPRINT: google/sdk_gphone_x86/generic_x86:8.0.0/...
Build.MODEL: Android SDK built for x86
Build.BRANDT: google
Build.PRODUCT: sdk_gphone_x86
```

Tnnnguyen:
@bilal-alsharifi 
Yeah, I agree the check `Build.MODEL: Android SDK built for x86` will catch it.  But the first statement was written incorrectly, so it should be fixed.  Also if the first statement was written correctly, the code would return true right away; instead, it would have to make 4 additional executions.

bilal-alsharifi:
@Tnnnguyen You are right. I am just worried that there is a case where `generic` exists in the fingerprint of an actual device (not an emulator). In that case, the method will return `true` even if the device is not an emulator.
The reason why I am trying not to change `isEmulator()` method, is because that it is kind of standard way to check if a device is an emulator and it is well tested by the Android dev community on Stack Overflow.

PR #863, opened 2018-09-07
Feature/issue 782 - Fix broadcast issue in SdlProxyBase

bilal-alsharifi:

### Risk

### Summary
Because `_proxyListener` is no longer an instance of `Service` after adding the managers layer, the method `sendBroadcastIntent` in `SdlProxyBase` is always returning without actually sending a broadcast:

https://github.com/smartdevicelink/sdl_android/blob/42ef06bb9a212371abf778a8f6d4f5ba255a98a0/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L896

Also the `getService()` method will always return `null` for the same reason:
https://github.com/smartdevicelink/sdl_android/blob/42ef06bb9a212371abf778a8f6d4f5ba255a98a0/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L867

### Changelog
* Pass context from`SdlManager` to `SdlProxyBase` to use it directly to send broadcast instead of getting the context from the service
* Deprecate `setAppService()` & `getAppService()` methods in `SdlSecurityBase`
* Add `setContext()` & `getContext()` methods to `SdlSecurityBase`

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import android.app.Service;
+import android.content.Context;
 
 import com.smartdevicelink.SdlConnection.SdlSession;
 import com.smartdevicelink.protocol.enums.SessionType;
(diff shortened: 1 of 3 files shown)


sdl_android/src/main/java/com/smartdevicelink/security/SdlSecurityBase.java
```diff
@@ -84,12 +86,22 @@ public void setAppId(String val) {
...
     public static Service getAppService() {
-    	return appService;
+        return appService;
     }
-    
+
+    @Deprecated
     public static void setAppService(Service val) {
-    	appService = val;
+        appService = val;
```
joeygrover: I would also check if `appService != null` and if true, set `context` from the service's context

BrettyWhite: added it

PR #881, opened 2018-10-01
Improve manager states - Screen & Permission & File & Lockscreen & Sdl

bilal-alsharifi:

### Summary
This PR redefines states changes  for:

- SdlManager
- ScreenManager
- LockScreenManager
- FileManager
- PermissionManager 

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
@@ -32,12 +32,12 @@ public abstract class BaseSubManager {
 
 	/**
 	 * Starts up a BaseSubManager, and calls provided callback once BaseSubManager is done setting up or failed setup.
-	 * @param listener CompletionListener that is called once the BaseSubManager's state is READY or ERROR
+	 * @param listener CompletionListener that is called once the BaseSubManager's state is READY, LIMITED, or ERROR
 	 */
 	public void start(CompletionListener listener){
 		this.completionListener = listener;
-		if((state == READY || state == ERROR) && completionListener != null){
-			completionListener.onComplete(state == READY);
+		if((state == READY || state == LIMITED || state == ERROR) && completionListener != null){
+			completionListener.onComplete(state == READY || state == LIMITED);
 			completionListener = null;
 		}
 	}
--- a/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
@@ -36,8 +36,8 @@ public abstract class BaseSubManager {
 	 */
 	public void start(CompletionListener listener){
 		this.completionListener = listener;
-		if((state == READY || state == LIMITED || state == ERROR) && completionListener != null){
-			completionListener.onComplete(state == READY || state == LIMITED);
+		if((getState() == READY || getState() == LIMITED || getState() == ERROR) && completionListener != null){
+			completionListener.onComplete(getState() == READY || getState() == LIMITED);
 			completionListener = null;
 		}
 	}
@@ -53,8 +53,8 @@ public abstract class BaseSubManager {
 		synchronized (STATE_LOCK) {
 			this.state = state;
 		}
-		if((state == READY || state == LIMITED || state == ERROR) && completionListener != null ){
-			completionListener.onComplete(state == READY || state == LIMITED);
+		if((getState() == READY || getState() == LIMITED || getState() == ERROR) && completionListener != null ){
+			completionListener.onComplete(getState() == READY || getState() == LIMITED);
 			completionListener = null;
 		}
 	}
--- a/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
@@ -53,11 +53,8 @@ public abstract class BaseSubManager {
 		synchronized (STATE_LOCK) {
 			this.state = state;
 		}
-		if( (state == READY || state == LIMITED) && completionListener != null ){
-			completionListener.onComplete(true);
-			completionListener = null;
-		}else if(state == ERROR && completionListener != null){
-			completionListener.onComplete(false);
+		if((state == READY || state == LIMITED || state == ERROR) && completionListener != null ){
+			completionListener.onComplete(state == READY || state == LIMITED);
 			completionListener = null;
 		}
 	}
--- a/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
@@ -36,8 +36,9 @@ public abstract class BaseSubManager {
 	 */
 	public void start(CompletionListener listener){
 		this.completionListener = listener;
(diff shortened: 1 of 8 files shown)


sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
```diff
@@ -53,7 +53,7 @@ protected void transitionToState(int state) {
 		synchronized (STATE_LOCK) {
 			this.state = state;
 		}
-		if(state == READY && completionListener != null){
+		if( (state == READY || state == LIMITED) && completionListener != null ){
 			completionListener.onComplete(true);
 			completionListener = null;
 		}else if(state == ERROR && completionListener != null){
```
joeygrover: Can we match the logic between the `start()` method and this one? I think the `start()` method logic is cleaner.

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -129,94 +127,137 @@ public void onError(String info, Exception e){
...
+
+		// Set the app icon
+		if (state == BaseSubManager.READY || state == BaseSubManager.LIMITED) {
+			setAppIcon();
+		}
+	}
+
+	private void setAppIcon(){
+		if (SdlManager.this.appIcon != null && SdlManager.this.appIcon.getName() != null) {
+			if (!fileManager.hasUploadedFile(SdlManager.this.appIcon)) {
```
BrettyWhite: should we check if the filemanager is not null or in a ready state here?

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -129,94 +127,137 @@ public void onError(String info, Exception e){
...
+				managerListener.onError(info, null);
+			} else {
+				managerListener.onStart();
+			}
+		}
+
+		// Set the app icon
+		if (state == BaseSubManager.READY || state == BaseSubManager.LIMITED) {
+			setAppIcon();
+		}
```
joeygrover: This logic seems a little out of place/out of scope of the method that it is in. This should probably be in the `checkState` method or the `transitionToState` method.

joeygrover: Or create a new method `onReady` in the manager that can handle anything that needs to be handled during the transition to READY.

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -129,94 +127,137 @@ public void onError(String info, Exception e){
...
+		// Start sub managers
+		this.permissionManager.start(subManagerListener);
+		this.fileManager.start(subManagerListener);
+		if (lockScreenConfig.isEnabled()){
+			this.lockScreenManager.start(subManagerListener);
+		}
+		this.screenManager.start(subManagerListener);
+	}
+
+	public int getState() {
```
joeygrover: Can we add some javadoc to this public method? Probably with `#see` tags to the states

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -129,94 +127,137 @@ public void onError(String info, Exception e){
...
 
 		if(this.videoStreamingManager != null) {
 			this.videoStreamingManager.dispose();
 		}
 
-		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
-			if (this.audioStreamManager != null) {
-				this.audioStreamManager.dispose();
-			}
+		if (this.audioStreamManager != null && Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
joeygrover: Honestly we probably don't need the version check here. If the `AudioStreamManager` is not null it got created because this check has already passed and if it is not null it should be `disposed`

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -427,15 +459,16 @@ public SdlManager build() {
...
+			sdlManager.transitionToState(BaseSubManager.SETTING_UP);
 
 			return sdlManager;
 		}
 	}
 
 	private void checkSdlManagerState(){
-		if (state != BaseSubManager.READY){
-			throw new IllegalStateException("SdlManager is not ready for use, be sure to initialize with start() method, implement callback, and use SubManagers in the SdlManager's callback");
+		if (state != BaseSubManager.READY && state != BaseSubManager.LIMITED){
```
joeygrover: The state should be accessed through the `getState` method here since there is synchronization around the getting and setting.

sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
```diff
@@ -32,12 +32,12 @@ public BaseSubManager(@NonNull ISdl internalInterface){
...
 	/**
 	 * Starts up a BaseSubManager, and calls provided callback once BaseSubManager is done setting up or failed setup.
-	 * @param listener CompletionListener that is called once the BaseSubManager's state is READY or ERROR
+	 * @param listener CompletionListener that is called once the BaseSubManager's state is READY, LIMITED, or ERROR
 	 */
 	public void start(CompletionListener listener){
 		this.completionListener = listener;
-		if((state == READY || state == ERROR) && completionListener != null){
-			completionListener.onComplete(state == READY);
+		if((state == READY || state == LIMITED || state == ERROR) && completionListener != null){
```
joeygrover: `state` should be accessed through the getState method. I would just add a line `int state = getState();` before the if statement

sdl_android/src/main/java/com/smartdevicelink/managers/BaseSubManager.java
bilal-alsharifi: Sure. Will do that.

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
bilal-alsharifi: Good catch. I will add the checks.

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
bilal-alsharifi: Makes sense. I will create `onReady()` method to handle that logic.

sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -129,94 +130,144 @@ public void onError(String info, Exception e){
...
+			} else {
+				managerListener.onStart();
+			}
+		}
+	}
+
+	private void onReady(){
+		// Set the app icon
+		if (SdlManager.this.appIcon != null && SdlManager.this.appIcon.getName() != null) {
+			if (fileManager != null && (fileManager.getState() == BaseSubManager.READY || fileManager.getState() == BaseSubManager.LIMITED) && !fileManager.hasUploadedFile(SdlManager.this.appIcon)) {
```
BrettyWhite: i don't think file manager is ever in a limited state

bilal-alsharifi: Makes sense. I will remove that check.

PR #900, opened 2018-10-10
Fix: make sure `MultiplexTcpTransport` creates a TCP socket over Wi-Fi network

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_android/issues/899

### Risk
This PR only affects `MultiplexTcpTransport` class.

### Testing Plan
Repeat the steps mentioned in my comment https://github.com/smartdevicelink/sdl_android/issues/899#issuecomment-428445159 and make sure that the library establishes a TCP connection to Core.

### Summary
This PR updates `MultiplexTcpTransport` to make sure it creates a `Socket` instance which is bound to Wi-Fi network.

##### Breaking Changes
* None.

##### Enhancements
* None.

##### Bug Fixes
* fixes `MultiplexTcpTransport` opening a TCP socket via mobile network when the head unit has no Internet connectivity over Wi-Fi network and the phone is associated with cellular network.

### Tasks Remaining:
- None

### CLA

### DIFF
--- a/sdl_android/src/main/java/com/smartdevicelink/transport/MultiplexTcpTransport.java
+++ b/sdl_android/src/main/java/com/smartdevicelink/transport/MultiplexTcpTransport.java
@@ -32,6 +32,7 @@
 
 package com.smartdevicelink.transport;
 
+import android.content.Context;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Message;
@@ -39,6 +40,7 @@ import android.util.Log;
 
 import com.smartdevicelink.protocol.SdlPacket;
 import com.smartdevicelink.transport.enums.TransportType;
+import com.smartdevicelink.transport.utl.WiFiSocketFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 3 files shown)


shoamano83:
Hi @joeygrover , please kindly review.
According to Ford team this is in priority, so I opened a PR against RC branch. Let me know if I should also create one for `develop`. Thanks.

joeygrover:
Priority is understood, note that we are in our testing period of the release so I can't guarantee this PRs inclusion into the 4.7 release, however, if changes requested are made it makes it much more likely.

I do have issues with this PR just being outright accepted. The PR itself would create a new requirement for all app developers to add the `ACCESS_NETWORK_STATE` permission to their app which would be a breaking change since it will throw an exception and cause their app to crash. Because of this, the router service would get closed down and cause all other apps to disconnect from the head unit which is a very poor user experience. Furthermore, I do not see a requirement that the IP address of the module _must_ be part of a local network; theoretically the module could send a static IP it has that can be accessed via the WAN connection regardless of WiFi connection or not. 

I would propose the following changes:

1.  Do not throw an exception when it can't access a WiFi socket.
2. A WiFi socket may be preferred, but if one can not be created for whatever reason, a default socket should be created and returned.
3. The `createWiFiSocket` method must check for the `ACCESS_NETWORK_STATE` permission and return null if does not have it. Which then, following bullet #2 a default socket would be created and returned.
4. Please add some unit tests around this new class, likely will have to use mockito to mock the network interfaces and system service.
5. Proper null checks should be put into place in the `createWiFiSocket`to ensure the router service does not crash from an NPE

shoamano83:
Thank you @joeygrover for your feedbacks. I'll definitely reflect them into my PR.

Just one thing, I was less worried about `ACCESS_NETWORK_STATE` since the official integration guide https://smartdevicelink.com/en/guides/android/getting-started/integration-basics/ states that the permission is required. I still agree that checking the permission and returning null is a good idea, so I'll follow it.

joeygrover:
@shoamano83 you are right. I will have to revisit why we added that permission to the guide, because I can't think of where it is used off hand. That makes this PR easier to accept with the changes we agreed upon.

shoamano83:
Hi @joeygrover , I updated my code and also added unit tests.
Thanks.

PR #991, opened 2019-03-12
Add RPC notification listeners to SdlManager's Builder 

bilal-alsharifi:
This PR adds RPC notification listeners to SdlManager's Builder in Android and JavaSE

### DIFF
--- a/android/sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
@@ -55,6 +55,7 @@ import com.smartdevicelink.util.Version;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Vector;
 
 /**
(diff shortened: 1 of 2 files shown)


android/sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -584,6 +586,9 @@ public void onTransportEvent(List<TransportRecord> connectedTransports, boolean
 				if (sdlSecList != null && !sdlSecList.isEmpty()) {
 					proxy.setSdlSecurityClassList(sdlSecList);
 				}
+				for (FunctionID functionID: onRPCNotificationListeners.keySet()) {
```
joeygrover: Get keySet and set to an instance, check for null and empty set before looping through

bilal-alsharifi: Okay I updated the pr to do the null checks

PR #992, opened 2019-03-12
Feature/sync cloud methods

BrettyWhite:
- expose getAuthToken in SdlManager
- add logic in LifecycleManager to download cloud icons onSystemRequest


javaSE/src/main/java/com/smartdevicelink/managers/lifecycle/LifecycleManager.java
```diff
@@ -651,7 +678,7 @@ public void onProtocolServiceDataACK(SessionType sessionType, int dataSize, byte
 
     @Override
     public void onAuthTokenReceived(String token, byte sessionID) {
```
bilal-alsharifi: This method should be called in another class to set the token. I think the value should be set in SdlProtocol then propagated to LifecycleManager

PR #1016, opened 2019-03-19
Retrieve version name for all projects from one file in the root repo

bilal-alsharifi:
Retrieve version name for all projects from one VERSION file in the root repo directory

### CLA


VERSION
```diff
@@ -0,0 +1 @@
+4.7.2
```
joeygrover: Should be `RC1-4.8.0`

PR #1066, opened 2019-05-16
Fix issues in sendRPCs and sendSequentialRPCs

### LINKED ISSUE #1061: `sendSequentialRPCs` doesn't call `onResponse` and stops on first `onError`
##### Reproduction Steps
1. Create a list of RPCs where one is for sure invalid
2. call sdlManager.sendSequentialRPCs and set a completion handler overriding all methods
3. observe handler method calls

##### Expected Behavior
1. All RPCs should be sent one after the other
2. `OnMultipleRequestListener.onResponse` should be called to **all** RPCs regardless of success. This is very important on unsuccessful subscribing to vehicle data
3. `OnMultipleRequestListener.onError` should be called for the unsuccessful request. The error should not interrupt the sequential operation.
4. `onResponse` listener set into the RPC should be called

##### Observed Behavior
1. `OnMultipleRequestListener.onResponse` is never called.
2. Requests are not sent after first error
3. `onResponse` listener set into the RPC is overriden by the manager

* Android Version: Any
* SDL Android Version: 4.8
* Testing Against: hello_sdl_android

bilal-alsharifi:
Fixes #1061

### Risk

### Summary
This PR fixes some issue in `sendRPCs` and `sendSequentialRPCs` methods:
- Calls the `onResponse/onError` for every RPC's `OnRPCResponseListener` 
- Calls the `onResponse/onError` in the `OnMultipleRequestListener` for every RPC 
- Continue sending RPCs if one failed
- `onUpdate` in `OnMultipleRequestListener`  is called for every RPC incuding the failed ones 

### Testing Plan
- Create a list of RPCs that you want to send and make sure one of them at least fails. The one that fails should not be the last one.
- Set an `OnRPCResponseListener` listener for all RPCs
- Send the RPCs using `sendSequentialRPCs`
- Make sure that `onResponse/onError` is called for every RPC in both the RPC's `OnRPCResponseListener` and the `OnMultipleRequestListener`
- Make sure that `onUpdate` is called for every RPC in the `OnMultipleRequestListener` and that `onFinished()` is called once in the end
- Repeat the same test using `sendRPCs`
- Repeat the tests again for both `sendRPCs` and `sendSequentialRPCs` in JavaSE library

### CLA

### DIFF
--- a/android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
@@ -4381,6 +4381,18 @@ public abstract class SdlProxyBase<proxyListenerType extends IProxyListenerBase>
 		} else {
 			// Notifications and Responses
 			sendRPCMessagePrivate(rpc);
+			if (listener != null && rpcs.size() > 0) {
+				listener.onUpdate(rpcs.size());
+			}
+			// recurse after sending a notification or response as there is no response.
+			try {
+				sendSequentialRequests(rpcs, listener);
+			} catch (SdlException e) {
+				e.printStackTrace();
+				if (listener != null) {
+					listener.onError(0, Result.GENERIC_ERROR, e.toString());
+				}
+			}
 		}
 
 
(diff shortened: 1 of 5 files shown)


android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -4313,7 +4315,7 @@ public void sendSequentialRequests(final List<? extends RPCMessage> rpcs, final
 			throw new SdlException("You must send some RPCs", SdlExceptionCause.INVALID_ARGUMENT);
 		}
 
-		int requestCount = rpcs.size();
+		final int requestCount = rpcs.size();
```
BrettyWhite: This name was changed in the `LifeCycleManager` version. We should maintain one or the other, your choice

PR #973, opened 2019-02-05
Fix for issue #972: Adding multi-touch support to VideoStreamingManager

shiniwat:
Fix for the issue #972 where multi-touch events are not working in VideoStreamingManager.

Fixes #972 

This PR is **ready** for review.

### Risk
This PR makes **no** API changes.

### Testing Plan
This PR includes unit test added to SdlProxyBaseTests.java.

### Summary
Added multi-touch support to VideoStreamingManager.

##### Breaking Changes
* n/a

##### Enhancements
* n/a

##### Bug Fixes
* #972 

### Tasks Remaining:
- n/a

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #1071, opened 2019-05-21
Add support for handling multiple TouchEvents in one OnTouchEvent notification 

### LINKED ISSUE #972: VideoStreamingManager should support multi-touch events
### Bug Report
VideoStreamingManager currently does not support multi-touch events. Because OnTouchEvent RPC can notify multiple TouchEvents, VideoStreamingManager should support them. 

##### Reproduction Steps
1. Start VPM
2. Made some multi-touch gesture.
3. See how ON_TOUCH_EVENT notification listener in VideoStreamingManager handles the OnTouchEvent.

##### Expected Behavior
When OnTouchEvent in RPC notification contains multiple TouchEvents, VideoStreamingManager should handle all of them.

##### Observed Behavior
VideoStreamingManager handles single MotionEvent in this case.

* Android Version: Android 9.0
* SDL Android Version: sdl_android develop branch at 2facdaee8a97ac30f8964e5746d9ca496153656e
* Testing Against: SDL Core 4.5 + our own HMI.

Will make a PR later.

bilal-alsharifi:
Fixes #972 and it is based on #973 

### Risk

### Testing Plan
Unit tests will be added to test converting one `OnTouchEvent` that has multiple `TouchEvents` to multiple `MotionEvents`

### Summary
-  Update `VideoStreamManager` to be able to convert one Sdl `OnTouchEvent` to multiple Android `MotionEvents`
-  Update `SdlProxyBase` to be able to convert one Sdl `OnTouchEvent` to multiple Android `MotionEvents`
- Pass list of all active pointers to `MotionEvent` so the developer can later get the correct number of active pointers from the `MotionEvent` and the coordinates for each pointer
- Update the converted `MotionEvent` actions to exactly match the native Android ones. For example, if the user touches the screen using two fingers at the same time, the action will be `ACTION_POINTER_DOWN (1)` instead of just `ACTION_POINTER_DOWN` for the second touch 

 ### Tasks Remaining:
- [x] [Update `VideoStreamManager`]
- [x] [Update `SdlProxyBase`]
- [x] [Write new tests]

### CLA

### DIFF
--- a/android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
@@ -544,6 +544,7 @@ public class VideoStreamManager extends BaseVideoStreamManager {
 
 			if(motionEventAction == MotionEvent.ACTION_UP || motionEventAction == MotionEvent.ACTION_CANCEL){
 				//If the motion event should be finished we should clear our reference
+				sdlMotionEvent.pointers.clear();
 				sdlMotionEvent = null;
 				break;
 			} else if((motionEventAction & MotionEvent.ACTION_MASK) == MotionEvent.ACTION_POINTER_UP){
--- a/android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
@@ -478,6 +478,26 @@ public class VideoStreamManager extends BaseVideoStreamManager {
 		}
 	}
 
+	@Override
+	protected void onTransportUpdate(List<TransportRecord> connectedTransports, boolean audioStreamTransportAvail, boolean videoStreamTransportAvail){
+
+		isTransportAvailable = videoStreamTransportAvail;
+
+		if(internalInterface.getProtocolVersion().isNewerThan(new Version(5,1,0)) >= 0){
+			if(videoStreamTransportAvail){
+				checkState();
+			}
+		}else{
+			//The protocol version doesn't support simultaneous transports.
+			if(!videoStreamTransportAvail){
+				//If video streaming isn't available on primary transport then it is not possible to
+				//use the video streaming manager until a complete register on a transport that
+				//supports video
+				transitionToState(ERROR);
+			}
+		}
+	}
+
 	List<MotionEvent> convertTouchEvent(OnTouchEvent onTouchEvent){
 		List<MotionEvent> motionEventList = new ArrayList<MotionEvent>();
 
@@ -648,26 +668,6 @@ public class VideoStreamManager extends BaseVideoStreamManager {
 
 		void removePointerById(int id){
 			pointers.remove(getPointerById(id));
-		}
-	}
-
-	@Override
-	protected void onTransportUpdate(List<TransportRecord> connectedTransports, boolean audioStreamTransportAvail, boolean videoStreamTransportAvail){
-
-		isTransportAvailable = videoStreamTransportAvail;
-
-		if(internalInterface.getProtocolVersion().isNewerThan(new Version(5,1,0)) >= 0){
-			if(videoStreamTransportAvail){
-				checkState();
-			}
-		}else{
-			//The protocol version doesn't support simultaneous transports.
-			if(!videoStreamTransportAvail){
-				//If video streaming isn't available on primary transport then it is not possible to
-				//use the video streaming manager until a complete register on a transport that
-				//supports video
-				transitionToState(ERROR);
-			}
 		}
 	}
 
--- a/android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
@@ -152,8 +152,10 @@ public class VideoStreamManager extends BaseVideoStreamManager {
 		public void onNotified(RPCNotification notification) {
 			if(notification != null && remoteDisplay != null){
 				List<MotionEvent> motionEventList = convertTouchEvent((OnTouchEvent)notification);
-				for (MotionEvent motionEvent : motionEventList) {
-					remoteDisplay.handleMotionEvent(motionEvent);
+				if (motionEventList != null && !motionEventList.isEmpty()) {
+					for (MotionEvent motionEvent : motionEventList) {
+						remoteDisplay.handleMotionEvent(motionEvent);
+					}
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -476,145 +476,176 @@ public void run() {
...
+		TouchType touchType = onTouchEvent.getType();
+		if (touchType == null) { return null; }
 
-		MotionEvent.PointerProperties[] pointerProperties = new MotionEvent.PointerProperties[eventListSize];
-		MotionEvent.PointerCoords[] pointerCoords = new MotionEvent.PointerCoords[eventListSize];
+		if(sdlMotionEvent == null) {
+			if (touchType == TouchType.BEGIN) {
+				sdlMotionEvent = new SdlMotionEvent();
+			} else{
+				return motionEventList;
```
joeygrover: Why is an empty array passed here instead of null?

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -476,145 +476,176 @@ public void run() {
...
+					motionEventAction = MotionEvent.ACTION_MOVE;
+					eventTime = downTime + touchEvent.getTimestamps().get(touchEvent.getTimestamps().size() - 1) - downTimeOnHMI;
 					break;
 				case END:
-					//Clears out pointers that have ended
-					setPointerStatuses(MotionEvent.ACTION_UP, pointerProperties);
-
-					if(pointerStatuses.size() == 0){
+					if(pointers.size() > 1){
+						motionEventAction = MotionEvent.ACTION_POINTER_UP | pointers.indexOf(getPointerById(touchEvent.getId())) << MotionEvent.ACTION_POINTER_INDEX_SHIFT;
```
joeygrover: For these shifts I think we should check to make sure the pointers are available first. `indexOf` will return a -1 value if it is not contained in the list. This would include if the `getPointerById` returned null.

android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -8061,150 +8061,182 @@ public void onServiceError(SdlSession session, SessionType type, String reason)
...
 				default:
 					break;
 			}
-			return motionEvent;
+			return motionEventAction;
 		}
 
-		private void setPointerStatuses(int motionEvent, MotionEvent.PointerProperties[] pointerProperties){
+		Pointer getPointerById(int id) {
+			for (Pointer pointer : pointers) {
```
joeygrover: Maybe add a null and size check to `pointers`

android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -8061,150 +8061,182 @@ public void onServiceError(SdlSession session, SessionType type, String reason)
...
+				this.y = 0.0f;
+			}
+
+			void setCoords(float x, float y) {
+				this.x = x;
+				this.y = y;
+			}
+		}
+
+		private List<Pointer> pointers = new ArrayList<>();
```
joeygrover: This is an unsynchronized list but has multiple methods that can alter its contents which can lead to unexpected results including concurrent modification exceptions. Either all methods accessing this list need to be synchronized to the same lock, or the data type will need to be changed.

android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java
```diff
@@ -8061,150 +8061,182 @@ public void onServiceError(SdlSession session, SessionType type, String reason)
...
-
 	}
 
 	/**
 	 * Keeps track of the current motion event for VPM
 	 */
-	private static class SdlMotionEvent{
-		long startOfEvent;
-		SparseIntArray pointerStatuses = new SparseIntArray();
+	private static class SdlMotionEvent {
```
joeygrover: Can you move the method `onTransportUpdate` above this class and the `convertTouchEvent` so that this class is at the end of the file?

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -476,145 +476,176 @@ public void run() {
...
+		return motionEventList;
 	}
 
 	/**
 	 * Keeps track of the current motion event for VPM
 	 */
-	private static class SdlMotionEvent{
-		long startOfEvent;
-		SparseIntArray pointerStatuses = new SparseIntArray();
+	private static class SdlMotionEvent {
```
joeygrover: Can you move the method onTransportUpdate above this class and the convertTouchEvent so that this class is at the end of the file?

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -476,145 +476,176 @@ public void run() {
...
+				this.x = 0.0f;
+				this.y = 0.0f;
+			}
+			void setCoords(float x, float y) {
+				this.x = x;
+				this.y = y;
+			}
+		}
+
+		private List<Pointer> pointers = new ArrayList<>();
```
joeygrover: This is an unsynchronized list but has multiple methods that can alter its contents which can lead to unexpected results including concurrent modification exceptions. Either all methods accessing this list need to be synchronized to the same lock, or the data type will need to be changed.

joeygrover:
@shiniwat can you confirm this fix works for you? This PR is based off your work in #972.

shiniwat:
@joeygrover Sorry for the slow response, but I found a bunch of implementation difference between our PR and feature/multi_touch_events branch, including unit test. I am trying to run the equivalent unit test code in multi_touch_events branch right now...

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
bilal-alsharifi: Theoretically, the else shouldn't be hit. It makes more sense to return null though. I will update that.

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
bilal-alsharifi: Makes sense. I will use CopyOnWriteArrayList for the pointers list

shiniwat:
We're double-checking this change on our end. Please allow us to review this in next couple of days. Thanks.

theresalech:
@shiniwat we will need to merge this PR by 2019-06-07 in order to stay on track with our 4.9 release timeline.  Please provide any feedback by that time.  Thank you!

shiniwat:
We double checked this change, running the unit test we had in our PR, along with running our test app with this change. Finally finished verification and the change works fine. Thanks.

bilal-alsharifi:
Awesome. Thanks for confirming that.

PR #1107, opened 2019-07-11
Align SystemCapabilityManager with iOS

### LINKED ISSUE #1255: SystemCapabilityManager doesn't fully support capability subscriptions
### Bug Report
 The capability subscription ability has been added to the `sdl_java_suite` library as described in this [proposal](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0167-app-services.md). The `SystemCapabilityManager` handles the `OnSystemCapabilityUpdated` notifications correctly. However, currently, there is no way to subscribe to a specific capability using the managers which makes capability subscription not very usable.
`sdl_ios` has that support so `sdl_java_suite` should add the ability to support capablity subscription to align with ios.

* Android Version: 4.10.1
* SDL Android Version: 10
* Testing Against: Core 6.0

bilal-alsharifi:
Fixes #1255 

### Risk

### Testing Plan
Unit tests have been added to test the newly added code. Also, these smoke tests can be done:
* Test getting a capability using the new `getCapabilty()` method and make sure capability is retrieved correctly.
* Test adding and removing an `OnSystemCapabilityListener` and confirm that the manger subscribes the first time we add a listener for that type. And unsubscribe when we remove the last listener for that type.

### Summary
This PR does the follwing changes:
* Adds a new ` getCapability(systemCapabilityType, listener, forceUpdate)`  method to the `SystemCapabilityManager` 
* Depreciates the two old `getCapability()` methods
* Update adding and removing listeners methods is SCM to support subscriptions

### CLA


base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -496,15 +559,23 @@ private void retrieveCapability(final SystemCapabilityType systemCapabilityType,
...
 			public void onResponse(int correlationId, RPCResponse response) {
-				if(response.getSuccess()){
+				if (response.getSuccess()) {
 					Object retrievedCapability = ((GetSystemCapabilityResponse) response).getSystemCapability().getCapabilityForType(systemCapabilityType);
 					cachedSystemCapabilities.put(systemCapabilityType, retrievedCapability);
-					if(scListener!=null){scListener.onCapabilityRetrieved(retrievedCapability);	}
-				}else{
-					if(scListener!=null){scListener.onError(response.getInfo());}
+					if (scListener != null) {
+						scListener.onCapabilityRetrieved(retrievedCapability);
```
joeljfischer: Would this cause the listener to receive this information twice, once from here and once from `notifyListeners`

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -415,34 +417,89 @@ public boolean isCapabilitySupported(SystemCapabilityType type){
...
+
+		return capability;
+	}
+
 	/**
 	 * @param systemCapabilityType Type of capability desired
 	 * @param scListener callback to execute upon retrieving capability
 	 */
+	@Deprecated
 	public void getCapability(final SystemCapabilityType systemCapabilityType, final OnSystemCapabilityListener scListener){
```
joeljfischer: Is it possible to have this simply call `getCapability(type, scListener, subscribe)` above with `subscribe` as `null`?

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -415,34 +417,89 @@ public boolean isCapabilitySupported(SystemCapabilityType type){
...
+		retrieveCapability(systemCapabilityType, scListener, null);
 	}
 
 	/**
 	 * @param systemCapabilityType Type of capability desired
 	 * @return Desired capability if it is cached in the manager, otherwise returns a null object
 	 * and works in the background to retrieve the capability for the next call
 	 */
+	@Deprecated
 	public Object getCapability(final SystemCapabilityType systemCapabilityType){
```
joeljfischer: See above, but with `scListener` and `subscribe` as `null`.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -452,7 +509,7 @@ public Object getCapability(final SystemCapabilityType systemCapabilityType){
 	 * @param listener callback to execute upon retrieving capability
 	 */
 	public void addOnSystemCapabilityListener(final SystemCapabilityType systemCapabilityType, final OnSystemCapabilityListener listener){
```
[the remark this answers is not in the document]
bilal-alsharifi: It is a little different cause the `addOnSystemCapabilityListener` adds the listener to the list of listeners that can be called in the future if there are future updates. The `GetCapablity` only calls the listener one time.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
bilal-alsharifi: I don't think so because `notifyListeners` should not get called in that case.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
bilal-alsharifi: Good idea. I will do that.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -496,15 +559,23 @@ private void retrieveCapability(final SystemCapabilityType systemCapabilityType,
 		}
 		final GetSystemCapability request = new GetSystemCapability();
 		request.setSystemCapabilityType(systemCapabilityType);
+		request.setSubscribe(subscribe != null ? subscribe : isSubscribedToSystemCapability(systemCapabilityType)); // If subscribe is null, then don't change the current subscription status
 		request.setOnRPCResponseListener(new OnRPCResponseListener() {
 			@Override
 			public void onResponse(int correlationId, RPCResponse response) {
-				if(response.getSuccess()){
+				if (response.getSuccess()) {
```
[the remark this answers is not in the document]
bilal-alsharifi: You are right. I will go over them and try to make them more consistent.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
bilal-alsharifi: I think you are right. I will test that.

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
```diff
@@ -211,6 +234,98 @@ public void onError(String info) {
...
+		onSystemCapabilityListener = mock(OnSystemCapabilityListener.class);
+		doAnswer(createOnSendGetSystemCapabilityAnswer(true)).when(internalInterface).sendRPC(any(GetSystemCapability.class));
+		scm.setCapability(SystemCapabilityType.VIDEO_STREAMING, videoStreamingCapability);
+		retrievedCapability =  (VideoStreamingCapability) scm.getCapability(SystemCapabilityType.VIDEO_STREAMING, onSystemCapabilityListener, true);
+		assertTrue(Test.TRUE, Validator.validateVideoStreamingCapability((VideoStreamingCapability) systemCapability.getCapabilityForType(SystemCapabilityType.VIDEO_STREAMING), retrievedCapability));
+		verify(internalInterface, times(1)).sendRPC(any(GetSystemCapability.class));
+		verify(onSystemCapabilityListener, times(1)).onCapabilityRetrieved(any(Object.class));
+
+
+		// Test case 3 (capability cached, listener not null, forceUpdate true)
```
joeljfischer: This test case states it is the same as 2

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
```diff
@@ -211,6 +234,98 @@ public void onError(String info) {
...
+		internalInterface = mock(ISdl.class);
+		when(internalInterface.getSdlMsgVersion()).thenReturn(sdlMsgVersion);
+		scm = new SystemCapabilityManager(internalInterface);
+		onSystemCapabilityListener = null;
+		doAnswer(createOnSendGetSystemCapabilityAnswer(true)).when(internalInterface).sendRPC(any(GetSystemCapability.class));
+		scm.setCapability(SystemCapabilityType.VIDEO_STREAMING, videoStreamingCapability);
+		retrievedCapability =  (VideoStreamingCapability) scm.getCapability(SystemCapabilityType.VIDEO_STREAMING, onSystemCapabilityListener, false);
+		assertTrue(Test.TRUE, Validator.validateVideoStreamingCapability((VideoStreamingCapability) systemCapability.getCapabilityForType(SystemCapabilityType.VIDEO_STREAMING), retrievedCapability));
+		verify(internalInterface, times(0)).sendRPC(any(GetSystemCapability.class));
+	}
```
joeljfischer: Shouldn't `addOnSystemCapabilityListener` and `removeOnSystemCapabilityListener` tests with subscriptions, such as unsubscribing when the last listener is removed? This isn't required, but tests should exist for that if possible.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -415,45 +417,91 @@ public boolean isCapabilitySupported(SystemCapabilityType type){
...
+	 * @return desired capability if it is cached in the manager, otherwise returns a null object and works in the background to retrieve the capability for the next call
 	 */
-	public void getCapability(final SystemCapabilityType systemCapabilityType, final OnSystemCapabilityListener scListener){
-		Object capability = cachedSystemCapabilities.get(systemCapabilityType);
-		if(capability != null){
-			scListener.onCapabilityRetrieved(capability);
-			return;
-		}else if(scListener == null){
-			return;
+	private Object getCapabilityPrivate(final SystemCapabilityType systemCapabilityType, final OnSystemCapabilityListener scListener, final Boolean subscribe, final boolean forceUpdate) {
```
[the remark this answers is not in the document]
bilal-alsharifi: Honestly, It is not common to use underscores in method names in java. I was trying to follow the code style that java_suite had. Like this method for example:
https://github.com/smartdevicelink/sdl_java_suite/blob/797b7a14a7fd67fdf6faa504865e4e22228858ef/android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L2241

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
bilal-alsharifi: Yea that will be better you are right. I will add more tests for the two methods.

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
```diff
@@ -211,13 +234,139 @@ public void onError(String info) {
...
+		scm = new SystemCapabilityManager(internalInterface);
+		onSystemCapabilityListener = null;
+		doAnswer(createOnSendGetSystemCapabilityAnswer(true, null)).when(internalInterface).sendRPC(any(GetSystemCapability.class));
+		scm.setCapability(SystemCapabilityType.VIDEO_STREAMING, videoStreamingCapability);
+		retrievedCapability =  (VideoStreamingCapability) scm.getCapability(SystemCapabilityType.VIDEO_STREAMING, onSystemCapabilityListener, false);
+		assertTrue(Test.TRUE, Validator.validateVideoStreamingCapability((VideoStreamingCapability) systemCapability.getCapabilityForType(SystemCapabilityType.VIDEO_STREAMING), retrievedCapability));
+		verify(internalInterface, times(0)).sendRPC(any(GetSystemCapability.class));
+	}
+
+	public void testAddOnSystemCapabilityListener() {
```
joeljfischer:
```suggestion
	public void testAddOnSystemCapabilityListenerWithSubscriptionsSupported() {
```

Ideally, you'd have tests without subscriptions supported as well, but for now, at least make sure that your assumptions are mentioned in the method name so that future Bilal can add tests if needed.

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
```diff
@@ -211,13 +234,139 @@ public void onError(String info) {
...
+				}
+				response.setSystemCapability(systemCapability);
+				getSystemCapability.getOnRPCResponseListener().onResponse(CorrelationIdGenerator.generateId(), response);
+				return null;
+			}
+		};
+		return onSendGetSystemCapabilityAnswer;
+	}
+
+	public void testGetCapability() {
```
joeljfischer:
```suggestion
	public void testGetCapabilityWithSubscriptionsSupported() {
```

Make sure that all of your relevant assumptions are mentioned in the method name.

PR #1119, opened 2019-07-22
[SDL 0179]Pixel density and Scale

### LINKED ISSUE #804: [SDL 0179] Pixel density and Scale
Proposal: [Pixel density and Scale](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0179-pixel-density-and-scale.md)

> This proposal is about adding parameters to `VideoStreamingCapabilities` for pixel density and scale.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/526

skibyte:
Fixes #804 

### Risk

### Testing Plan
Unit tests added for:
- VideoStreamManager
- VideoStreamingCapability
- VideoStreamingParameters
- HapticManagerTest

Functional testing can be done by editing the preferred resolution,
and scale in HMI and testing that the captured view has the proper size and touches are reported
correctly and with the correct coordinates while clicking on the HMI interface.
Also the Automatic Haptic Rect sent to HMI should have the propper size according to the scale value set. 

### Summary
The main change is related to changing the size of the view captured.
This is done by adding a new scale parameter in the video
streaming capability and by dividing the preferred resolution from the
HMI and the scale.

Changes in the calculation of the touch coordinates are not needed due
to the array VideoStreamManager.touchScalar, this array is already
handling the differences in the size between the preferred resolution
and the captured view.
The touch coordinates are correctly being divided by the scale value.

Automatic haptic rects are resized according to the scale value.
### CLA

### DIFF
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/rpc/datatypes/VideoStreamingCapabilityTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/rpc/datatypes/VideoStreamingCapabilityTests.java
@@ -14,11 +14,9 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 
 
 public class VideoStreamingCapabilityTests extends TestCase {
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/rpc/datatypes/VideoStreamingCapabilityTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/rpc/datatypes/VideoStreamingCapabilityTests.java
@@ -14,9 +14,11 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
 
 public class VideoStreamingCapabilityTests extends TestCase {
@@ -30,6 +32,9 @@ public class VideoStreamingCapabilityTests extends TestCase {
 		msg.setPreferredResolution(Test.GENERAL_IMAGERESOLUTION);
 		msg.setMaxBitrate(Test.GENERAL_INT);
 		msg.setIsHapticSpatialDataSupported(Test.GENERAL_BOOLEAN);
+		msg.setDiagonalScreenSize(Test.GENERAL_DOUBLE);
+		msg.setPixelPerInch(Test.GENERAL_DOUBLE);
+		msg.setScale(Test.GENERAL_DOUBLE);
 	}
 
 	/**
@@ -41,12 +46,18 @@ public class VideoStreamingCapabilityTests extends TestCase {
 		ImageResolution res = msg.getPreferredResolution();
 		Integer maxBitrate = msg.getMaxBitrate();
 		Boolean isHapticSpatialDataSupported = msg.getIsHapticSpatialDataSupported();
+		Double diagonalScreenSize = msg.getDiagonalScreenSize();
+		Double pixelPerInch = msg.getPixelPerInch();
+		Double scale = msg.getScale();
 
 		// Valid Tests
 		assertEquals(Test.MATCH, (List<VideoStreamingFormat>) Test.GENERAL_VIDEOSTREAMINGFORMAT_LIST, format);
 		assertEquals(Test.MATCH, (ImageResolution) Test.GENERAL_IMAGERESOLUTION, res);
 		assertEquals(Test.MATCH, (Integer) Test.GENERAL_INT, maxBitrate);
 		assertEquals(Test.MATCH, (Boolean) Test.GENERAL_BOOLEAN, isHapticSpatialDataSupported);
+		assertEquals(Test.MATCH, Test.GENERAL_DOUBLE, diagonalScreenSize);
+		assertEquals(Test.MATCH, Test.GENERAL_DOUBLE, pixelPerInch);
+		assertEquals(Test.MATCH, Test.GENERAL_DOUBLE, scale);
 
 		// Invalid/Null Tests
 		VideoStreamingCapability msg = new VideoStreamingCapability();
@@ -56,6 +67,9 @@ public class VideoStreamingCapabilityTests extends TestCase {
 		assertNull(Test.NULL, msg.getPreferredResolution());
 		assertNull(Test.NULL, msg.getSupportedFormats());
 		assertNull(Test.NULL, msg.getIsHapticSpatialDataSupported());
+		assertNull(Test.NULL, msg.getDiagonalScreenSize());
+		assertNull(Test.NULL, msg.getPixelPerInch());
+		assertNull(Test.NULL, msg.getScale());
 	}
 
 	public void testJson() {
@@ -66,6 +80,9 @@ public class VideoStreamingCapabilityTests extends TestCase {
 			reference.put(VideoStreamingCapability.KEY_PREFERRED_RESOLUTION, Test.GENERAL_IMAGERESOLUTION);
 			reference.put(VideoStreamingCapability.KEY_SUPPORTED_FORMATS, Test.GENERAL_VIDEOSTREAMINGFORMAT_LIST);
 			reference.put(VideoStreamingCapability.KEY_HAPTIC_SPATIAL_DATA_SUPPORTED, Test.GENERAL_BOOLEAN);
+			reference.put(VideoStreamingCapability.KEY_DIAGONAL_SCREEN_SIZE, Test.GENERAL_DOUBLE);
+			reference.put(VideoStreamingCapability.KEY_PIXEL_PER_INCH, Test.GENERAL_DOUBLE);
+			reference.put(VideoStreamingCapability.KEY_SCALE, Test.GENERAL_DOUBLE);
 
 			JSONObject underTest = msg.serializeJSON();
 			assertEquals(Test.MATCH, reference.length(), underTest.length());
@@ -87,6 +104,9 @@ public class VideoStreamingCapabilityTests extends TestCase {
 					for(VideoStreamingFormat vsf : vsfReference){
 						assertTrue(Validator.validateSupportedFormats(vsf, new VideoStreamingFormat(JsonRPCMarshaller.deserializeJSONObject(vsfArray.getJSONObject(i++)))));
 					}
+				} else if (key.equals(VideoStreamingCapability.KEY_DIAGONAL_SCREEN_SIZE) || key.equals(VideoStreamingCapability.KEY_PIXEL_PER_INCH) ||
+						key.equals(VideoStreamingCapability.KEY_SCALE)) {
+					assertEquals(JsonUtils.readDoubleFromJsonObject(reference, key), JsonUtils.readDoubleFromJsonObject(underTest, key), 0.0005);
 				}
 			}
 		} catch (JSONException e) {
--- a/android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
@@ -107,8 +107,8 @@ public class VirtualDisplayEncoder {
     }
 
     @SuppressWarnings("unused")
-    public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, VideoStreamingFormat format) {
-        this.streamingParams = new VideoStreamingParameters(displayDensity, frameRate, bitrate, interval, resolution, format);
+    public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, double scale, VideoStreamingFormat format) {
+        this.streamingParams = new VideoStreamingParameters(displayDensity, frameRate, bitrate, interval, scale, resolution, format);
     }
 
     @SuppressWarnings("unused")
--- a/android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
@@ -107,14 +107,8 @@ public class VirtualDisplayEncoder {
     }
 
     @SuppressWarnings("unused")
-    @Deprecated
     public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, VideoStreamingFormat format) {
         this.streamingParams = new VideoStreamingParameters(displayDensity, frameRate, bitrate, interval, resolution, format);
-    }
-
-    @SuppressWarnings("unused")
-    public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, double scale, VideoStreamingFormat format) {
-        this.streamingParams = new VideoStreamingParameters(displayDensity, frameRate, bitrate, interval, scale, resolution, format);
     }
 
     @SuppressWarnings("unused")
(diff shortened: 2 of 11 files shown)


skibyte:
Hi, @joeygrover  could you please review when you have some time?

android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
```diff
@@ -107,8 +107,8 @@ public VideoStreamingParameters getStreamingParams(){
     }
 
     @SuppressWarnings("unused")
-    public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, VideoStreamingFormat format) {
-        this.streamingParams = new VideoStreamingParameters(displayDensity, frameRate, bitrate, interval, resolution, format);
+    public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, double scale, VideoStreamingFormat format) {
```
BrettyWhite: This is changing a public method which cannot be done in a minor update. Please deprecate the old method and then add a separate one with this new signature

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -499,6 +499,11 @@ protected void onTransportUpdate(List<TransportRecord> connectedTransports, bool
 		}
 	}
 
+    void createTouchScalar(ImageResolution resolution, DisplayMetrics displayMetrics) {
```
BrettyWhite: This code is used in other places in this class, please refactor to not have duplicate code

base/src/main/java/com/smartdevicelink/streaming/video/VideoStreamingParameters.java
```diff
@@ -73,11 +76,13 @@ public VideoStreamingParameters(){
     }
 
     public VideoStreamingParameters(int displayDensity, int frameRate, int bitrate, int interval,
-                                    ImageResolution resolution, VideoStreamingFormat format){
+                                    double scale, ImageResolution resolution,
```
BrettyWhite: Please deprecate the old constructor and add a new one with the new signature.

BrettyWhite:
There is also an older `VideoStreamingManager` class within `SdlProxyBase` that appears to have been missed: https://github.com/smartdevicelink/sdl_java_suite/blob/master/android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L7929

BrettyWhite:
Does the `HapticInterfaceManager` also need to take into account the scale of the screen? https://github.com/smartdevicelink/sdl_java_suite/blob/master/android/sdl_android/src/main/java/com/smartdevicelink/haptic/HapticInterfaceManager.java

skibyte:
> There is also an older `VideoStreamingManager` class within `SdlProxyBase` that appears to have been missed: https://github.com/smartdevicelink/sdl_java_suite/blob/master/android/sdl_android/src/main/java/com/smartdevicelink/proxy/SdlProxyBase.java#L7929
Hi @BrettyWhite thanks for your feedback,
The SDLManager.getVideoStreamManager() used by the clients of the library returns an instance of the class which I modified (com.smartdevicelink.managers.video.VideoStreamingManager).

Do you think that VideoStreamingManager from SdlProxyBase is still used?

skibyte:
> Does the `HapticInterfaceManager` also need to take into account the scale of the screen? https://github.com/smartdevicelink/sdl_java_suite/blob/master/android/sdl_android/src/main/java/com/smartdevicelink/haptic/HapticInterfaceManager.java
> 
> Edit: If youre not familiar with it, here is what it does: https://smartdevicelink.com/en/guides/android/video-streaming-for-navigation-apps/supporting-haptic-input/

Thanks a lot, I was not aware of the haptic device, but it sounds like it could be affected by the scale,
do you happen to know by chance if haptic can be tested with HMI sdl?

android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
skibyte: DONE

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
skibyte: DONE

base/src/main/java/com/smartdevicelink/streaming/video/VideoStreamingParameters.java
skibyte: DONE

android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
BrettyWhite: well, this needs to go out as a minor change, so what we can do is keep the old method, and simply add a new one with your new parameter. that makes it a minor change.

BrettyWhite:
There are some app partners that have not updated to the manager system, so yes we need to update legacy items as well until we can get rid of them in a major release, such as the VideoStreamManager in proxy base. 

As for testing the haptic interface manager, we usually create custom views with buttons in known positions, we then have to modify the core's hmicapabilities file to allow support for haptic data. I then place a log in the HapticInterfaceManager and log out the positions it finds for the buttons. If it matches where i know the button positions are, then it worked.

android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
BrettyWhite: So what you did here is correct, this is a minor change, which is what we were aiming for ✅

skibyte:
DONE.

@BrettyWhite 
I had to change the automatic focusing rects in order to consider the scale value,
otherwise the rect would have a different size than the widgets in the head.
Manual rects have not been changed.

About the legacy code, I analyzed the code and it uses the same video parameters than than the current code, therefore no changes are needed in the old VideoStreamManager.
The method createTouchScalar that I did was just made for testing purposes and does not change the implementation logic, I just extracted the code in that method.

Let me know any comments.

android/sdl_android/src/main/java/com/smartdevicelink/haptic/HapticInterfaceManager.java
```diff
@@ -91,6 +93,15 @@ private void findHapticRects(View root, final List<HapticRect> hapticRects) {
         List<View> focusables = new ArrayList<>();
         getFocusableViews(root, focusables);
 
+        double scale = 1.0;
+        ISdl proxy = proxyHolder.get();
```
BrettyWhite: Potential NPE. This needs to be wrapped in a `if (proxyHolder.get() != null){ `

skibyte: The condition is there but in the caller method: refreshHapticData
https://github.com/smartdevicelink/sdl_java_suite/blob/30b3ba4e6e93b8b66133b3eb971887d5286b0440/android/sdl_android/src/main/java/com/smartdevicelink/haptic/HapticInterfaceManager.java#L77-L88

BrettyWhite: Calling `.get()` on a WeakReference can cause an NPE as there is no guarantee that that reference will be kept. So in the case pointed out, you will need one around line 78 and one around line 97. 

I see that that existed previously, but it is a potential NPE, so it should be fixed. 

https://developer.android.com/reference/java/lang/ref/package-summary.html#reachability

skibyte: DONE

BrettyWhite:
Hi, 

While reviewing this proposal, I am getting

```
2019-08-13 10:46:18.374 9340-9367/com.sdl.hellosdlandroid E/VirtualDisplayEncoder: Unable to create Virtual Display.
2019-08-13 10:46:18.375 9340-9367/com.sdl.hellosdlandroid W/System.err: java.lang.RuntimeException: java.lang.IllegalStateException: start failed
2019-08-13 10:46:18.375 9340-9367/com.sdl.hellosdlandroid W/System.err:     at com.smartdevicelink.encoder.VirtualDisplayEncoder.start(VirtualDisplayEncoder.java:152)
```
on more than one occasion. 

The first, is where I leave the default

```js
    videoStreamingCapability: {
         preferredResolution: {
             resolutionWidth: window.screen.availWidth,
             resolutionHeight: window.screen.availHeight
         },
```

in `systemCapabilities.js` in the corresponding sdl_hmi pr: https://github.com/smartdevicelink/sdl_hmi/pull/165/files

The second is when I did modify the `systemCapabilities.js` file to 

```js
SDL.systemCapabilities = 
{
    videoStreamingCapability: {
        preferredResolution: {
            resolutionWidth: 800,
            resolutionHeight: 350
        },
        maxBitrate: 2147483647,
        supportedFormats: [
            {
                protocol:  "RAW",
                codec: "H264"
                
            }
        ],
        hapticSpatialDataSupported: true,
        diagonalScreenSize: 22,
        pixelPerInch: 96,
        scale: 2
    }
}
```

When scale == 1, it will launch fine in gstreamer. When scale is modified to be greater than 1, such as 2, it will fail with the error above. Similarly, changing the diagonalScreenSize also creates that error. 

I am testing against the PRs in the sdl_core and sdl_hmi repositories:

https://github.com/smartdevicelink/sdl_hmi/pull/165/files and https://github.com/smartdevicelink/sdl_core/pull/2757

I am using VMWare Fusion with Ubuntu 18.04 and Gstreamer 1.14.5

skibyte:
Hi,
It seems the codec in the base code is not able to handle every resolution, but this happens even without the scale feature.
In the example that you mentioned where you use:
` resolutionWidth: 800,
  resolutionHeight: 350
scale: 2`
This will send to the core a view with a resolution of 400x175 you can test setting that resolution without the scale feature and it will cause the same codec error.

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/rpc/datatypes/VideoStreamingCapabilityTests.java
```diff
@@ -14,9 +14,11 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
```
BrettyWhite: this is unused

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/rpc/datatypes/VideoStreamingCapabilityTests.java
```diff
@@ -14,9 +14,11 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.HashMap;
```
BrettyWhite: this is unused

android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
```diff
@@ -107,10 +107,16 @@ public VideoStreamingParameters getStreamingParams(){
     }
 
     @SuppressWarnings("unused")
+    @Deprecated
     public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, VideoStreamingFormat format) {
         this.streamingParams = new VideoStreamingParameters(displayDensity, frameRate, bitrate, interval, resolution, format);
     }
 
+    @SuppressWarnings("unused")
+    public void setStreamingParams(int displayDensity, ImageResolution resolution, int frameRate, int bitrate, int interval, double scale, VideoStreamingFormat format) {
```
BrettyWhite: For any new public method added, can we please have some javadocs with them? just the params with a brief description should be adequate.

android/sdl_android/src/test/java/com/smartdevicelink/haptic/HapticInterfaceManagerTest.java
```diff
@@ -170,4 +253,43 @@ public Boolean answer(InvocationOnMock invocation) throws Throwable {
...
+    }
+
+    private void assertViewWithCapability(int x, int y, int w, int h, Rectangle expected, VideoStreamingCapability capability) {
+        when(mockProxy.getCapability(SystemCapabilityType.VIDEO_STREAMING)).thenReturn(capability);
+
+        View button = createView(x, y, w, h);
+
+
+        hapticMgr.refreshHapticData(button);
+        verify(mockProxy).sendRPCRequest(captor.capture());
```
BrettyWhite: can we use `sendRPC` here please? This method is deprecated.

base/src/main/java/com/smartdevicelink/proxy/rpc/ImageResolution.java
```diff
@@ -97,6 +97,9 @@ public ImageResolution(@NonNull Integer resolutionWidth, @NonNull Integer resolu
     }
     
     public void setResolutionWidth(@NonNull Integer resolutionWidth) {
```
BrettyWhite: For this method and the corresponding height method, can we please add javadocs explaining why the input may be modified?

base/src/main/java/com/smartdevicelink/proxy/rpc/VideoStreamingCapability.java
```diff
@@ -92,4 +96,31 @@ public Boolean getIsHapticSpatialDataSupported() {
 	public void setIsHapticSpatialDataSupported(Boolean hapticSpatialDataSupported) {
 		setValue(KEY_HAPTIC_SPATIAL_DATA_SUPPORTED, hapticSpatialDataSupported);
 	}
+
+	public Double getDiagonalScreenSize() {
```
BrettyWhite: can we please add javadocs to the new public methods?

base/src/main/java/com/smartdevicelink/streaming/video/VideoStreamingParameters.java
```diff
@@ -72,22 +75,34 @@ public VideoStreamingParameters(){
...
     }
 
+
+    @Deprecated
     public VideoStreamingParameters(int displayDensity, int frameRate, int bitrate, int interval,
                                     ImageResolution resolution, VideoStreamingFormat format){
+        this(displayDensity, frameRate, bitrate, interval, DEFAULT_SCALE, resolution, format);
+    }
+
+    public VideoStreamingParameters(int displayDensity, int frameRate, int bitrate, int interval,
```
BrettyWhite: javadocs please

base/src/main/java/com/smartdevicelink/streaming/video/VideoStreamingParameters.java
```diff
@@ -177,6 +197,14 @@ public int getInterval() {
         return interval;
     }
 
+    public void setScale(double scale) {
```
BrettyWhite: javadocs please

android/sdl_android/src/main/java/com/smartdevicelink/haptic/HapticInterfaceManager.java
```diff
@@ -75,22 +77,35 @@ public void setHapticData(List<HapticRect> hapticData) {
...
+                List<HapticRect> hapticRects = new ArrayList<>();
+                findHapticRects(root, hapticRects);
 
-            SendHapticData msg = new SendHapticData();
-            msg.setHapticRectData(hapticRects);
+                SendHapticData msg = new SendHapticData();
+                msg.setHapticRectData(hapticRects);
 
-            proxy.sendRPCRequest(msg);
+                proxy.sendRPCRequest(msg);
```
BrettyWhite: can we switch this to `sendRPC` please?

android/sdl_android/src/main/java/com/smartdevicelink/encoder/VirtualDisplayEncoder.java
skibyte: Thanks a lot for the review,
now  I realize I should not modify this method nor VideoStreamingParameters. The scale value is coming from the head.
By adding the scale parameter in setStreamingParams, I was allowing the user to modify the scale, which is wrong.

BrettyWhite: alright, sounds good. please let me know when you are ready for re-review. We have our machine set up so we shouldn't have any weird VM issues this time. 👍

skibyte: thanks, Ready for re review

PR #1122, opened 2019-07-23
Implement SDL-0216 Widget Support

### LINKED ISSUE #1069: [SDL 0216] Widget support
Proposal: [Widget support](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0216-widget-support.md)

> This proposal is about widget support for SDL applications on modern infotainment systems and large displays. At the same time the proposed feature should prepare SDL for multi-window and multi-display support.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/664

Steering Committee Decision:

>The Steering Committee voted to accept this proposal with the agreed upon revisions, described in [this comment](https://github.com/smartdevicelink/sdl_evolution/issues/664#issuecomment-492421396) and included in PR #729. 

> The proposal .md file was updated to reflect these revisions on 5/20/19.

mjuarez-ford:
Fixes  #1069 

### Risk

### Testing Plan
Unit test for new classes and properties.
Unit test and smoke test for the screen manager. (display capability redesign).

### Summary
Implement proposal as defined 

### Tasks Remaining:
- [x] [Implement classes]
- [x] [Unit test to classes]
- [x] [Apply changes suggested in the merge comments]
- [x] [Test against Core]

### CLA


mjuarez-ford:
This PR is **ready** for review @joeljfischer

bilal-alsharifi:
Thanks, @mjuarez-ford.  I think you mean Joey Grover though because Joel is part of the iOS team, not the Java Suite team.

mjuarez-ford:
Thanks @bilal-alsharifi, 
Hi @joeygrover, this PR is ready for review.

theresalech:
@mjuarez-ford as there is a currently an Evolution Proposal in review regarding this feature (here, we will need to wait for that proposal to be voted on by the SDLC Steering Committee before reviewing this PR. The vote is scheduled to take place 2019-07-30.

kshala-ford:
@theresalech PR is ready for review. Different to the iOS PR we already fixed the number discrepancy to Core...

theresalech:
Fixes #1069

base/src/main/java/com/smartdevicelink/proxy/rpc/SystemCapability.java
```diff
@@ -115,4 +117,23 @@ public void setCapabilityForType(SystemCapabilityType type, RPCStruct capability
...
         }
     }
+
+    /**
+     * set the displayCapabilities portion of SystemCapability.
+     * {@code 1<= displayCapabilities.size() <= 100}
+     *
+     * @param displayCapabilities Contains a list of capabilities of display.
+     */
+    public void setDisplayCapabilities(List<DisplayCapability> displayCapabilities) {
```
bilal-alsharifi: These two methods should not be added because `getCapabilityForType()`, `setCapabilityForType()` should work for all capabilties incldunbg DISPLAYS

base/src/main/java/com/smartdevicelink/proxy/rpc/SystemCapability.java
```diff
@@ -49,6 +50,7 @@
 	public static final String KEY_VIDEO_STREAMING_CAPABILITY = "videoStreamingCapability";
```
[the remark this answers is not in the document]
bilal-alsharifi: We can change the type in the `getCapabilityForType()`/`setCapabilityForType()` to return/accept java `Object` instead of `RPCStruct` so they can use `List<DisplayCapability>`. That shouldn't be a breaking change because `Object` is a more generic type than `RPCStruct` so existing code that uses these APIs should still work in most cases.

base/src/main/java/com/smartdevicelink/proxy/rpc/SystemCapability.java
```diff
@@ -37,6 +37,7 @@
 import com.smartdevicelink.proxy.rpc.enums.SystemCapabilityType;
```
bilal-alsharifi: There are some spacing/formatting issues in this file. They are not introduced in this PR though. Could you please fix them in this PR as well since we are already making changes to the file?

base/src/main/java/com/smartdevicelink/proxy/rpc/SystemCapability.java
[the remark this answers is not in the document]
bilal-alsharifi: Also, could you please add Javadoc for `getCapabilityForType(), setCapabilityForType()`?

base/src/main/java/com/smartdevicelink/proxy/rpc/SystemCapability.java
mjuarez-ford: Hi @bilal-alsharifi thanks for the recommendation, but the method  `setCapabilityForType()` receives a RPCStruct. `setDisplayCapabilities()` receives a `List<DisplayCapability>`. 
What is your recommendation?

bilal-alsharifi: You are right. I think the best thing we can do is to change the `getCapabilityForType()`,  and `setCapabilityForType()` to return/take  java `Object` instead of `RPCStruct`. I explained more about that in other comments on the same `SystemCapability.java` file. What do you think?

bilal-alsharifi:
Hi @mjuarez-ford we will also need to update `ScreenManager` to monitor only `OnHmiStatus` for the main display. The `ScreenManager` doesn't work well when `OnHmiStatus` for multiple windows are received.
Because currently, for example, if a developer is using the `ScreenManager` and if `CreateWindow` RPC is sent, the `ScreenManager` will receive `OnHMIStatus` with level NONE for the newly-created window. Which will block the `ScreenManager` from sending any `SHOW` RPCs until `OnHMIStatus` not NONE is received again. 
Another example where it will fail is when we set a text field using the `ScreenManager`, then create a window, then send `SHOW` to the new window without using the`ScreenManager` (because the `ScreenManager` cannot send a`SHOW` with `windowID` currently), then put widget in HMI FULL, the `ScreenManager` will send the old `SHOW` again (because it receives HMI FULL for the window) and that will override the new text.
There are multiple other scenarios where the `ScreenManager` will not work correctly because of the `OnHmiStatus` updates for widget windows. 
This is based on a discussion in this proposal:
https://github.com/smartdevicelink/sdl_evolution/issues/797

base/src/main/java/com/smartdevicelink/proxy/rpc/SystemCapability.java
kshala-ford: @bilal-alsharifi because the method is public this change is difficult to make. I'm not sure loosing type parameter type from an `Object` subclass to `Object` causes breaking changes but again... they are public methods.

bilal-alsharifi: You are right they are public.  But because `Object` is more generic than `RPCStruct` the change shouldn't break existing code because most developers will cast to the capability type anyway like the following :
```
RemoteControlCapabilities remoteControlCapabilities = (RemoteControlCapabilities) systemCapability.getCapabilityForType(SystemCapabilityType.REMOTE_CONTROL);
```
That will keep working whether the return type is `RPCStruct` or `Object`. I know cases like the following will break. But I think the most developer will not even do that:
```
RPCStruct rpcStruct = systemCapability.getCapabilityForType(SystemCapabilityType.REMOTE_CONTROL);
```
It should only be a compile-time error that devs can easily fix by casting to the right type.

I am looking at the internal uses for that method in the library just to have in an idea for developer will often use the method.  they will all not break.
https://github.com/smartdevicelink/sdl_java_suite/blob/014401c0dc52f185bd73431b6e386fd59e1a3c07/base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java#L111
and
https://github.com/smartdevicelink/sdl_java_suite/blob/014401c0dc52f185bd73431b6e386fd59e1a3c07/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java#L138

I do understand your concerns though. But also if we don't handle the `DISPLAYS` capability in that method, it will return null even if the capability exists which can lead to more serious run time error. also having a generic method for all types except the `DISPLAYS` sounds is a little confusing for the developers.

I know that changing that public method is not a great way but I think its pros outweigh the cons.  
The other option that we can do is deprecating the current methods and add new ones that take/return `Object`

mjuarez-ford:
@bilal-alsharifi thank you for the feedback. The PR should be complete now and is ready for review.

bilal-alsharifi:
Thanks, @mjuarez-ford for making the changes. It looks like Travis is failing on some tests though, do you mind taking a look at that before we review it?

kshala-ford:
Hello @bilal-alsharifi hmm, we discussed an issue how to make `PredefinedWindows` to be of type int (was string typed enum). Per mobile api this enum is int. Would you mind to check out the pattern of PrefefinedWindows.java? We don’t have that many integer based enums in our libraries. If you agree to the pattern we will update the tests. Otherwise if you suggest another method, we will change the enum implementation as well.

bilal-alsharifi:
Hi @kshala-ford maybe we can keep the int way and just add annotations similar to how we do it here:
https://github.com/smartdevicelink/sdl_java_suite/blob/014401c0dc52f185bd73431b6e386fd59e1a3c07/base/src/main/java/com/smartdevicelink/managers/permission/BasePermissionManager.java#L70

This is how Google recommends doing the int enums for Android as far as I know. JavaSE will just ignore these annotations I belive. What do you think?

joeygrover:
@kshala-ford & @bilal-alsharifi it is part of the RPC spec and therefore shouldn't made into the newer style of enums for Android. It should remain similar to other enums for the RPC spec. See [`BitsPerSample.java`](https://github.com/smartdevicelink/sdl_java_suite/blob/014401c0dc52f185bd73431b6e386fd59e1a3c07/base/src/main/java/com/smartdevicelink/proxy/rpc/enums/BitsPerSample.java). Instead of saving the internal name, create a `VALUE` int.

joeygrover:
Something like the following:

```
public enum PredefinedWindows {
    /**
     * The default window is a main window pre-created on behalf of the app.
     */
    DEFAULT_WINDOW (0),
    /**
     * The primary widget of the app.
     */
    PRIMARY_WIDGET (1),

    ;


    final int VALUE;
    /**
     * Private constructor
     */
    PredefinedWindows (int value) {
        this.VALUE = value;
    }

    public static PredefinedWindows valueForInt(int value) {

        for (PredefinedWindows anEnum : EnumSet.allOf(PredefinedWindows.class)) {
            if (anEnum.getValue() == value) {
                return anEnum;
            }
        }
        return null;
    }

    public int getValue(){
        return VALUE;
    }
}

```

mjuarez-ford:
@joeygrover @bilal-alsharifi Thanks for all the support and time for this proposal. I finished the suggested changes and the test are running back again.

theresalech:
We will hold off on merging until [SDL 0243](https://github.com/smartdevicelink/sdl_evolution/issues/797) is implemented in Java Suite as well.

kshala-ford:
@theresalech I don’t think the PM can make this decision without asking the steering committee. In fact SDL 0243 is not part of the roadmap for the upcoming release and would need another Steering Committee vote. Please don’t get me wrong. As the author I already started to work to code the proposal and we would see this proposal as part of the upcoming release. It is also unfair to the developer and donator to take this PR as a hostage as other merges into develop can cause conflicts that need to be worked on again.

Please merge this PR into develop.

joeygrover:
From the voted on and accepted proposal:

> Subsequent proposals should define how to refactor managers (ScreenManager and SystemCapabilityManager) to read window capabilities notifications as well as the deprecated parameters. The managers update will include deprecating DISPLAY system capability. **The proxy library releases including this feature should also include mentioned manager updates to ensure this feature to be easily accessible with the managers from the start.**

kshala-ford:
I’m sorry @joeygrover but proposals cannot define or change a release roadmap especially to a feature which was not defined at the time of voting.

joeygrover:
@kshala-ford maybe there was a misunderstanding during the vote for that revision to the proposal including language that tied the two together, but we did discuss that the release should include both proposals. This was due to the fact they can be contained based on the timing differences between Core and proxy deliverables. At this time, all we have to go on is that statement which was voted on by the SDLC. If the PM were to go against that statement, they would be going against the SDLC.

I think it's worth noting that there is no foreseen concerns that both PRs will successfully be merged into the release. Also the Core and related PRs are unaffected by the subsequent proxy PR and will be merged into mainline when approved.

kshala-ford:
This is an evidence of how the PM is bending their self made rules to their desires and an evidence that the PM doesn't trust members of the SDLC. This way of miscommunication (or even no communication) is a real challenge to the patience for code donator. This is not acceptable.

I will make the communication here now very clear. Here's what the SC has accepted as revisions for 0216 Widget Support:

> The revisions will include specifying that managers will need to be updated to support the release of this feature within the proxy libraries, and this update will include the deprecating of `DISPLAY`. 

This revision resulted into the following sentence in the proposal paper:

> Subsequent proposals should define how to refactor managers (ScreenManager and SystemCapabilityManager) to read window capabilities notifications as well as the deprecated parameters. The managers update will include deprecating DISPLAY system capability. The proxy library releases including this feature should also include mentioned manager updates to ensure this feature to be easily accessible with the managers from the start.

It says that subsequent proposals should do this change but it cannot be used to make decisions on the SDL roadmap because the proposals did not exist at the time of acceptance. The briefly described idea doesn't specify anything. 

With SDL 0243 we voted on an update for the system capability manager only and we came to an agreement which is:

> 4. Sounds good. I just wanted to make sure everyone was on the same page. While I would really like for the screen manager to be updated to provide a more intuitive API for the widgets feature, I understand that this proposal does not include those changes, and there will have to be another proposal to discuss them.

SDL 0243 was accepted with other revisions not related to the release date or roadmap. This is where I referring to. Because 0243 was not added to the roadmap it's is not documented (not even in some minutes) that is will be part of the next release. It's not even documented who is actually working on it or when the due date should be.

Don't get me wrong. We support SDL 0243 to be part of the next release and majority of the code is included in the proposal. However it's not in the roadmap and therefore not in the repo projects. Adding SDL 0243 to the projects, assigning the donator and specify timing should be more than enough. The developers would develop in a new branch and PR into develop where this PR should be already merged into. There is no need to hold back this PR.

So please merge this PR into develop and add #1154 to the 4.10.0 project of this repo. The sooner this is done the earlier we can branch from the latest "develop" to start coding SDL 0243

joeygrover:
I apologize for the situation and hope that we can take steps to prevent this. As you know in our last call we discussed that the we as the PM are going to be taking some steps to hopefully help reduce these types of situations with better defined process. We really want to have a clear process that is easy to follow and eliminates the need for assumption or opinion based decisions. 

One of the things we can work on is making the minutes more succinct. While we try to keep them brief it is possible not all elements of the discussion get captured. In that event, relying on the conversation on the issue prior to the meeting is helpful tool. In this instance we can see the [last comment](https://github.com/smartdevicelink/sdl_evolution/issues/792#issuecomment-516509398) made that was agreed to (gave a thumbs up) by everyone in the discussion.

So I think based on your last reply, that comment, and the SDLC vote, the manager layer changes must be in this proxy release.

Why haven't we merge it now? As I tried to state in my previous comment and edit, while we have __every__ intention to merge this PR into the project, in the WORST case scenario where the manager layer changes PR does not make it in time for the release, the project maintainer would go to the SDLC Steering Committee and ask for guidance on what should be done. This would likely have three options 1. Delay the release in order to contain both PRs, 2. Release the proxy libraries with only the RPC changes and not the manager layer, or 3. Remove the feature from the release.  If in the _absolute worst case scenario_ the SDLC voted to remove the feature, it is much cleaner to not have any of the feature merged yet vs trying to remove the feature after it has been merged. This is the only reason why it hasn't been merged yet. I hope that you can see we have reviewed, tested, and approved this PR with our absolute intention to get this feature in and on time.

With that being said, I want to make sure everyone is satisfied by how the project is going, so in good faith we can merge this feature with the agreement from the PR author @mjuarez-ford that in the extremely unlikely but WORST case scenario they will be available for assistance during a potential removal and/or reopening of this PR. Can I get a thumbs up from everyone?

kshala-ford:
@joeygrover I think you don't understand the catch-22 and the delay to the whole project that you are causing. There are other proposals that need attention. Instead we need to deal with this unnecessary bureaucracy because the existing process applies to the case we have here.

> In this instance we can see the last comment made that was agreed to (gave a thumbs up) by everyone in the discussion.

Of course I support #1154 to the next release as I noted many times in this conversation. I don't understand why you keep thinking I wouldn't. I want to provide the pull requests but blocking this PR I need to ask: How am I going to provide the code for #1154 without this PR being in "develop" branch? From where should I branch from?

Just as described in https://github.com/smartdevicelink/sdl_java_suite/blob/master/.github/CONTRIBUTING.md
I want to branch from "develop" and provide the code for #1154. This requires this PR to be merged. Alternatively #1154 could be added into this PR but this can be problematic across different authors (SDLC members are not able to push code to other author's PRs anymore).

The hypothetical issue you describe is irrelevant compared my very real problem to contribute. As you want #1154 in the next release it should be added to the next release's project backlog. 

Until now you would have already merged the PR for #1154 into develop.

joeygrover:
If problems arise the project maintainer will just handle it. Please see the issue #1154 linked was already added to the project but was waiting to be triaged since its acceptance, as of last week I tiraged it as a contribution.  Please let us know when we can expect it.

PR #1156, opened 2019-08-23
Implement SDL-0184 CancelInteraction RPC

### LINKED ISSUE #840: [SDL 0184] Cancel Interaction RPC
Proposal: [Cancel Interaction RPC](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0184-cancel-interaction.md)

> Add the ability to allow interaction alerts to be canceled.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/539

NicoleYarroch:
Fixes #840 

### Risk

### Testing Plan
Smoke tests added for the `CancelInteraction` response and request classes. Test cases added to the screen manager tests.

### Summary
* Added `CancelInteraction` request and response classes to the library. 
* Added `cancel` method to choice set manager.
* Added `dismissKeyboard:` method to screen manager
* Updated the current `presentKeyboard:` method to return an id that can be passed to the `dismissKeyboard:` method

##### Enhancements
* Added new RPC `CancelInteraction` to the library
* Screen manager now supports canceling/dismissing a choice set and dismissing a presented keyboard.

### Tasks Remaining:
- [x] Test against sdl_core v.6.0
- [x] Set the finalized `functionID` for `CancelInteraction`
- [x] Update the current `presentKeyboard:` method to return a `cancelID` that can be used to dismiss the keyboard
- [x] Update the `dismissKeyboard:`method

### CLA


android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/choiceset/PresentChoiceSetOperationTests.java
```diff
@@ -102,4 +119,114 @@ private KeyboardProperties getKeyBoardProperties(){
...
+		// Only supported with RPC spec versions 6.0.0+
+		presentChoiceSetOperation.sdlMsgVersion = new SdlMsgVersion(5, 3);
+		presentChoiceSetOperation.run();
+
+		choiceSet.cancel();
+
+		verify(internalInterface, never()).sendRPC(any(CancelInteraction.class));
+	}
+
+	public void testCancelingChoiceSetIfHeadUnitDoesNotSupportFeatureButThreadIsNotRunning(){
```
bilal-alsharifi: I see an issue with interrupting threads to cancel operations. Basically, we have three different threads `SdlService` thread, `Executor` thread, `SdlProxyBase` dispatcher thread. They can all run code in the operation class at some point which can cause some issues.

For example, the `SdlService` thread can call `choiceSet.cancel()` which interrupts that `SdlService` thread and not the `Executor` thread. However, when we check if the thread is interrupted in `AsynchronousOperation.isCancelled()`, it checks if the current thread is interrupted and because the `Executor` is running the operation.run(), it checks if that `Executor` thread is interrupted and it will be not be interrupted because it is a different thread.
 
That issue can clearly be seen if you run the operation in this test `testCancelingChoiceSetIfHeadUnitDoesNotSupportFeatureButThreadIsNotRunning()` in another thread (which what happens in an actual app)

try to replace:
```
presentChoiceSetOperation.run();
```
with this
```
new Thread(presentChoiceSetOperation).start();
```

You will see the test fails. I suggest not using thread interruptions cause they may lead to unpredictable scenarios.

PR #1209, opened 2019-10-24
sdl manager crashes when non-mandatory language is null

BrettyWhite:
Sdl manager crashes if language is null in the RAIR. THis param is optional, therefore it should not crash

### CLA


bilal-alsharifi:
I think  that should be updated for javaSE too

PR #1229, opened 2019-11-19
Fix potential overflow issue in BestRouterComparator

### LINKED ISSUE #1227: There is a risk of data overflow that may give results leading to unintended behavior
### Bug Report
There is a risk of data overflow that may give results leading to unintended behavior

##### Preconditions
1. App1.lastUpdateTime > App2.lastUpdateTime
2. The lastUpdateTime difference between APP1 and APP2 is outside of the values permitted for an int.

##### Reproduction Steps
1. Register App1
2. Activate App1

##### Expected Behavior
The RouterService of APP1 is used.

##### Observed Behavior
The RouterService of APP2 is used.

There is a risk of data overflow in the following statement.
`(int)(two.lastUpdateTime - one.lastUpdateTime);`

I know that maybe int is large enough to store the difference value.
This is certainly a theoretical risk but in practice there is almost no possibility of overflow happening.
And, I don't think it's necessary to compare the `lastUpdateTime`.
If the `routerServiceVersion` are different, we choose the newer one. If not, we can use either one.

```
int versionCompare =  two.routerServiceVersion	- one.routerServiceVersion;

if(versionCompare == 0){ //Versions are equal so lets use the one that has been updated most recently
	int updateTime =  (int)(two.lastUpdateTime - one.lastUpdateTime);
	if(updateTime == 0){
		//This is arbitrary, but we want to ensure all lists are sorted in the same order
		return	one.routerServiceComponentName.getPackageName().compareTo(two.routerServiceComponentName.getPackageName());
	}else{
		return updateTime;
	}
}else{
	return versionCompare;
}
```

* Android Version: Android 9
* SDL Android Version: 4.9.0
* Testing Against: SDL Core 5.1.3 + HMI running on our development environment.

##### Occurrence
50%

##### Recovery Method
None

bilal-alsharifi:
Fixes #1227 

### Risk

### Testing Plan
Install two sdl apps on the phone and make sure the router service is selected based on the compare function in `BestRouterComparator`

### Summary
This PR fixes a potential overflow issue that could happen if the result of subtracting to `long` numbers is outside the boundaries of an `int`

### CLA

### DIFF
--- a/android/sdl_android/src/main/java/com/smartdevicelink/util/SdlAppInfo.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/util/SdlAppInfo.java
@@ -153,12 +153,12 @@ public class SdlAppInfo {
                     int versionCompare =  two.routerServiceVersion  - one.routerServiceVersion;
 
                     if(versionCompare == 0){ //Versions are equal so lets use the one that has been updated most recently
-                        int updateTime =  (int)(two.lastUpdateTime - one.lastUpdateTime);
+                        long updateTime =  two.lastUpdateTime - one.lastUpdateTime;
                         if(updateTime == 0){
                             //This is arbitrary, but we want to ensure all lists are sorted in the same order
                             return  one.routerServiceComponentName.getPackageName().compareTo(two.routerServiceComponentName.getPackageName());
                         }else{
-                            return updateTime;
+                            return (updateTime < 0 ? -1 : 1);
                         }
                     }else{
                         return versionCompare;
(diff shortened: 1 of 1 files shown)


android/sdl_android/src/main/java/com/smartdevicelink/util/SdlAppInfo.java
```diff
@@ -153,12 +153,12 @@ public int compare(SdlAppInfo one, SdlAppInfo two) {
...
+                        long updateTime =  two.lastUpdateTime - one.lastUpdateTime;
                         if(updateTime == 0){
                             //This is arbitrary, but we want to ensure all lists are sorted in the same order
                             return  one.routerServiceComponentName.getPackageName().compareTo(two.routerServiceComponentName.getPackageName());
                         }else{
-                            return updateTime;
+                            return (updateTime < 0 ? -1 : 1);
                         }
                     }else{
                         return versionCompare;
```
joeygrover:
```suggestion
                        return (versionCompare < 0 ? -1 : 1);
```

joeygrover: Since we're updating the other return statement, it would make sense to do it here too.

PR #1250, opened 2020-01-07
Fix an issue with starting video stream after closing the app

### LINKED ISSUE #1248: Video streaming doesn't start again after closing the app using voice commands
### Bug Report
When a video streaming app is closed using voice commands, the app will go to HMI Level `NONE`. If the user opens the app again and tries to start video streaming, the streaming will not work until the app is force closed.

##### Reproduction Steps
1. Start the video stream
2. Close the app using voice commands 
3. Open the app again and start the video stream

##### Expected Behavior
 Video should play again

##### Observed Behavior
Video streaming doesn't work

* Android Version: 9
* SDL Android Version: 4.10.1
* Testing Against: Sync 3.0 TDK

bilal-alsharifi:
Fixes #1248 

### Risk

### Testing Plan
* Start video stream on a SYNC 3 head unit
* Close the app using voice commands
* Open the app again and start the video stream
* Make sure streaming works correctly 

### Summary
This PR handles a bug in some head units where the `StartService` for video streaming gets a `NAK` response after the app is being closed using voice commands. 
The PR fixes that by manually sending `EndService` from app side when the app is closed using voice commands. 

### CLA


android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -148,10 +148,16 @@ public void onNotified(RPCNotification notification) {
 				if (onHMIStatus.getWindowID() != null && onHMIStatus.getWindowID() != PredefinedWindows.DEFAULT_WINDOW.getValue()) {
 					return;
 				}
+				HMILevel prevHMILevel = hmiLevel;
 				hmiLevel = onHMIStatus.getHmiLevel();
 				if(hmiLevel.equals(HMILevel.HMI_FULL)){
 					checkState();
 				}
+				if (hasStarted && (prevHMILevel.equals(HMILevel.HMI_FULL) || prevHMILevel.equals(HMILevel.HMI_LIMITED)) && (hmiLevel.equals(HMILevel.HMI_NONE) || hmiLevel.equals(HMILevel.HMI_BACKGROUND))){
```
joeygrover: Potential NPE using the `equals` method here. You can either use the `==` comparator  or switch the enum instance calling the `equals()` metod, eg  `HMILevel.HMI_FULL.equals(prevHMILevel)` as the `HMILevel` enum instances will always be valid. Either solution will work.

joeygrover:
Looking at the original issue, I would also look at the potential issue [here](https://github.com/smartdevicelink/sdl_java_suite/blob/8e8bf1c3368b8208a84bee30a7aab902a6b218fe/android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java#L128) on line 128.

The manager transitions the streaming state machine to state `NONE` but if the remote display was not null it will be stopped and transition to state `STOPPED`. We should actually make the call to stop the remote display first, then transition to the streaming state machine to state `NONE` for the state machine. We would also want to transition the managers state to `SETTING_UP` based on this PR, right?

In code:

```
		@Override
		public void onServiceEnded(SdlSession session, SessionType type) {
			if(SessionType.NAV.equals(type)){
				if(remoteDisplay!=null){
					stopStreaming();
				}
				stateMachine.transitionToState(StreamingStateMachine.NONE);
				transitionToState(SETTING_UP);
			}
		}
```

PR #1273, opened 2020-02-12
[SDL-0234] Proxy Library RPC Generation

### LINKED ISSUE #1090: [SDL 0234] Proxy Library RPC Generation
Proposal: [Proxy Library RPC Generation](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0234-proxy-rpc-generation.md)

> This proposal adds automatic RPC generation for iOS and Java from the XML spec via a python script.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/741

Steering Committee Decision:

>The Steering Committee voted to approve this proposal with the following revision: there will be parsing code in each repository, and the rpc spec will be a submodule that pins which rpc spec will be used in each version of the library.

> The proposal .md file was updated to reflect these revisions on 6/10/19.

vladmu:
Fixes #1090

### Risk

### Testing Plan
- [x] I have verified that I have not introduced new warnings in this PR

#### Unit Tests
Covered by unit tests and PyLint

### Summary
According to `[SDL-0234] Proxy Library RPC Generation` proposal, this PR adds the Python generator script based on XML Parser from `smartdevicelink/rpc_spec` linked to the repository as Git submodule

##### Enhancements
* Added `utils/generator/rpc_spec` submodule pointed to `develop` branch of `smartdevicelink/rpc_spec` repository
* Created `Proxy RPC Generator` based on Python 3.5 and linked with parser from the `utils/generator/rpc_spec` submodule
* Added `README.md` with usage description and transformation rules for Proxy RPC Generator

### CLA


vladmu:
@theresalech @joeljfischer @joeygrover 
Based on this conversation https://github.com/smartdevicelink/sdl_java_suite/issues/1090#issuecomment-586310102 we'd like to notify here that we have found missing parameters in 3 enums that exist in XML but don't exist in the current code:
```java
FunctionID.RESERVED
AppInterfaceUnregisteredReason.UNSUPPORTED_HMI_RESOURCE
Result.CHAR_LIMIT_EXCEEDED
```
This leads to the failed result when we try to check the generated code with existing unit tests. Because they are part of the current MOBILE API specification, it seems missing parameters should be included in the current RPC code and unit tests should be adjusted.

utils/generator/README.md
```diff
@@ -0,0 +1,2638 @@
...
+     *
+     * @param hash The Hashtable to use
+     */
+    public [name](Hashtable<String, Object> hash) {
+        super(hash);
+    }
+```
+Where `[name]` is the value from the `"name"` attribute of `<struct>`.
+
+### Constructor with all required parameters, based on `"mandatory"` attribute of the `<param>`
```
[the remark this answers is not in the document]
ksologubov: > I am curious about the implementation. What's the order of the constructor parameters? Is it following the order of mandatory params elements in the struct?

Yes, it's following the order of mandatory parameters

utils/generator/README.md
```diff
@@ -0,0 +1,2638 @@
...
+ *
+ * @deprecated
+ * @since SmartDeviceLink [since_version]
+ * @see [see_reference]
+ */
+```
+Where:
+* `[description]` is `<description>` of the current `<enum>`, if exists.
+* `@deprecated` indicates the deprecation state if the `"deprecated"` attribute exists and is "true".
+* `@since` should be present, if the `"since"` attribute exists, and `[since_version]` is the `Major.Minor.Patch` formatted value of this attribute.
```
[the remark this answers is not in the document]
vladmu: @kshala-ford `<element name="FR-CA" internal_name="FR_CA">` will not have the `@since` JavaDoc as described in the document. Should it appear based on the parent `since` attribute?

vladmu:
@kshala-ford `exception of *Response` added, all suggested changes accepted, questions answered. Please take a look.

utils/generator/README.md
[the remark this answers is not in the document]
kshala-ford: Not really because @since is added to the class’s inline documentation but it would better appear in the ide documentary. I’m okay to leave as is.

kshala-ford:
@theresalech I approve this pull request.

theresalech:
@vladmu, if this is ready for Livio review, can you please update the PR description from `This PR is **[not ready]** for review.` to `This PR is **[ready]** for review.`?  Thanks!

vladmu:
@theresalech done, the PR is ready for Livio review. Thank you!

bilal-alsharifi:
Thanks, @vladmu for implementing the PR. I am just wondering if there is a specific reason for making the files in `templates/scripts/` have the `.groovy` extension ? can we make them use `.java` instead? because they are snippets from java files.

ksologubov:
Hello, @bilal-alsharifi I completely agree with your point. Will do this now.

bilal-alsharifi:
Thanks, @ksologubov for making the changes.

bilal-alsharifi:
As it was discussed in the [iOS issue](https://github.com/smartdevicelink/sdl_ios/issues/1298), no mappings for corner cases should be specified in the generator. Generic rules are fine (like replacing “sync” with “sdl” for example). The script should use the RPC specs without any case by case handling. Even if that would generate a different code than what the Java Suite currently has. Handling RPCs on case by case basis will make the scripts hard to maintain. Some examples of how things in the current mappings can be removed are:

- Enum values can always use the `name` attribute and completely ignore the `internal_name`.  This, I believe, will get rid of the use for the `kind` (simple, custom, complex) attribute that is specified for the enums in `mapping.json`. More generic rules can be added to make the variable accepted by the compiler like always adding `_` if the names start with a number.
- Float can always be used if the RPC spec says so. It doesn’t need to be replaced with Double.
- Because `FunctionID` is very different than any other enum, it shouldn’t be generated by the generator. We can always manually modify it.
- No need to add any methods that do not come directly from the spec (examples: `format()`, `valueForString()` ...)
- Imports shouldn’t be specified as corner cases. I think that would be possible after removing the corner cases methods as mentioned previously.

The previous list just gives some examples and it is not comprehensive. But the important part is to avoid making special cases for each item.

kshala-ford:
@theresalech Please add an agenda item about SDL-0234 implementation to the next steering committee meeting. The requested changes here and the discussion in sdl_ios are not compliant to the accepted proposal.

.gitignore
```diff
@@ -69,8 +69,14 @@ proguard/
 build/
 .gradle
 /local.properties
-/.idea/workspace.xml
```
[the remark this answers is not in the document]
ksologubov: This was just removed because whole .idea/ already included in .gitignore above. Maybe I'm missing something, so I will restore this 2 lines.

utils/generator/README.md
```diff
@@ -0,0 +1,2638 @@
...
+
+This constructor requires the import of `Hashtable` class
+```java
+import java.util.Hashtable;
+```
+
+Template:
+```java
+    /**
+     * Constructs a new [name] object indicated by the Hashtable parameter\
```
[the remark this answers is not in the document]
ksologubov: fixed

theresalech:
@kshala-ford Sure, we can include this on next week's agenda.  Livio will continue to review the implementation PRs (for Java Suite and iOS) as planned in the meantime as to stay on track with the current release timeline.

utils/generator/README.md
```diff
@@ -0,0 +1,2638 @@
...
+    * If the `<param>` has the `"type"` attribute value as the one of `<enum>` or `<struct>` name:
+        * The getter should call and return the result of the `getObject` method, where the first parameter is the corresponding Struct or Enum class, the second is the value of the corresponding static field described above;
+
+Setter template:
+```java
+    /**
+     * Sets the [name].
+     * [description]
+     *
+     * @param [name]
```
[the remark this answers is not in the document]
ksologubov: Hello @bilal-alsharifi 
parameter description moved to getter/setter/constructor javadoc as you described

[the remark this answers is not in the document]
bilal-alsharifi: Thanks @ksologubov for making the change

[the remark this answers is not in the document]
bilal-alsharifi: @ksologubov  would it be possible to make the param description starts on the same line as the param? that applies to the @retrun as well.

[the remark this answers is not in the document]
bilal-alsharifi: also, if we can move the `@since comments` from the `public static final String` keys to the getter/setters as well. Because developers will look at the docs for the methods not the keys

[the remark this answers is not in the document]
bilal-alsharifi: So the expected result for `iconResumed` for example will be something like this :

```
public static final String KEY_ICON_RESUMED = "iconResumed";

/**
 * Sets the iconResumed.
 *
 * @param iconResumed Existence of apps icon at system. If true, apps icon was resumed at system.
 * If false, apps icon is not resumed at system
 *
 * @since SmartDeviceLink 5.0.0
 */
public void setIconResumed(Boolean iconResumed) {
    setParameters(KEY_ICON_RESUMED, iconResumed);
}

/**
 * Gets the iconResumed.
 *
 * @return Boolean Existence of apps icon at system. If true, apps icon was resumed at system.
 * If false, apps icon is not resumed at system
 *
 * @since SmartDeviceLink 5.0.0
 */
public Boolean getIconResumed() {
    return getBoolean(KEY_ICON_RESUMED);
}
```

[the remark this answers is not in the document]
ksologubov: Thanks for example Bilal, I'm going to update this

utils/generator/templates/scripts/OnHMIStatus_format.java
```diff
@@ -0,0 +1,25 @@
+
+    @Override
```
bilal-alsharifi: The generator is making the `firsRun` as deprecated for some reason:
```
/**
 * @deprecated
 */
@Deprecated
private Boolean firstRun;
```

utils/generator/README.md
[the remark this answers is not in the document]
ksologubov: @bilal-alsharifi Generator was corrected per your comments

utils/generator/templates/scripts/OnHMIStatus_format.java
ksologubov: fixed

utils/generator/transformers/structs_producer.py
```diff
@@ -0,0 +1,138 @@
...
+            param.name = self.replace_sync(param.name)
+            i, p = self.extract_param(param)
+            imports.update(i)
+            params[param.name] = p
+
+        render = OrderedDict()
+        render['class_name'] = class_name
+        render['extends_class'] = extends_class
+        render['package_name'] = self.package_name
+        render['imports'] = imports
```
bilal-alsharifi: is not that duplicate of what is done in line 65? that applies to other producers as well.

utils/generator/test/runner.py
```diff
@@ -0,0 +1,41 @@
...
+    suite.addTests(TestLoader().loadTestsFromTestCase(TestFunctionsProducer))
+    suite.addTests(TestLoader().loadTestsFromTestCase(TestEnumsProducer))
+    suite.addTests(TestLoader().loadTestsFromTestCase(TestStructsProducer))
+    suite.addTests(TestLoader().loadTestsFromTestCase(CodeFormatAndQuality))
+
+    runner = TextTestRunner(verbosity=2)
+    runner.run(suite)
+
+
+if __name__ == '__main__':
```
bilal-alsharifi: Would it be possible to trigger the tests on every push using Git Workflow CI? similar to how we do that for Android, JavaSE, and JavaEE tests here:
https://github.com/smartdevicelink/sdl_java_suite/blob/develop/.github/workflows/android.yml

vladmu:
> @theresalech Please add an agenda item about SDL-0234 implementation to the next steering committee meeting. The requested changes here and the discussion in sdl_ios are not compliant to the accepted proposal.

@kshala-ford @theresalech respecting the committee decided to remove the following section from the proposal downsides:
>3. The generator could produce bad RPC classes, however, the library authors will ensure that this is not the case and that the first version of the generated code matches what already exists, and the unit tests will continue to pass.

It would be appreciated if you could summarize and define action points for us here regarding what we should do in this PR to follow the SDLC decision. Would it be enough to remove all custom mappings from the generator? Do we need to remove the customization tool itself? Etc.

theresalech:
@vladmu please address @bilal-alsharifi's review comments regarding removing custom mapping.  To answer your question specifically, you should remove the customization tool/logic.

joeljfischer:
@vladmu I do want to clarify, customization can and should still occur in _general_ cases that can be applied to everything of a particular type, e.g. all enum names that start with a number need to be prefixed by a `_`. If that necessitates that your customization logic remains, then so be it, but if the customization logic is only for corner case logic, then it can be removed.

vladmu:
> @vladmu I do want to clarify, customization can and should still occur in _general_ cases that can be applied to everything of a particular type, e.g. all enum names that start with a number need to be prefixed by a `_`. If that necessitates that your customization logic remains, then so be it, but if the customization logic is only for corner case logic, then it can be removed.

@joeljfischer thank you for clarification, that's clear. General rules are defined in the code and don't relate to the custom mappings tool. My questions were exactly about the customizations which were applied before for corner cases to make the generated code compatible with the old one.

utils/generator/generator.py
```diff
@@ -0,0 +1,443 @@
...
+            return loads(''.join(intermediate))
+        except (FileNotFoundError, JSONDecodeError) as message1:
+            self.logger.warning(message1)
+            return OrderedDict()
+
+    def write_file(self, file_name, template, data):
+        """
+        Calling producer/transformer instance to transform initial Model to dict used in jinja2 templates.
+        Applying transformed dict to jinja2 templates and writing to appropriate file
+        :param file_name: output js file
```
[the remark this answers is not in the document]
ksologubov: fixed, thanks

utils/generator/generator.py
```diff
@@ -0,0 +1,443 @@
...
+        :param template: name of template
+        :param data: transformed moder ready for apply to Jinja2 template
```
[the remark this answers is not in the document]
ksologubov: fixed, thanks

utils/generator/requirements.txt
```diff
@@ -0,0 +1,5 @@
+xmlschema
```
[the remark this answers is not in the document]
ksologubov: Bilal, docs.google.com link is not working for me, could you please check it?
I committed section for generator and it's dependencies in third_party.md

utils/generator/templates/enum_template.java
```diff
@@ -0,0 +1,201 @@
...
+    {%- elif kind == "custom" %}
+    {{param.name}}({{param.internal}}){{ "," if not loop.last }}
+    {%- elif kind == "complex" %}
+    {{param.name}}({{param.value}}, {{param.internal}}){{ "," if not loop.last }}
+    {%- endif %}
+    {%- endfor %};
+
+    {%- if kind == "simple" %}
+
+    /**
```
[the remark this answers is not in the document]
ksologubov: Added additional javadocs as per your comment. No, kind is coming by default from code logic, depending on "value" and "internal" name attributes of enum element. All support for corner cases (mapping code) and mapping.json itself was removed completely to comply with committee decision.

utils/generator/templates/struct_function_template.java
```diff
@@ -0,0 +1,80 @@
+{% extends "base_template.java" %}
+
+{% block body %}
+public class {{class_name}} extends {{extends_class}} {
+    {%- if params is defined %}
+    {%- for p in params %}
+    {%- if p.see is defined or p.deprecated is not none %}
+    /**
+     {%- if p.deprecated is not none %}
+     * @deprecated
```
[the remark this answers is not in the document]
ksologubov: As we removed mapping leaving it as is

utils/generator/templates/function_template.java
```diff
@@ -0,0 +1,62 @@
...
+        super(FunctionID.{{function_id}}.toString());
+    }{% endblock -%}
+
+    {%- block constructor_complex %}
+    {%- if kind is defined and kind == "response" %}
+
+    /**
+     * Constructs a new {{class_name}} object
+     *
+     * @param success    whether the request is successfully processed
```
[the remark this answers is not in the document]
ksologubov: It was the same across all java library, including RPCResponse.java where this methods defined. Although I updated this comments to different ones in generator template.

utils/generator/templates/function_template.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param resultCode whether the request is successfully processed
+     */
+    public {{class_name}}(@NonNull Boolean success, @NonNull Result resultCode) {
```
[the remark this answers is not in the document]
ksologubov: I got the issue, going to fix this

utils/generator/transformers/structs_producer.py
ksologubov: Removed duplication, thanks

utils/generator/test/runner.py
ksologubov: I will add this to github actions file.

utils/generator/requirements.txt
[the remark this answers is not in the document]
bilal-alsharifi: You are right. Sorry about that. This is the working link:
https://docs.google.com/forms/d/1VNR8EUd5b46cQ7uNbCq1fJmnu0askNpUp5dudLKRGpU/viewform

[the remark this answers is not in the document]
ksologubov: We using pylint python library for python code checks/linting. It licensed under GPL license per project page https://github.com/PyCQA/pylint. Could you please confirm we need to remove this dependency and consider other library for code checks (which distributed under supported license)?
All other dependencies is under allowed list of licenses (MIT, BSD3 and Apache2.0)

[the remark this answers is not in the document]
vladmu: @bilal-alsharifi meanwhile we have used PyLint based on the case that it is already used in [smartdevicelink/sdl_core interface_generator](https://github.com/smartdevicelink/sdl_core/blob/master/tools/InterfaceGenerator/requirements.txt) Is that wrong and outdated usage with the unsupported license too?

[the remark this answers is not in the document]
bilal-alsharifi: @theresalech do you know what we should do in that case?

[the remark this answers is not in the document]
ksologubov: By the way, I already tried to replace pylint with flake8 which is licensed by MIT license https://pypi.org/project/flake8/ and it works fine for code checks, so we can replace it without a problem. Both pylint and flake8 are 2 most popular python code check tools so I don't think it will be a problem. flake8 also was among dependencies for sdl_core interface generator.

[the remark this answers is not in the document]
bilal-alsharifi: That is great then. Thanks for doing that.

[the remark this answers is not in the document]
ksologubov: No problem, moved this to PR branch

ksologubov:
There is following discussion in sdl_ios pull request about all params with name "id" should be renamed to "${rpcName}Id" 
https://github.com/smartdevicelink/sdl_ios/pull/1556#issuecomment-592521294 (point 7)
>> The struct TouchEvent has a parameter called id. This is a keyword in objc. The scripts should never use this keyword. A rule should be added if the param name is id the generated objc property should be called "${rpcName}Id" so in the example of touch event it should be touchEventId.
> I'm fine with that approach to keywords. We should define a list of keywords to do this approach with, and do the same for Java_Suite.
Bilal, could you please confirm you are agree with that and I should make this change?

ksologubov:
@bilal-alsharifi 
I finished working on all review comments. Could you please check once more

utils/generator/test/runner.py
ksologubov: Added tests to github actions, build works fine

utils/generator/templates/function_template.java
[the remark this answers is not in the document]
ksologubov: fixed

bilal-alsharifi:
>  could you please confirm you are agree with that and I should make this change?
I don't think that is needed in java suite. Because we add `get` & `set` before the getters and setter names.

bilal-alsharifi:
@ksologubov could you please merge the latest `develop` into your branch? there was an issue in Github Actions setup that making the CI not getting triggered if the PR is from a fork. That has been just fixed and merged #1297.

ksologubov:
@bilal-alsharifi sure, merged, github actions build is in progress

bilal-alsharifi:
@ksologubov thanks for merging develop. The CI actions work well now 👌 
I am noticing an issue with `ShowConstantTBT` RPC. When the script overwrites what we already have. It names the class correctly `ShowConstantTBT` however, the file name stays `ShowConstantTbt.java` which makes the file not a valid java file because of the naming difference.

ksologubov:
@bilal-alsharifi But ShowConstantTBT is not present currently in sdl_java_suite as I see. I believe ShowConstantTbt.java (camel case) is leftover from previous runs of generator when mapping was present. Are you using Windows machine? Because in this case ShowConstantTbt.java and ShowConstantTBT.java is the same file and generator replacing content and leaving name as is. I'm going to add explicit file removal to get rid of this issue on Windows.

ksologubov:
This issue reproduced only on case-insensitive filesystems, e.g. on Windows. Same will be for OASISAddress. I'm going to fix it with explicit file removal before overwriting file content.

ksologubov:
Could you please try now

bilal-alsharifi:
Thanks for the very quick response and fix. I am using Mac OS which should be case-sensitive but I agree it is strange why the issue was happening. Your fix though that deletes the file before overwriting it fixed it 👌

utils/generator/templates/struct_function_template.java
```diff
@@ -0,0 +1,70 @@
...
+        {%- endfor %}
+    }
+    {%- endif %}
+    {%- endif %}
+
+    {%- if params is defined and ((kind is defined and kind != "response") or kind is not defined) %}
+    {%- block setter %}
+    {%- endblock%}
+    {%- endif %}
+
```
bilal-alsharifi: Can we remove the extra line here? this generates an extra line at the end of all functions and structs.

ksologubov: Sure, removed

bilal-alsharifi:
It looks like the script is still using the `internalName` for enums. To make it simpler, as discussed [here](https://github.com/smartdevicelink/sdl_java_suite/pull/1273#issuecomment-588470674), we can completely ignore the `internalName` tag from the RPC spec:

- If the `name` in the RPC spec doesn't start with a number, we can use that name with no modifications and ignore the `internalName` in the RPC spec. For example, `HMILevel` will be:

```
public enum HMILevel {
    FULL,
    LIMITED,
    BACKGROUND,
    NONE,
    ...
}
```

- If the `name` in the RPC spec starts with a number, we can use that name as a value for the enum, and add `_` as a prefix for the developer-facing value. This is still not using the `internalName` from RPC spec. For example:

```
public enum BitsPerSample {
    _8_BIT("8_BIT"),
    _16_BIT("16_BIT");

    private final String VALUE;
    ...
}
```


- If the `value` tag already exists in the RPC spec, we can keep using it as we do now (the script is handling that case in a correct way). For example:

```
public enum LightName {
    FRONT_LEFT_HIGH_BEAM(0),
    FRONT_RIGHT_HIGH_BEAM(1),
    FRONT_LEFT_LOW_BEAM(2),
    ...
   
    private final int VALUE;
    ...
}
```

o-mishch:
@bilal-alsharifi
> * If the `name` in the RPC spec doesn't start with a number, we can use that name with no modifications and ignore the `internal_name` in the RPC spec.

in case when `name` contains character `-`, e.g. enum `Language` should we take name and replace `-` with `_` to avoid java compiling error?

utils/generator/README.md
```diff
@@ -0,0 +1,2256 @@
+# Proxy Library RPC Generator
+
+## Overview
+
+This script provides the possibility to auto-generate Java code based on a given SDL MOBILE_API XML specification.
+
+## Requirements
+
+The script requires Python 3.5 or later pre-installed in the system. This is the minimal Python 3 version that has not reached the end-of-life (https://devguide.python.org/devcycle/#end-of-life-branches).
```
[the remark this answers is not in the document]
vladmu: @bilal-alsharifi the code is compatible with 3.5 thus minimum required version is 3.5. This part just notifies that the generator will not work with previous versions. Even after 3.5 will reach EoL, the code will be compatible with that version until somebody puts the incompatible code. Do we still need this fix here?

vladmu:
> Language has "-" but not in the beginning so it would be generated like
@bilal-alsharifi the question exactly just in your confirmation that `-` converted to `_` is the correct rule for this. We did not use this before as all such cases were covered by `internal_name` attribute. Do you confirm we could use this rule?

bilal-alsharifi:
> > Language has "-" but not in the beginning so it would be generated like
> > @bilal-alsharifi the question exactly just in your confirmation that `-` converted to `_` is the correct rule for this. We did not use this before as all such cases were covered by `internal_name` attribute. Do you confirm we could use this rule?

Oh, my bad I misunderstood the question initially, in that case, we can use the `name` attribute and then replace `-` with `_` for the developer-facing value. but the actual value should never change. `Language`, for example, should be:

```
public enum Language {

    EN_US("EN-US"),
    ES_MX("ES-MX")
    ...
    
    private final String VALUE;

    private Language(String value) {
        this.VALUE = value;
    }
}
```

utils/generator/README.md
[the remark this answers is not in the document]
bilal-alsharifi: but stating that we are going to support a relatively old version will prevent us from using any of the features in the newer python version. Maybe 3.8 is very new. I suggest something in the middle like 3.6 or 3.7. Let's wait on doing that though until we discuss it with the other teams to make sure all platforms align on that.

utils/generator/templates/enum_template.java
```diff
@@ -0,0 +1,126 @@
...
+        try {
+            return valueOf(value);
+        } catch (Exception e) {
+            return null;
+        }
+    }
+    {%- elif kind == "complex" or kind == "custom" %}
+    {%- if return_type == "String" %}
+
+    private final String INTERNAL_NAME;
```
bilal-alsharifi: Can we also rename this var to be `VALUE` instead of `INTERNAL_NAME`?

utils/generator/templates/enum_template.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Private constructor
+     */
+    private {{class_name}}(String internalName) {
```
bilal-alsharifi: and also rename this var to be `value`

o-mishch: class_name renamed to value

utils/generator/templates/enum_template.java
o-mishch: done

utils/generator/templates/enum_template.java
bilal-alsharifi: @o-mishch Thanks for making the change. I meant the `internalName` should be renamed to value. Sorry I did not mean the class name. Can we change the class name back and just change `internalName` to `value`. and `INTERNAL_NAME` to `VAULE`

utils/generator/templates/enum_template.java
bilal-alsharifi: Thanks

utils/generator/templates/enum_template.java
bilal-alsharifi: Like the following :
```
private {{class_name}}(String value) {
	this.VALUE = value;
}
```

o-mishch: done

bilal-alsharifi: Thanks

utils/generator/README.md
```diff
@@ -0,0 +1,2256 @@
+# Proxy Library RPC Generator
+
+## Overview
```
bilal-alsharifi: The `README.md` seems to be a bit outdated now. Can we update it to reflect the latest changes like mapping removal and enum simplifications?

utils/generator/transformers/enums_producer.py
```diff
@@ -0,0 +1,118 @@
...
+        :return: dictionary to be applied to jinja2 template
+        """
+        imports = set()
+        params = OrderedDict()
+        if any(map(lambda l: l.name != self.converted(l.name) and getattr(l, 'value', None) is not None,
+                   getattr(item, self.container_name).values())):
+            kind = 'complex'
+            imports.add('java.util.EnumSet')
+        elif any(map(lambda l:
+                     l.name != self.converted(l.name) or l.name[0].isdigit() or getattr(l, 'value', None) is not None,
```
bilal-alsharifi: do you think we can remove this `l.name[0].isdigit()` from the condition since it is already covered in `l.name != self.converted(l.name)`?

utils/generator/transformers/enums_producer.py
```diff
@@ -0,0 +1,118 @@
...
+    def transform(self, item: Enum) -> dict:
+        """
+        Override
+        :param item: particular element from initial Model
+        :return: dictionary to be applied to jinja2 template
+        """
+        imports = set()
+        params = OrderedDict()
+        if any(map(lambda l: l.name != self.converted(l.name) and getattr(l, 'value', None) is not None,
+                   getattr(item, self.container_name).values())):
```
bilal-alsharifi: I am just wondering about what enums have the `complex`  kind? I don't see any of them do. However, if I made a case like that in the `MOBILE_API.xml`. Like this one, for example, the generator will not work well in that case:

```
<enum name="PredefinedWindows" since="6.0">
    <element name="0DEFAULT_WINDOW" value="0">
        <description>The default window is a main window pre-created on behalf of the app.</description>
    </element>
    <element name="1PRIMARY_WIDGET" value="1">
        <description>The primary widget of the app.</description>
    </element>
</enum>
```
Maybe I missing something.

utils/generator/templates/struct_function_template.java
```diff
@@ -0,0 +1,69 @@
...
+     {%- if p.see is defined %}
+     * @see {{p.see}}
+     {%- endif %}
+     */
+    {%- endif %}
+    {%- if p.deprecated is not none %}
+    @Deprecated
+    {%- endif %}
+    {%- if p.name is defined %}
+    private {% if p.modifier is defined %}{{p.modifier}} {% endif %}{{p.return_type}} {{p.name}}{% if p.value is defined %} = {{p.value}}{% endif %};
```
bilal-alsharifi: I am a little confused about what that generates exactly.

utils/generator/templates/struct_function_template.java
```diff
@@ -0,0 +1,69 @@
+{% extends "base_template.java" %}
```
bilal-alsharifi: This template has some issues with generating responses. All the responses are generated without any getter or setters.

o-mishch: there was redundant check:
```jinja
    {%- if params is defined and ((kind is defined and kind != "response") or kind is not defined) %}
```
After deleting works well

utils/generator/templates/struct_function_template.java
o-mishch: I didn't understand about what exactly line of code you are talking about, e.g. following xml:
```xml
<struct name="DisplayCapabilities" deprecated="true" since="6.0">
        <param name="displayType" type="DisplayType" mandatory="true" deprecated="true" since="5.0">
```
will be represent as following java code:
```java
@Deprecated
public class DisplayCapabilities extends RPCStruct {
    /**
     * @deprecated
     */
    @Deprecated
    public static final String KEY_DISPLAY_TYPE = "displayType";
```
If you're mean following: 
```jinja
{% if p.modifier is defined %}{{p.modifier}} {% endif %}
```
I guess it was added for some kind of manual customisation, and after we decide to do any manual customisation I believe it also useless.
Deleting it.

utils/generator/transformers/enums_producer.py
o-mishch: the idea of that type was to cover cases like that: 
```bash
sdl_java_suite/base/src/main/java/com/smartdevicelink/protocol/enums/FunctionID.java
```
```java
SYNC_P_DATA(65537, "SyncPData"),
```
but as we agreed to skip such cases, I agree that `complex` kind is useless.
Deleting it.

utils/generator/transformers/enums_producer.py
o-mishch: yes, you right. 
Deleting it.

utils/generator/README.md
o-mishch: @vladmu could you have look

utils/generator/templates/struct_function_template.java
bilal-alsharifi: I am referring to lines 20 & 21. It is adding a field with a `private` modifier. However, I am not sure in which case an RPC can have a field with a private modifier.

o-mishch:
```jinja
.name
```
was using for manual customisation.
Deleting it.

utils/generator/README.md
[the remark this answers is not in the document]
vladmu: @bilal-alsharifi the document states "3.5 or later" therefore, I don't think this note prevents you from using the newest features if needed. This only about the current state of the generator, but if someone puts incompatible features in the future, the document should be adjusted simultaneously.

utils/generator/templates/struct_function_template.java
ksologubov: But in original RPC library all function responses does not have any getters/setters as well as parameters list. Only requests have getters/setters and parameters list. For responses parameters "success" and "resultCode" defined in RPCResponse.

bilal-alsharifi: The shared getters/ setters are placed in the base RPCResponse class. But also each response has its own getters and setters. For example:
https://github.com/smartdevicelink/sdl_java_suite/blob/797b7a14a7fd67fdf6faa504865e4e22228858ef/base/src/main/java/com/smartdevicelink/proxy/rpc/GetVehicleDataResponse.java#L108

ksologubov: Got it, thanks, working on it

utils/generator/README.md
vladmu:
```suggestion
The script requires **Python 3** pre-installed on the host system. The minimal supported Python 3 version is **3.7.6**. It may work on versions back to 3.5 (the minimal version that has not yet reached [the end-of-life](https://devguide.python.org/devcycle/#end-of-life-branches)), but this is not supported and may break in the future.
```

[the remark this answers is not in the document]
vladmu: @bilal-alsharifi done in https://github.com/smartdevicelink/sdl_java_suite/pull/1273/commits/7599c4180f36c420179a492e5eb09cd6c34e5eda the same way as discussed with @joeljfischer in iOS PR https://github.com/smartdevicelink/sdl_ios/pull/1556#discussion_r390300250

utils/generator/README.md
ksologubov: Readme updated

utils/generator/README.md
```diff
@@ -0,0 +1,2245 @@
...
+* `@since` should be present, if the `"since"` attribute exists, and `[since_version]` is the `Major.Minor.Patch` formatted value of this attribute.
+* `@see` shows the custom reference in `[see_reference]`.
+
+The class should have the `@Deprecated` decorator if the `"deprecated"` attribute of the `<enum>` exists and is "true".
+
+The set of `<element>` should be mapped to the set of Enum constants. Based on the `<element>` attributes, constants could be with or without fields.
+
+The following list are general rules for constant names and its fields:
+1. The `"name"` attribute of `<element>` is the default name of the constant.
+    * if the name starts from digit, the leading `_` (underscore) symbol should be added.
```
bilal-alsharifi: I think that case doesn't exist anymore in the code. Also, I believe, the case where we replace "-" with "_" is not mentioned in the readme.

ksologubov: No, this exists because there is Enums which will fail without it, e.g. BitsPerSample
element name="8_BIT" should be converted to _8_BIT

bilal-alsharifi: I meant the other point:
> if the name starts with _ (underscore) symbol and the next character is a letter of alphabet, the leading _ (underscore) symbol should be removed.

do we still have this case?

bilal-alsharifi: Also, I am not sure why, but the CI doesn't pass on this PR.

ksologubov: Thanks, Bilal, updated on both questions, waiting for current build to see the problem.

bilal-alsharifi: Awesome. It passes now. Thanks.

utils/generator/README.md
```diff
@@ -0,0 +1,2245 @@
...
+    * if the name consists of `-` (dash) then it replaced with `_` (underscore)
```
bilal-alsharifi:
```suggestion
    * if the name contains a `-` (dash) then it should be replaced with `_` (underscore)
```

utils/generator/README.md
ksologubov: As I see from history it was failing either on android artifacts download with 504 error or git checkout which is far from RPC generator changes.

bilal-alsharifi: Yes, that is strange. it fails again now and keeps failing when I restart it.
Failing on downloading the Android artifacts is normal and happens rarely sometimes but simply restarting the action fixes it. However, I have never seen it failing on checkout before. I am guessing it is something has to do with checking out out the submodule that was added here:
https://github.com/smartdevicelink/sdl_java_suite/blob/07ad3eee3682e0ce9306b523821f0d3e85dbcc95/.github/workflows/android.yml#L13

Maybe try to remove that line (just for testing) and see if it passes the checkout stage. There are other alternatives available online to checkout submodules if that is what is causing the issue.

ksologubov: Agree, removed my android.yml changes, let's see if build will be stable without checking out submodules

bilal-alsharifi: It is still failing on the checkout stage which is strange. I am still guessing it is something has to do with submodules addition because it passes on all other PRs. Can you please try to remove the `.gitmodules` (just temporarily) to see if that makes it pass?

bilal-alsharifi: It could be a bug in the `checkout` action. I have made PR #1305  to update the actions `android.yml` to the latest version. It is merged into `develop` now. Can you merge the latest develop into your branch to see if that fixes the issue?

ksologubov: I was able to fix this with checkout@master action. Details is in last comments of https://github.com/actions/checkout/issues/81
In short, checkout@v2 did not work with submodules at all now, but updated submodule processing was already comitted to master of checkout action. It will be moved soon to v2 tag

bilal-alsharifi: Awesome. Thanks for doing that research. We can keep the master tag for now and change to v2 later when the fix is merged.

vladmu:
@bilal-alsharifi we managed all comments, do you have additional remarks for this PR? Or could you proceed with the approval/merge?

bilal-alsharifi:
@vladmu We are waiting for the iOS PR to be done in case the iOS team suggested a change that applies to the java suite one. We would like to make sure that the two PRs align before merging them.

bilal-alsharifi:
Also, we found out that the generated `CancelInteraction` RPC doesn't compile because it has a method called `getFunctionID()` that clashes with the method in `RPCMessage`.
We will probably need to put a list of reserved keywords to fix that case. We are discussing that with the iOS team to see the best way to handle that. We will keep you updated.

bilal-alsharifi:
Because of the `getFunctionID()` issue that we mentioned in the previous comment and some other  similar name conflicting issues in iOS, we added a list of reserved keywords that RPC Generators in all platforms (iOS, Java Suite, JavaScript Suite) should avoid using by following these rules:

- If the generator needs to create a variable or method and the name was one of the names in the reserved keyword, the generator should append the suffix `Param` to it. For example `getFunctionID`  > `getFunctionIDParam`
- Any empty line in the reserved keywords file should be ignored
- Any line in the reserved keywords file that starts with one or more `#` symbols then `space` should be considered a comment and ignored (like `#<space>` or `##<space>` or `###<space>` …). Note that some reserved keywords have a `#` without a `<space>` so it can't just check for `#`.
- Reserved keywords matching should be case insensitive
- Although the reserved keywords list has a section for each platform, the generator for each platform should avoid using any of the keywords whether it belongs to that platform or another one.

The reserved keywords list is part of the `rpc_sepc` repo so all generators can access it. Please look at this [PR](https://github.com/smartdevicelink/rpc_spec/pull/232) for more info about the reserved keywords.

utils/generator/README.md
```diff
@@ -0,0 +1,2245 @@
+# Proxy Library RPC Generator
```
bilal-alsharifi: Can we make the readme file more concise similar to how the iOS one?
https://github.com/vladmu/sdl_ios/blob/SDL-0234-Proxy-Library-RPC-Generation/generator/README.md

It is currently very lengthy and it will be hard to keep it updated in the future whenever we make an update to the generator.

bilal-alsharifi:
Also after a discussion with the iOS and Core teams, we found out that all handling of "numeric" enums should be removed because Core expects string values. This has already been done in the iOS RPC Generator and should be reflected in Java Suite in a similar way. Please look at this [comment](https://github.com/smartdevicelink/sdl_ios/pull/1556#discussion_r397150161) for more details.

utils/generator/README.md
vladmu: @bilal-alsharifi it is lengthy due to numerous full examples with big copyright comments included in the document, will the removing of those examples solve the problem?

ksologubov: If we remove all full examples as @vladmu suggested, README size will decrease 3 times, from 2450 lines to only 850 lines which is the same as we have in sdl_ios. @bilal-alsharifi please confirm if this looks good for you.

bilal-alsharifi: I think samples are important to explain the rules. iOS also has sample codes. We can use fewer samples, pick classes with fewer properties as samples, and remove the license header form the samples in the readme to make the readme file shorter and easier to maintain. Please use the iOS one as a reference:
https://github.com/vladmu/sdl_ios/tree/SDL-0234-Proxy-Library-RPC-Generation/generator

vladmu: @bilal-alsharifi that iOS document was based on this Java document, and it just does not include the full examples, almost all other sections and their order are similar. So it is hard to understand what you want to change here. I've proposed above to remove FULL examples from the document, all SHORT samples stay in place. Please review simplified readme @ksologubov provided [here (7da2be7)](https://github.com/smartdevicelink/sdl_java_suite/blob/7da2be78ce3d0b0e6e2381e33d972cabcd9377ec/utils/generator/README.md)  and if that doesn't fit your expectation, please provide us more details about what you would like to change exactly. Thank you!

bilal-alsharifi: Thanks, @vladmu for the details. I looked at the android simplified readme. I don't see any generated java sample code at all in it. I see snippets from the templates but not java code. However, in iOS, there are some full but concise samples like [this](https://github.com/vladmu/sdl_ios/tree/SDL-0234-Proxy-Library-RPC-Generation/generator#request-functions-sdlrpcrequest) for example. We just would like the readme to be easy to read but not extremely long.
As I suggested earlier, using fewer samples, picking classes with fewer properties as samples, and removing the license header from the samples may reduce the size of the file but still keep it self contained.

bilal-alsharifi: Like for example, before today's updates, the readme had [three](https://github.com/smartdevicelink/sdl_java_suite/tree/744d2bfd76beb8f9adcc3158dca176ef05da270a/utils/generator#below-is-the-full-example-of-the-struct-class-with-simple-and-enum-parameters-inside) lengthy full sample for the `<struct>` section. After the update, they are all removed. I think having one concise sample that covers the basic cases will be a good balance between the two cases.

ksologubov: Ok, got it, thank you, will update with basic samples

ksologubov: @bilal-alsharifi Hi Bilal, I updated README with samples, could you please check

bilal-alsharifi: Thank you. I will review that.

ksologubov:
Thanks @bilal-alsharifi. Corrected all cases.

utils/generator/README.md
bilal-alsharifi: @ksologubov can we try v2 tag again? it seems to be fixed now.

ksologubov: @bilal-alsharifi  set to v2, works good

bilal-alsharifi: Awesome. Thanks

bilal-alsharifi:
Can we put the RPC generator code in `/generator` directory instead of `utils/generator` to make it more aligned with `sdl_ios` and `sdl_javascript_suite`?

ksologubov:
> Can we put the RPC generator code in `/generator` directory instead of `utils/generator` to make it more aligned with `sdl_ios` and `sdl_javascript_suite`?

Ok, done

ksologubov:
> Because of the `getFunctionID()` issue that we mentioned in the previous comment and some other similar name conflicting issues in iOS, we added a list of reserved keywords that RPC Generators in all platforms (iOS, Java Suite, JavaScript Suite) should avoid using by following these rules:
> 
> * If the generator needs to create a variable or method and the name was one of the names in the reserved keyword, the generator should append the suffix `Param` to it. For example `getFunctionID`  > `getFunctionIDParam`
> * Any empty line in the reserved keywords file should be ignored
> * Any line in the reserved keywords file that starts with one or more `#` symbols then `space` should be considered a comment and ignored (like `#<space>` or `##<space>` or `###<space>` …). Note that some reserved keywords have a `#` without a `<space>` so it can't just check for `#`.
> * Reserved keywords matching should be case insensitive
> * Although the reserved keywords list has a section for each platform, the generator for each platform should avoid using any of the keywords whether it belongs to that platform or another one.
> 
> The reserved keywords list is part of the `rpc_sepc` repo so all generators can access it. Please look at this [PR](https://github.com/smartdevicelink/rpc_spec/pull/232) for more info about the reserved keywords.

@bilal-alsharifi Keywords logic was applied to PR. Could you please check from your side?

bilal-alsharifi:
Thanks, @ksologubov for making the changes. It looks like at some point in the recent updates to the PR. The script stopped setting the default `OUTPUT_DIRECTORY` automatically.  So when I run this command now:
```
python3 ./generator.py -y
```
I get this message:
```
usage: generator.py [-h] [-v] [-xml SOURCE_XML] [-xsd SOURCE_XSD] -d
                    OUTPUT_DIRECTORY [-t [TEMPLATES_DIRECTORY]]
                    [-r REGEX_PATTERN] [--verbose] [-e] [-s] [-m] [-y] [-n]
generator.py: error: the following arguments are required: -d/--output-directory
```
It should always default to the correct `OUTPUT_DIRECTORY` if no other directory is specified using the `-d` argument

bilal-alsharifi:
@ksologubov While I am testing the new changes, I found that in a lot of cases, the generator appends "Param" when it shouldn't. Here are a few examples:

#### Example 1:
```
public class HapticRect extends RPCStruct {
    public static final String KEY_ID_PARAM = "idParam";
}
```
Note: The JSON values should _never_ change:

#### Example 2:
```
public GetAppServiceDataResponse(@NonNull Boolean successParam, @NonNull EnumSubset resultCodeParam) {
    this();
    setSuccessParam(successParam);
    setResultCodeParam(resultCodeParam);
} 
```
- Note: These params/setters shouldn't have "Param" since they are calling methods in predefined base classes. Also I am not sure why result is of type `EnumSubset` instead of `Result`

#### Example 3:
```
public class CancelInteraction extends RPCRequest {
    public static final String KEY_FUNCTION_IDPARAM = "functionID";
}
```
- Note: The key should be `KEY_FUNCTION_ID_PARAM` not `KEY_FUNCTION_IDPARAM`


The aforementioned snippets are just samples and not a comprehensive list. I recommend using a directory comparison tool (I use Meld) to compare the RPC output before the reserved keywords changes and after them to confirm that only cases that actually need to be changed have changed.

ksologubov:
> #### Example 2:
> 
> ```
> public GetAppServiceDataResponse(@NonNull Boolean successParam, @NonNull EnumSubset resultCodeParam) {
>     this();
>     setSuccessParam(successParam);
>     setResultCodeParam(resultCodeParam);
> } 
> ```
> 
>     * Note: These params/setters shouldn't have "Param" since they are calling methods in predefined base classes. Also I am not sure why result is of type `EnumSubset` instead of `Result`
> 
> 

@bilal-alsharifi Bilal, get/setSuccess and get/setResultCode both explicitly specified in keywords list. From what you said it should not be there. What should we do in this case? I believe better to remove them from keywords list.

bilal-alsharifi:
I understand the confusion. In that example, they are calling methods from the base class. The reserved keywords rules apply to APIs in the subclasses (RPCs) but don't apply to APIs in base classes (`RPCMessage`, `RPCStruct`, `RPCNotification`, `RPCRequest`, `RPCResponse`). The `setSuccess()`, `setResultCode()` are APIs in the base class so they should not be affected by the reserved keywords rules.

```
public GetAppServiceDataResponse(@NonNull Boolean success, @NonNull Result resultCode){
	this();
	super.setSuccess(success);
	super.setResultCode(resultCode);
}
`

**Comment by ksologubov** [comment]
@bilal-alsharifi but according to https://github.com/smartdevicelink/rpc_spec/pull/232
> If the generator needs to create a variable or method and the name was one of the names in the reserved keyword, the generator should append the suffix Param to it. For example getFunctionID > getFunctionIDParam
In this case generator working exactly as proposal says by replacing setSuccess keyword in method name with setSuccessParam and so on. "super" was not used in original library implementation so I'm not sure why you provided it above. Please clarify why setSuccess/setResultCode/success/resultCode should not be removed from keyword list if it should not be replaced. I don't see how it's used in any other cases except calling methods from RPCResponse base class from response type functions.

**Comment by bilal-alsharifi** [comment]
@ksologubov I think you are right. I removed `info`, `success`, `resultCode` from reserved keywords. I will do another review and let you know if there is any more feedback.

**Comment by bilal-alsharifi** [comment]
@ksologubov I see `set/getFunctionId()` inside the class `CancelInteraction` are still not affected by the reserved keywords. They should be `set/getFunctionIdParam()` as they are the main reason for adding the reserved words.

**Comment by bilal-alsharifi** [comment]
Can we also pull the latest changes for the rpc_spec submodule after removing the response keywords?

File: generator/transformers/common_producer.py [hunk 28]
```diff
@@ -0,0 +1,120 @@
...
+    def replace_sync(name):
+        """
+        :param name: string with item name
+        :return: string with replaced 'sync' to 'Sdl'
+        """
+        if name:
+            return re.sub(r'^([sS])ync(.+)$', r'\1dl\2', name)
+        return name
+
+    def replace_keywords(self, name: str = '') -> str:
```
bilal-alsharifi, line 89: The comparison in this method seems to be case sensitive. It should be insensitive.

ksologubov: @bilal-alsharifi restored case insensitive matching, and also fixed functionID parameter replacement, thanks

ksologubov:
@bilal-alsharifi thanks for removing extra keywords. All above problems was fixed, and suggestions was applied, I checked classes diff with Meld and they looking good from my side, could you please check from your side?

bilal-alsharifi:
@ksologubov Thanks for making the changes. I have reviewed and tested the PR. It looks good now.

ksologubov:
@bilal-alsharifi thank you very much for reviews

PR #1277, opened 2020-02-17
Update SCM to notify listeners if fresh copy of capability is received

### LINKED ISSUE #1275: SystemCapabilityManager should call listeners if GetSystemCapabilityResponse is received
### Bug Report
If a capability listener is added using `addOnSystemCapabilityListener()` then the developer calls `getCapability()` and sets `forceUpdate` to `true`, that doesn't notify the listener that was added previoulsy.

##### Reproduction Steps
1. Use `SCM.addOnSystemCapabilityListener()` to add a listener for `NAVIGATION` capability for example
2. Call `SCM.getCapability(SystemCapabilityType.NAVIGATION, ..., true)`
3. Check if the listener that was added in step 1  is called 

##### Expected Behavior
The listener that is added in step 1 should be called

##### Observed Behavior
The listener doesn't get called 

* Android Version: 10
* SDL Android Version: 4.10.1
* Testing Against: Core develop branch

bilal-alsharifi:
Fixes #1275 

### Risk

### Testing Plan
- [x] I have tested Android

#### Unit Tests
Added more unit tests to test the fix

#### Core Tests
1. Use `SCM.addOnSystemCapabilityListener()` to add a listener for `NAVIGATION` capability for example
2. Make sure the listener is called
3. Call `SCM.getCapability(SystemCapabilityType.NAVIGATION, ..., true)`
3. Confirm the listener that was added in step 1  is called again 

### Summary
This PR fixes an issue with `addOnSystemCapabilityListener()` method that causes the listeners not to be triggered if a fresh copy of the capability is received when the developer calls `getCapability()` and sets `forceUpdate` to `true`

### CLA

### DIFF
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
@@ -608,7 +608,7 @@ public class SystemCapabilityManagerTests extends AndroidTestCase2 {
 
 
 		// Test case 3 (Remove listener)
-		// When the last DISPLAY listener is removed, GetSystemCapability request should not go out
+		// When the last DISPLAYS listener is removed, GetSystemCapability request should not go out
 		scm.removeOnSystemCapabilityListener(SystemCapabilityType.DISPLAYS, onSystemCapabilityListener1);
 		verify(internalInterface, times(0)).sendRPC(any(GetSystemCapability.class));
 	}
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
@@ -490,6 +490,94 @@ public class SystemCapabilityManagerTests extends AndroidTestCase2 {
 		verify(internalInterface, times(1)).sendRPC(any(GetSystemCapability.class));
 	}
 
+	public void testAddOnSystemCapabilityListenerThenGetCapability() {
+		SdlMsgVersion sdlMsgVersion = new SdlMsgVersion(5, 0); // This version doesn't support capability subscriptions
+		sdlMsgVersion.setPatchVersion(0);
+		ISdl internalInterface = mock(ISdl.class);
+		when(internalInterface.getSdlMsgVersion()).thenReturn(sdlMsgVersion);
+		SystemCapabilityManager scm = new SystemCapabilityManager(internalInterface);
+		scm.setCapability(SystemCapabilityType.VIDEO_STREAMING, videoStreamingCapability);
+
+
+		// Add listener1
+		// When the first listener is added, GetSystemCapability request should go out with subscribe=true
+		OnSystemCapabilityListener onSystemCapabilityListener1 = mock(OnSystemCapabilityListener.class);
+		doAnswer(createOnSendGetSystemCapabilityAnswer(true, false)).when(internalInterface).sendRPC(any(GetSystemCapability.class));
+		scm.addOnSystemCapabilityListener(SystemCapabilityType.VIDEO_STREAMING, onSystemCapabilityListener1);
+		verify(internalInterface, times(0)).sendRPC(any(GetSystemCapability.class));
+		verify(onSystemCapabilityListener1, times(1)).onCapabilityRetrieved(any(Object.class));
+
+
+		// Get Capability (should notify listener1 again)
+		scm.getCapability(SystemCapabilityType.VIDEO_STREAMING, null, true);
+		verify(internalInterface, times(1)).sendRPC(any(GetSystemCapability.class));
+		verify(onSystemCapabilityListener1, times(2)).onCapabilityRetrieved(any(Object.class));
+
+
+		// Add listener2
+		OnSystemCapabilityListener onSystemCapabilityListener2 = mock(OnSystemCapabilityListener.class);
+		scm.addOnSystemCapabilityListener(SystemCapabilityType.VIDEO_STREAMING, onSystemCapabilityListener2);
+		verify(onSystemCapabilityListener2, times(1)).onCapabilityRetrieved(any(Object.class));
+
+
+		// Get Capability (should notify listener1 & listener2 again)
+		scm.getCapability(SystemCapabilityType.VIDEO_STREAMING, null, true);
+		verify(internalInterface, times(2)).sendRPC(any(GetSystemCapability.class));
+		verify(onSystemCapabilityListener1, times(3)).onCapabilityRetrieved(any(Object.class));
+		verify(onSystemCapabilityListener2, times(2)).onCapabilityRetrieved(any(Object.class));
+
+
+		// Add listener3
+		OnSystemCapabilityListener onSystemCapabilityListener3 = mock(OnSystemCapabilityListener.class);
+		scm.addOnSystemCapabilityListener(SystemCapabilityType.VIDEO_STREAMING, onSystemCapabilityListener3);
+		verify(onSystemCapabilityListener3, times(1)).onCapabilityRetrieved(any(Object.class));
+
+
+		// Get Capability (should notify listener1 & listener2 & listener3 again)
+		scm.getCapability(SystemCapabilityType.VIDEO_STREAMING, null, true);
+		verify(internalInterface, times(3)).sendRPC(any(GetSystemCapability.class));
+		verify(onSystemCapabilityListener1, times(4)).onCapabilityRetrieved(any(Object.class));
+		verify(onSystemCapabilityListener2, times(3)).onCapabilityRetrieved(any(Object.class));
+		verify(onSystemCapabilityListener3, times(2)).onCapabilityRetrieved(any(Object.class));
+
+
+		// Remove listener1
+		scm.removeOnSystemCapabilityListener(SystemCapabilityType.VIDEO_STREAMING, onSystemCapabilityListener1);
+
+
+		// Get Capability (should notify listener2 & listener3 again)
+		scm.getCapability(SystemCapabilityType.VIDEO_STREAMING, null, true);
+		verify(internalInterface, times(4)).sendRPC(any(GetSystemCapability.class));
+		verify(onSystemCapabilityListener1, times(4)).onCapabilityRetrieved(any(Object.class));
+		verify(onSystemCapabilityListener2, times(4)).onCapabilityRetrieved(any(Object.class));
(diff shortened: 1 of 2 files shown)


base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -549,18 +550,27 @@ private void retrieveCapability(final SystemCapabilityType systemCapabilityType,
...
 				if (response.getSuccess()) {
 					Object retrievedCapability = ((GetSystemCapabilityResponse) response).getSystemCapability().getCapabilityForType(systemCapabilityType);
-					cachedSystemCapabilities.put(systemCapabilityType, retrievedCapability);
+					setCapability(systemCapabilityType, retrievedCapability);
+					// If the listener is not included in the onSystemCapabilityListeners map, then notify it
 					if (scListener != null) {
-						scListener.onCapabilityRetrieved(retrievedCapability);
+						synchronized (LISTENER_LOCK) {
+							CopyOnWriteArrayList<OnSystemCapabilityListener> notifiedListeners = onSystemCapabilityListeners.get(systemCapabilityType);
+							boolean listenerAlreadyNotified = notifiedListeners != null && notifiedListeners.contains(scListener);
```
joeljfischer:
```suggestion
							boolean listenerAlreadyNotified = (notifiedListeners != null) && notifiedListeners.contains(scListener);
```

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
```diff
@@ -490,6 +490,94 @@ public void testAddOnSystemCapabilityListenerWithSubscriptionsNotSupportedAndCap
...
+		SdlMsgVersion sdlMsgVersion = new SdlMsgVersion(5, 0); // This version doesn't support capability subscriptions
+		sdlMsgVersion.setPatchVersion(0);
+		ISdl internalInterface = mock(ISdl.class);
+		when(internalInterface.getSdlMsgVersion()).thenReturn(sdlMsgVersion);
+		SystemCapabilityManager scm = new SystemCapabilityManager(internalInterface);
+		scm.setCapability(SystemCapabilityType.VIDEO_STREAMING, videoStreamingCapability);
+
+
+		// Add listener1
+		// When the first listener is added, GetSystemCapability request should go out with subscribe=true
```
joeljfischer: Based on the change to `retrieveCapability()` I believe this should _not_ send `subscribe=true`, right?

```java
final boolean shouldSubscribe = ((subscribe != null) ? subscribe : isSubscribedToSystemCapability(systemCapabilityType)) && supportsSubscriptions();
request.setSubscribe(shouldSubscribe);
```

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -450,16 +450,17 @@ private boolean isSubscribedToSystemCapability(SystemCapabilityType systemCapabi
...
 	private Object getCapabilityPrivate(final SystemCapabilityType systemCapabilityType, final OnSystemCapabilityListener scListener, final Boolean subscribe, final boolean forceUpdate) {
 		Object cachedCapability = cachedSystemCapabilities.get(systemCapabilityType);
-		OnSystemCapabilityListener listener = scListener;
 
-		if (!forceUpdate && cachedCapability != null && listener != null) {
-			listener.onCapabilityRetrieved(cachedCapability);
-			listener = null;
+		boolean shouldUpdateSystemCapabilitySubscription = (subscribe != null) && (subscribe != isSubscribedToSystemCapability(systemCapabilityType)) && supportsSubscriptions();
+		boolean shouldSendGetCapabilityRequest = forceUpdate || (cachedCapability == null) || shouldUpdateSystemCapabilitySubscription;
+		boolean shouldCallListenerWithCachedValue = (cachedCapability != null) && (scListener != null) && !shouldSendGetCapabilityRequest;
```
joeljfischer: 1. Currently, iOS calls the listener with a cached value of `nil`. This may be a point of alignment needed.
2. Why does it not call with the cached value if we are sending a `GetCapability` request? Should the `&&` there be an `||`?

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -549,18 +550,27 @@ private void retrieveCapability(final SystemCapabilityType systemCapabilityType,
 		}
 		final GetSystemCapability request = new GetSystemCapability();
 		request.setSystemCapabilityType(systemCapabilityType);
-		request.setSubscribe((subscribe != null) ? subscribe : isSubscribedToSystemCapability(systemCapabilityType)); // If subscribe is null, then don't change the current subscription status
+		// If subscribe is null, then don't change the current subscription status
+		final boolean shouldSubscribe = ((subscribe != null) ? subscribe : isSubscribedToSystemCapability(systemCapabilityType)) && supportsSubscriptions();
```
joeljfischer:
```suggestion
		final boolean shouldSubscribe = ((subscribe != null) ? subscribe : (isSubscribedToSystemCapability(systemCapabilityType)) && supportsSubscriptions());
```

I think the name of this is a bit misleading, because if it's `false`, it `shouldUnsubscribe()`. I almost think you need an enum with values `subscribe`, `unsubscribe`, and `keepCurrentState`. Or at least the documentation needs to be improved here.

I also need clarification on a point here. If `subscribe` is `false` and we are currently subscribed, and we support subscriptions, then `shouldSubscribe` will be `true`, right? And that's the opposite of what we want, since we want to unsubscribe?

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -335,11 +335,11 @@ public void onReceived(RPCMessage message) {
 	 * Sets a capability in the cached map. This should only be done when an RPC is received and contains updates to the capability
 	 * that is being cached in the SystemCapabilityManager.
 	 * @param systemCapabilityType the system capability type that will be set
-	 * @param capability the value of the capability that will be set
+	 * @param capability           the value of the capability that will be set
```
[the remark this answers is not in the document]
bilal-alsharifi: Yes, you are right, we don't usually do that. I used Android Studio Reformatter tool to reformat this method. It looks like the tool adds spaces to align the params. I will remove that for consistency.

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
bilal-alsharifi: Ah, you are right, the test actually asserts that `subscribe=false` as you mentioned. The comment is wrong. I will update it.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -549,18 +550,27 @@ private void retrieveCapability(final SystemCapabilityType systemCapabilityType,
...
+		final boolean shouldSubscribe = ((subscribe != null) ? subscribe : isSubscribedToSystemCapability(systemCapabilityType)) && supportsSubscriptions();
+		request.setSubscribe(shouldSubscribe);
 		request.setOnRPCResponseListener(new OnRPCResponseListener() {
 			@Override
 			public void onResponse(int correlationId, RPCResponse response) {
 				if (response.getSuccess()) {
 					Object retrievedCapability = ((GetSystemCapabilityResponse) response).getSystemCapability().getCapabilityForType(systemCapabilityType);
-					cachedSystemCapabilities.put(systemCapabilityType, retrievedCapability);
+					setCapability(systemCapabilityType, retrievedCapability);
+					// If the listener is not included in the onSystemCapabilityListeners map, then notify it
```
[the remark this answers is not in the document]
bilal-alsharifi: Forcing an update is not the only case that triggers this logic. If `forceUpdate=false`, and the cap is not cashed, when we call `GetCapablity()`, this logic will be triggered. But you are right the comment is a little vague. I will update the comment.

android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/SystemCapabilityManagerTests.java
```diff
@@ -490,6 +490,129 @@ public void testAddOnSystemCapabilityListenerWithSubscriptionsNotSupportedAndCap
 		verify(internalInterface, times(1)).sendRPC(any(GetSystemCapability.class));
 	}
 
+	public void testAddOnSystemCapabilityListenerThenGetCapability() {
```
joeljfischer:
```suggestion
	public void testAddOnSystemCapabilityListenerThenGetCapabilityWhenSubscriptionsAreNotSupported() {
```

You may also want to consider underscores to define conditions, like:

`testAddOnSystemCapabilityListener_thenGetCapability_whenSubscriptionsAreNotSupported()`

bilal-alsharifi: The use of underscores in java is not very common. Except for naming const variables. I prefer the first suggested name as it complies more with Java naming conventions.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -550,7 +551,13 @@ private void retrieveCapability(final SystemCapabilityType systemCapabilityType,
 		}
 		final GetSystemCapability request = new GetSystemCapability();
 		request.setSystemCapabilityType(systemCapabilityType);
-		// If subscribe is null, then don't change the current subscription status
+
+		/*
+		The subscription flag in the request should be set based on multiple variables:
+		- if subscribe is null (no change), shouldSubscribe = current subscription status
+		- if subscribe is false, then shouldSubscribe = false
+		- if subscribe is true and the HU supports subscriptions, then shouldSubscribe = true
```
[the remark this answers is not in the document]
bilal-alsharifi: But if the HU doesn't support subscriptions, `shouldSubscribe` will be `false` right? even if `subscribe = true`? because we are doing && supportsSubscriptions()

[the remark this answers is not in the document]
joeljfischer: Based on the current code, I don't think that's true.

```java 
if (subscribe != null) {
    return subscribe;
} else {
    return isSubscribedToSystemCapability(systemCapabilityType)) && supportsSubscriptions();
}
```

So if `subscribe == true`, then it will always return `true`.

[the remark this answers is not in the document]
bilal-alsharifi: It is actually confusing. I will simplify the condition a bit.

PR #1303, opened 2020-03-05
Add overwrite property

### LINKED ISSUE #1302: No Overwrite property for SdlFiles
### Bug Report
IOS has an overwrite property for SdlFiles, that will block files from being uploaded more than once if their overwrite property is set to false. Android does not have this feature

##### Expected Behavior
Android and IOS to be aligned and have the same behavior.

##### Observed Behavior
Android and IOS upload file behavior is different.

JulianKast:
Fixes #1302 

### Risk

### Testing Plan

#### Unit Tests
Added unit test to test file not overwriting itself in BaseFileManager.uploadfiles as well as BaseFileManager.uploadFiles

#### Core Tests
Tested using manticore.

### Summary
Added overwrite property to SdlFile in Android and JavaSE, added logic to baseFileManager.java to handle the overwrite property when uploading files.

Overwrite property is always set to true, in next major release to align with IOS we need to change it to false

To test I recommend after uploading one file to set the SdlFIle's overwrite property to false and upload it the second time in the completion listener's onComplete of the first upload to verify that the file has uploaded before it tries to upload it again. Ex:

```
		SdlFile livio = new SdlFile("livio", FileType.GRAPHIC_PNG, R.drawable.sdl, false);

		sdlManager.getFileManager().uploadFile(livio, new CompletionListener() {
			@Override
			public void onComplete(boolean success) {
				SdlFile livio2 = new SdlFile("livio", FileType.GRAPHIC_PNG, R.drawable.sdl, false);
				livio2.setOverwrite(false);
				sdlManager.getFileManager().uploadFile(livio2, new CompletionListener() {
					@Override
					public void onComplete(boolean success) {
					}
				});
			}
		});
```

##### Enhancements
* Added overwrite property to SdlFile

### CLA

### DIFF
--- a/base/src/main/java/com/smartdevicelink/managers/file/BaseFileManager.java
+++ b/base/src/main/java/com/smartdevicelink/managers/file/BaseFileManager.java
@@ -402,13 +402,23 @@ abstract class BaseFileManager extends BaseSubManager {
 		}
 		final List<PutFile> putFileRequests = new ArrayList<>();
 		for (SdlFile file : files) {
-			if (!file.isStaticIcon()) {
-				putFileRequests.add(createPutFile(file));
+			if (file.isStaticIcon()) {
+				Log.w(TAG, "Static icons don't need to be uploaded");
+				continue;
 			}
+			// In sdl_ios: HAX: [#827](https://github.com/smartdevicelink/sdl_ios/issues/827) Older versions of Core had a bug where list files would cache incorrectly.
+			// This led to attempted uploads failing due to the system thinking they were already there when they were not.
+			if (!file.isPersistent() && !hasUploadedFile(file)) {
+				file.setOverwrite(true);
+			}
+			if (!file.isOverwrite() && hasUploadedFile(file)) {
+				Log.w(TAG, "Files that have already uploaded and have an Overwrite property set to false will now upload re-Upload");
+				continue;
+			}
+			putFileRequests.add(createPutFile(file));
 		}
 		// if all files are static icons we complete listener with no errors
 		if (putFileRequests.isEmpty()) {
-			Log.w(TAG, "Static icons don't need to be uploaded");
 			listener.onComplete(null);
 		} else {
 			final Map<String, String> errors = new HashMap<>();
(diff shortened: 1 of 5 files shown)


base/src/main/java/com/smartdevicelink/managers/file/BaseFileManager.java
```diff
@@ -329,6 +329,16 @@ public void uploadFile(@NonNull final SdlFile file, final CompletionListener lis
 			listener.onComplete(true);
 			return;
 		}
+		// In sdl_ios: HAX: [#827](https://github.com/smartdevicelink/sdl_ios/issues/827) Older versions of Core had a bug where list files would cache incorrectly.
+		// This led to attempted uploads failing due to the system thinking they were already there when they were not.
+		if (!file.isPersistent() && !hasUploadedFile(file)) {
+			file.setOverwrite(true);
+		}
+		if (!file.isOverwrite() && hasUploadedFile(file)) {
+			Log.w(TAG, "Files that have already uploaded and have an Overwrite property set to false will not be re-Upload");
```
bilal-alsharifi:
```suggestion
			og.w(TAG, String.format("%s has already been uploaded and the overwrite property is set to false. It will not be uploaded again", file.getName()));
```

bilal-alsharifi: Can we also update the static icon warning message to show the file name?

base/src/main/java/com/smartdevicelink/managers/file/BaseFileManager.java
```diff
@@ -329,6 +329,16 @@ public void uploadFile(@NonNull final SdlFile file, final CompletionListener lis
 			listener.onComplete(true);
 			return;
 		}
+		// In sdl_ios: HAX: [#827](https://github.com/smartdevicelink/sdl_ios/issues/827) Older versions of Core had a bug where list files would cache incorrectly.
+		// This led to attempted uploads failing due to the system thinking they were already there when they were not.
+		if (!file.isPersistent() && !hasUploadedFile(file)) {
```
bilal-alsharifi: In java suite, this not necessary since we are covering that case in `hasUploadedFile()`. So that if statement can be removed. This also applies to `uploadFiles()`

PR #1315, opened 2020-03-20
Inconsistencies in ScreenManager behavior if displayCapabilities is null 

### LINKED ISSUE #1310: Inconsistently in ScreenManager behavior if displayCapabilities is null
### Bug Report
Currently, if RAIR.displayCapabilities == null, the SCM will create WindowCapablity that is not null but all fields inside it will be null. Previously, we used to assume that everything is enabled when displayCapabilities is null. However, after the change and because WindowCapablity is always not null, the case is not true in all places in ScreenManager.

```
int getNumberOfLines() {
    if (defaultMainWindowCapability == null){
        return 4;
    }
    int linesFound = 0;
    List<TextField> textFields = defaultMainWindowCapability.getTextFields();
    TextFieldName name;
    if (textFields != null && !textFields.isEmpty()) {
        for (TextField field : textFields) {
            if (field.getName() != null) {
                name = field.getName();
                if (name == TextFieldName.mainField1 || name == TextFieldName.mainField2 || name == TextFieldName.mainField3 || name == TextFieldName.mainField4) {
                    linesFound += 1;
                }
            }
        }
    }
    return linesFound;
}

```

Like in prev example, it's textFields are null, the function will return 0 even though it should return 4.

We have to go over all cases and clearly define what null means for capabilities. And pay special attention to NPEs.

JulianKast:
Fixes #1310 

### Risk

### Testing Plan

#### Unit Tests
[Describe the unit tests and behaviors added in this PR]

#### Core Tests
Tested using manticore 

### Summary
Fixed issue in TextAndGraphicsManager relating to WindowCapability, 
WindowCapability is never null, when we switched from DisplayCapability to using WindowCapability, 
we were checking for if DisplayCapability is null. Now we need to check if a certain capability within WindowCapability is null. 
There were a couple of spots in TextAndGraphiicsManager that weren't changed.

### CLA


base/src/main/java/com/smartdevicelink/managers/screen/BaseTextAndGraphicManager.java
```diff
@@ -742,7 +742,7 @@ private boolean sdlArtworkNeedsUpload(SdlArtwork artwork){
 	}
 
 	private boolean shouldUpdatePrimaryImage() {
-		if (defaultMainWindowCapability == null || defaultMainWindowCapability.getImageTypeSupported() == null || defaultMainWindowCapability.getImageTypeSupported().size() > 0) {
+		if (defaultMainWindowCapability.getImageFields() == null || defaultMainWindowCapability.getImageTypeSupported() == null || defaultMainWindowCapability.getImageTypeSupported().size() > 0) {
```
bilal-alsharifi: `defaultMainWindowCapability` may not be set and could be null if we get `onError` on line 200 or if we don't get a response at all. That applies to all other places where `defaultMainWindowCapability` is used.

theresalech:
Contents of PR are included in #1328

PR #1319, opened 2020-03-24
Cache Lock Screen Icons Retrieved from URL

### LINKED ISSUE #1316: Lock Screen Icon is not cached
### Bug Report
Currently we download the lock screen icon from the URL given to us by the `OnSystemRequest` every single time. We should cache the icon at the URL for 30 days and reuse that image for the given make / model.

##### Reproduction Steps
1. Connect to a head unit that provides a lock screen icon URL.
2. Note the download of the icon.
3. Reconnect to the head unit.

##### Expected Behavior
We should not need to re-download the image.

##### Observed Behavior
We re-download the image.

* Android Version: n/a
* SDL Android Version: n/a
* Testing Against: n/a

n/a

RHenigan:
Fixes #1316 

### Risk

### Testing Plan
 - I have tested this code against core and varified the behavior

#### Unit Tests
 - Added unit tests to ensure caching of lock screen icon is working as expected

### Summary
 - I have added a LockScreenDeviceIconManager class that will be used by the LockScreenManager to cache lock screen icon and to decide when a cached image needs to be updated. When an icon is cached details including the cached path and the last updated time will be writen to shared preferences. shared preferences will be used to verify if an icon is already cached, when an icon needs to be updated, and the path to the cached icon. The icon Url for the lock screen icon will be hashed and used as the key in the shared preferences.

For a list of edge cases please look below to the comments 

### CLA

### DIFF
--- a/android/hello_sdl_android/src/main/AndroidManifest.xml
+++ b/android/hello_sdl_android/src/main/AndroidManifest.xml
@@ -17,7 +17,6 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
-        android:usesCleartextTraffic="true"
         tools:ignore="DeepLinks">
         <activity
             android:name=".MainActivity"
--- a/android/hello_sdl_android/src/main/AndroidManifest.xml
+++ b/android/hello_sdl_android/src/main/AndroidManifest.xml
@@ -17,6 +17,7 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true"
         tools:ignore="DeepLinks">
         <activity
             android:name=".MainActivity"
--- a/android/sdl_android/src/androidTest/AndroidManifest.xml
+++ b/android/sdl_android/src/androidTest/AndroidManifest.xml
@@ -13,8 +13,7 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 
     <application
-        android:debuggable="true"
-        android:usesCleartextTraffic="true">
+        android:debuggable="true">
         <uses-library android:name="android.test.runner" />
     </application>
 
--- a/android/sdl_android/src/androidTest/AndroidManifest.xml
+++ b/android/sdl_android/src/androidTest/AndroidManifest.xml
@@ -12,8 +12,7 @@
     android:targetPackage="com.smartdevicelink.test" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 
-    <application
-        android:debuggable="true">
+    <application android:debuggable="true">
         <uses-library android:name="android.test.runner" />
     </application>
 
--- a/android/sdl_android/src/androidTest/AndroidManifest.xml
+++ b/android/sdl_android/src/androidTest/AndroidManifest.xml
@@ -6,12 +6,12 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <!-- Required to check if WiFi is enabled -->
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-
+    
     <instrumentation
     android:name="android.test.InstrumentationTestRunner"
     android:targetPackage="com.smartdevicelink.test" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
-
+    
     <application android:debuggable="true">
         <uses-library android:name="android.test.runner" />
     </application>
--- a/android/sdl_android/src/androidTest/AndroidManifest.xml
+++ b/android/sdl_android/src/androidTest/AndroidManifest.xml
@@ -6,13 +6,15 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <!-- Required to check if WiFi is enabled -->
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    
+
     <instrumentation
     android:name="android.test.InstrumentationTestRunner"
     android:targetPackage="com.smartdevicelink.test" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
-    
-    <application android:debuggable="true">
+
+    <application
+        android:debuggable="true"
+        android:usesCleartextTraffic="true">
         <uses-library android:name="android.test.runner" />
     </application>
 
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManagerTests.java
@@ -91,7 +91,7 @@ public class LockScreenDeviceIconManagerTests extends AndroidTestCase2 {
         Mockito.when(context.getSharedPreferences(anyString(), anyInt())).thenReturn(sharedPrefs);
         Mockito.when(sharedPrefs.getString(anyString(), (String) isNull())).thenReturn(daysToMillisecondsAsString(15));
         Mockito.when(sharedPrefs.edit()).thenReturn(sharedPrefsEditor);
-        Mockito.when(sharedPrefsEditor.remove(anyString())).thenReturn(sharedPrefsEditor);
+        Mockito.when(sharedPrefsEditor.clear()).thenReturn(sharedPrefsEditor);
 
         lockScreenDeviceIconManager = new LockScreenDeviceIconManager(context);
         lockScreenDeviceIconManager.retrieveIcon(ICON_URL, listener);
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManagerTests.java
@@ -18,6 +18,7 @@ import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.isNull;
@@ -28,7 +29,7 @@ public class LockScreenDeviceIconManagerTests extends AndroidTestCase2 {
 
     TemporaryFolder tempFolder = new TemporaryFolder();
     private LockScreenDeviceIconManager lockScreenDeviceIconManager;
-    private static final String ICON_URL = "http://i.imgur.com/TgkvOIZ.png";
+    private static final String ICON_URL = "https://i.imgur.com/TgkvOIZ.png";
     private static final String LAST_UPDATED_TIME = "lastUpdatedTime";
     private static final String STORED_PATH = "storedPath";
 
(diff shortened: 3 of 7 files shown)


android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,107 @@
...
+import org.json.JSONObject;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.Date;
+
+public class LockScreenDeviceIconManager {
```
bilal-alsharifi: can the class be package-private as it is not supposed to be used by the devs?

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,107 @@
...
+                e.printStackTrace();
+                return null;
+            }
+        } else {
+            return null;
+        }
+    }
+
+    private void writeDeviceIconParametersToSystemPreferences(String iconUrl, JSONObject jsonObject) throws JSONException {
+        SharedPreferences sharedPref = this.context.getSharedPreferences("sdl", Context.MODE_PRIVATE);
```
bilal-alsharifi: I would use a more descriptive name and create a constant for that string instead of hard-coding it in multiple places. That applies to other hardcoded strings in the code. Something like:
https://github.com/smartdevicelink/sdl_java_suite/blob/797b7a14a7fd67fdf6faa504865e4e22228858ef/android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java#L186

RHenigan:
Screen Shot 2020-03-25 at 10 07 04 AM

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -376,16 +379,34 @@ private void downloadDeviceIcon(final String url){
...
 			public void run(){
 				try{
-					deviceLogo = AndroidTools.downloadImage(url);
+					if(mLockScreenDeviceIconManager.updateCachedImage(url)) {
+						Log.d(TAG, "URL: " + url);
+						Log.d(TAG, "Image Update Needed");
+						deviceLogo = AndroidTools.downloadImage(url);
+						mLockScreenDeviceIconManager.saveFileToCache(deviceLogo, url);
+					} else {
+						Log.d(TAG, "Image Is Up To Date");
```
bilal-alsharifi: it is better to use DebugTool for logging D/I messages

bilal-alsharifi: This doesn't always mean the image is up to date right? I believe the app will hit this else statement if the image is not cached at all. 
I suggest updating the log to be more clear or restructuring the if statements to separate the three cases to make it easier to follow:
1-image not cached
2-image cached and valid 
3-image cached but expired

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManagerTests.java
```diff
@@ -0,0 +1,242 @@
...
+            long timeDifferenceInMilliSeconds = DaysOld * 1000 * 60 * 60 * 24;
+            jsonObject.put(LAST_UPDATED_TIME, System.currentTimeMillis() - timeDifferenceInMilliSeconds);
+            return jsonObject.toString();
+        } catch (JSONException e) {
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+    private String getMD5HashFromIconUrl(String iconUrl) {
```
bilal-alsharifi: Would it be possible to reuse the method in `LockScreenDeviceIconManager()` by making it package-private?

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private void writeDeviceIconParametersToSystemPreferences(String iconHash, JSONObject jsonObject) {
+        SharedPreferences sharedPref = this.context.getSharedPreferences(SDL_DEVICE_STATUS_SHARED_PREFS, Context.MODE_PRIVATE);
+        SharedPreferences.Editor editor = sharedPref.edit();
+        editor.putString(iconHash, jsonObject.toString());
+        editor.commit();
+    }
+
+    private JSONObject buildDeviceIconParameters(String storedUrl) throws JSONException {
```
bilal-alsharifi: Do you think we can simplify that a bit by only storing the hash as key and the time as a value instead of saving the JSON in shared prefs?

currently, this is how the share prefs xml look like. We save the hash twice. I feel that is unesceasry 
```
<?xml version='1.0' encoding='utf-8' standalone='yes' ?>
<map>
    <string name="935e06761f887b20a5a10242faaf6ecf">{&quot;storedUrl&quot;:&quot;\/data\/user\/0\/io.livio.sdltestsuite.app1\/cache\/sdl\/lock_screen_icon\/935e06761f887b20a5a10242faaf6ecf&quot;,&quot;lastUpdatedTime&quot;:1585600940171}</string>
</map>
```

Cannot we just store them like that? This way we don't even need to make s JSON.
```
<?xml version='1.0' encoding='utf-8' standalone='yes' ?>
<map>
    <string name="935e06761f887b20a5a10242faaf6ecf">1585600940171</string>
</map>
```

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,164 @@
...
+        JSONObject parametersJson = new JSONObject();
+        parametersJson.put(STORED_URL, storedUrl);
```
bilal-alsharifi: If I understand it correctly, this saves the file path, not the url right? If so, can we change the name to `filePath`  to make it easier to read?

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    void saveFileToCache(Bitmap icon, String iconUrl) {
+
+        String iconHash = getMD5HashFromIconUrl(iconUrl);
+
+        File f = new File(this.context.getCacheDir() + "/" + STORED_ICON_PATH, iconHash);
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        icon.compress(Bitmap.CompressFormat.PNG, 0 /*ignored for PNG*/, bos);
+        byte[] bitmapdata = bos.toByteArray();
```
bilal-alsharifi:
```suggestion
        byte[] bitmapData = bos.toByteArray();
```

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,164 @@
...
+                e.printStackTrace();
+                DebugTool.logError("Exception Trying to read shared preferences");
+                return true;
+            }
+        }
+    }
+
+    void saveFileToCache(Bitmap icon, String iconUrl) {
+
+        String iconHash = getMD5HashFromIconUrl(iconUrl);
```
bilal-alsharifi: Can we remove the extra lines before and after this line?

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final String TAG = "LockScreenManager";
+
+
+    LockScreenDeviceIconManager(Context context) {
+        this.context = context;
+        File lockScreenDirectory = new File(context.getCacheDir(), STORED_ICON_PATH);
+        lockScreenDirectory.mkdirs();
+    }
+
+    boolean updateCachedImage(String iconUrl) {
```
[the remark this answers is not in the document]
RHenigan: shouldUpdateCachedImage sounds good to me

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -376,16 +379,34 @@ private void downloadDeviceIcon(final String url){
 			@Override
 			public void run(){
 				try{
-					deviceLogo = AndroidTools.downloadImage(url);
+					if(mLockScreenDeviceIconManager.updateCachedImage(url)) {
+						Log.d(TAG, "URL: " + url);
```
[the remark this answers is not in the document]
RHenigan: changed to debugTool and removed unecessary URL log

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
RHenigan: changed to. DebugTool

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManagerTests.java
RHenigan: The way it is implemented I need to define the mock (using the md5 hash method) before the class in instantiated, if the class is instantiated before the mock is defined it breaks the test.

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -376,16 +379,34 @@ private void downloadDeviceIcon(final String url){
...
+					} else {
+						Log.d(TAG, "Image Is Up To Date");
+						deviceLogo = mLockScreenDeviceIconManager.getFileFromCache(url);
+						if (deviceLogo == null) {
+							deviceLogo = AndroidTools.downloadImage(url);
+							mLockScreenDeviceIconManager.saveFileToCache(deviceLogo, url);
+						}
+					}
+				} catch(IOException e){
+					Log.e(TAG, "device Icon Error Downloading");
```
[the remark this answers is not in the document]
RHenigan: made into one log statement

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
RHenigan: Refactored

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
RHenigan: the method should only return false if the image is cached and valid, all other instances return true meaning we need to cache a new icon

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
RHenigan: Implemented suggestion and fixed associated code

android/hello_sdl_android/src/main/AndroidManifest.xml
```diff
@@ -17,6 +17,7 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true"
```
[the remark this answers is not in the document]
bilal-alsharifi: We can do that in Java Suite. But what happens if we change `http` to `https` for URL and the server was not setup to support `https`? will it be acceptable if the request fails in that case?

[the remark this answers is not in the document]
joeljfischer: That's my point, it will on iOS guaranteed, with no workarounds. So do we want things to work on Android that don't on iOS? I think we should align the platforms here to the most restrictive.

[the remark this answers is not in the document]
bilal-alsharifi: Agreed

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private void writeDeviceIconParametersToSharedPreferences(String iconHash) {
+        SharedPreferences sharedPref = this.context.getSharedPreferences(SDL_DEVICE_STATUS_SHARED_PREFS, Context.MODE_PRIVATE);
+        SharedPreferences.Editor editor = sharedPref.edit();
+        editor.putString(iconHash, String.valueOf(System.currentTimeMillis()));
+        editor.commit();
+    }
+
+    String getMD5HashFromIconUrl(String iconUrl) {
```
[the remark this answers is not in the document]
RHenigan: This was changed to be package private to access from the tests to avoid duplicating this method in the test

android/hello_sdl_android/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
RHenigan: removed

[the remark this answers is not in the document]
bilal-alsharifi: please make sure that it is removed from the test `AndroidManifest.xml ` as well

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -376,16 +380,31 @@ private void downloadDeviceIcon(final String url){
...
+						}
+					} else {
+						DebugTool.logInfo("Lock Screen Icon Update Needed");
+						deviceLogo = AndroidTools.downloadImage(url);
+						mLockScreenDeviceIconManager.saveFileToCache(deviceLogo, url);
+					}
+				} catch(IOException e){
+					Log.e(TAG, "device Icon Error Downloading, Will attempt to grab cached Icon even if expired: \n" + e.toString());
+					deviceLogo = mLockScreenDeviceIconManager.getFileFromCache(url);
+				}
```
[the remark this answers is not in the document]
RHenigan: Logic pulled into LockScreenDeviceIconManager

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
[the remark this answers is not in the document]
RHenigan: No longer needed in tests, method made private

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,130 @@
...
+        } catch (Exception e) {
+            DebugTool.logError("Failed to save icon to cache");
+            e.printStackTrace();
+            return;
+        }
+
+        writeDeviceIconParametersToSharedPreferences(iconHash);
+    }
+
+    Bitmap getFileFromCache(String iconUrl) {
```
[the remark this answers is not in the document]
RHenigan: made private

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final String SDL_DEVICE_STATUS_SHARED_PREFS = "sdl.lockScreenIcon";
+    private static final String STORED_ICON_PATH = "sdl/lock_screen_icon/";
+
+    LockScreenDeviceIconManager(Context context) {
+        this.context = context;
+        File lockScreenDirectory = new File(context.getCacheDir(), STORED_ICON_PATH);
+        lockScreenDirectory.mkdirs();
+    }
+
+    boolean isIconCachedAndValid(String iconUrl) {
```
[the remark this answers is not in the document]
RHenigan: OnIconRetrievedListener Implemented

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -18,15 +20,45 @@
...
+                iconRetrievedListener.onImageRetrieved(icon);
+            } else {
+                DebugTool.logInfo("Lock Screen Icon Update Needed");
+                icon = AndroidTools.downloadImage(iconURL);
+                saveFileToCache(icon, iconURL);
+                iconRetrievedListener.onImageRetrieved(icon);
+            }
+        } catch (IOException e) {
+            iconRetrievedListener.onError("device Icon Error Downloading, Will attempt to grab cached Icon even if expired: \n" + e.toString());
+            icon = getFileFromCache(iconURL);
```
joeljfischer, line 56: If the icon is `null` here, it should return an error, right?

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -73,13 +105,13 @@ void saveFileToCache(Bitmap icon, String iconUrl) {
...
         String iconHash = getMD5HashFromIconUrl(iconUrl);
         SharedPreferences sharedPref = this.context.getSharedPreferences(SDL_DEVICE_STATUS_SHARED_PREFS, Context.MODE_PRIVATE);
         String iconLastUpdatedTime = sharedPref.getString(iconHash, null);
 
         if (iconLastUpdatedTime != null) {
-            Bitmap cachedIcon = BitmapFactory.decodeFile(this.context.getCacheDir() + "/" + STORED_ICON_PATH + "/" + iconHash);
+            Bitmap cachedIcon = BitmapFactory.decodeFile(this.context.getCacheDir() + "/" + STORED_ICON_DIRECTORY_PATH + "/" + iconHash);
             if(cachedIcon == null) {
                 DebugTool.logError("Failed to get Bitmap from decoding file cache");
                 clearIconDirectory();
```
joeljfischer, line 117: I don't think that this is necessarily what we want to do here, unlike iOS. On iOS, all the image file information is contained in one file. So on iOS, when you clear the icon directory, the file went along with it. Here, if the directory is cleared, all the icon information remains in `SharedPreferences`. That means that attempting to retrieve an icon may fail and then we'll have a cascading set of retrieval failures because old data is contained in `SharedPreferences` that we know isn't accurate.

I can think of a few possibilities to solve this:
1. We don't clear the icon directory if an error occurs. This isn't aligned with iOS' mitigation strategy and has the possibility of leaving icons in the directory that are unused.
2. Instead of storing the data in separate SharedPreferences keys, use a single key, store the information in JSON and store the JSON in the SharedPreferences under that single key. Then when you clear the icon directory, you clear the JSON document as well.

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -379,23 +379,17 @@ private void downloadDeviceIcon(final String url){
...
-						deviceLogo = mLockScreenDeviceIconManager.getFileFromCache(url);
-						if (deviceLogo == null) {
-							deviceLogo = AndroidTools.downloadImage(url);
-							mLockScreenDeviceIconManager.saveFileToCache(deviceLogo, url);
-						}
-					} else {
-						DebugTool.logInfo("Lock Screen Icon Update Needed");
-						deviceLogo = AndroidTools.downloadImage(url);
-						mLockScreenDeviceIconManager.saveFileToCache(deviceLogo, url);
+				mLockScreenDeviceIconManager.retrieveIcon(url, new LockScreenDeviceIconManager.OnIconRetrievedListener() {
```
joeljfischer, line 382: The indentation seems off on these

RHenigan: this is the removed code, the new code seems aligned correctly with the run method above to me.

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
RHenigan: I see what you mean, we would also have the option of sharedPref.edit().clear().commit();

https://developer.android.com/reference/android/content/SharedPreferences.Editor#clear()

would this be closer to the ios implementation?

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -73,13 +105,13 @@ void saveFileToCache(Bitmap icon, String iconUrl) {
         writeDeviceIconParametersToSharedPreferences(iconHash);
```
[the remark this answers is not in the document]
RHenigan: moved call to be within the try block, if write fails details will no longer be saved to shared preferences

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -18,15 +20,45 @@
...
+                DebugTool.logInfo("Icon Is Up To Date");
+                icon = getFileFromCache(iconURL);
+                if (icon == null) {
+                    DebugTool.logInfo("Icon from cache was null, attempting to re-download");
+                    icon = AndroidTools.downloadImage(iconURL);
+                    saveFileToCache(icon, iconURL);
+                }
+                iconRetrievedListener.onImageRetrieved(icon);
+            } else {
+                DebugTool.logInfo("Lock Screen Icon Update Needed");
```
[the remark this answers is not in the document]
RHenigan: added

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
RHenigan: added check, will call onError if downloaded icon is null

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
joeljfischer: I just pulled your latest updates, it still appears incorrect on the new code. They appear to have an extra tab of indentation.

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
joeljfischer: If that clears all data about cached icons and their expiration dates, then yes, I believe it would.

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,205 @@
...
+                if (icon != null) {
+                    saveFileToCache(icon, iconURL);
+                    iconRetrievedListener.onImageRetrieved(icon);
+                } else {
+                    iconRetrievedListener.onError("Icon downloaded was null");
+                }
+            }
+        } catch (IOException e) {
+            iconRetrievedListener.onError("device Icon Error Downloading, Will attempt to grab cached Icon even if expired: \n" + e.toString());
+            icon = getFileFromCache(iconURL);
```
joeljfischer, line 73: If the `getFileFromCache` returns a null icon, it shouldn't return it here. It should return an error through the `onImageRetrieved`, right?

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -18,15 +20,45 @@
...
 
     LockScreenDeviceIconManager(Context context) {
         this.context = context;
-        File lockScreenDirectory = new File(context.getCacheDir(), STORED_ICON_PATH);
+        File lockScreenDirectory = new File(context.getCacheDir(), STORED_ICON_DIRECTORY_PATH);
         lockScreenDirectory.mkdirs();
     }
 
-    boolean isIconCachedAndValid(String iconUrl) {
+    void retrieveIcon(String iconURL, OnIconRetrievedListener iconRetrievedListener) {
```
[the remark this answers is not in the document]
RHenigan: Documentation added, If you have any feedback or feel I am missing any details please let me know

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
RHenigan: null check added

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,209 @@
...
+     * @param iconURL url that the lock screen icon is downloaded from
+     * @param iconRetrievedListener an interface that will implement onIconReceived and OnError methods
+     */
+    void retrieveIcon(String iconURL, OnIconRetrievedListener iconRetrievedListener) {
+        Bitmap icon = null;
+        try {
+            if (isIconCachedAndValid(iconURL)) {
+                DebugTool.logInfo("Icon Is Up To Date");
+                icon = getFileFromCache(iconURL);
+                if (icon == null) {
```
bilal-alsharifi, line 54: what happens if the code hits the if statement on line 54 but failed to download the icon? I think it should call `onError()` in that case

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -375,17 +380,26 @@ private void downloadDeviceIcon(final String url){
 		new Thread(new Runnable(){
 			@Override
 			public void run(){
-				try{
-					deviceLogo = AndroidTools.downloadImage(url);
+				mLockScreenDeviceIconManager.retrieveIcon(url, new LockScreenDeviceIconManager.OnIconRetrievedListener() {
+					@Override
+					public void onImageRetrieved(Bitmap icon) {
```
bilal-alsharifi, line 385: I don't think that will work. We are retrieving the icon in an async way and immediately use the icon without waiting for onImageRetrieved () to be called. I think deviceLogo will always be null. I suggest moving the code after the async call to be inside the `onImageRetrieved()` callback.

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
```diff
@@ -0,0 +1,209 @@
...
+            return false;
+        } else {
+            DebugTool.logInfo("Icon Details Found");
+            long lastUpdatedTime = 0;
+            try {
+                lastUpdatedTime = Long.parseLong(iconLastUpdatedTime);
+            } catch (NumberFormatException e) {
+                DebugTool.logInfo("Invalid time stamp stored to shared preferences, clearing cache and share preferences");
+                clearIconDirectory();
+                sharedPref.edit().clear().commit();
```
bilal-alsharifi, line 102: This clears only the icon shared preferences not all shared prefs for the app right?

RHenigan: Yes, we create a new sharedPreference witht the key of sdl.lockScreenIcon, this is the only shared preference we are referencing withing this class

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
RHenigan: moved

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
RHenigan: If the download fails it throws an IOException which is caught below

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -82,11 +83,14 @@
 	private boolean mLockScreenHasBeenDismissed, lockscreenDismissReceiverRegistered, receivedFirstDDNotification;
 	private String mLockscreenWarningMsg;
 	private BroadcastReceiver mLockscreenDismissedReceiver;
+	private LockScreenDeviceIconManager mLockScreenDeviceIconManager;
```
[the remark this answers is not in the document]
RHenigan: removed

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenManager.java
```diff
@@ -232,6 +236,7 @@ public void onNotified(RPCNotification notification) {
 							msg.getUrl() != null) {
 						// send intent to activity to download icon from core
 						deviceIconUrl = msg.getUrl();
```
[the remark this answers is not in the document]
RHenigan: yes, made the change

android/sdl_android/src/androidTest/AndroidManifest.xml
```diff
@@ -6,13 +6,14 @@
     <uses-permission android:name="android.permission.INTERNET" />
```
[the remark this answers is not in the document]
bilal-alsharifi: for some reason, this file still appears as one of the changed files in this PR. something changed in its formatting.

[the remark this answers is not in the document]
RHenigan: I did a git checkout from develop now

android/sdl_android/src/main/java/com/smartdevicelink/managers/lockscreen/LockScreenDeviceIconManager.java
bilal-alsharifi: Hmm, that is right. In that case, the else statement on line 67 will never be hit right?

RHenigan: line 67 would occur if dowonloadImage returns null rather than throw an exception, so I would need to look and see if downloadImage can return null

RHenigan: downloadImage is returning the following
    Bitmap result = BitmapFactory.decodeStream(bis);
    bis.close(); 
    return result;
.decodeStream can return null so it is possible to get null from downloadImage without an excption being thrown

bilal-alsharifi: in that case, we should always check if the value is null after trying to download the image.

RHenigan: Yeah, I will add a null check

PR #1321, opened 2020-03-25
Align SystemCapabilityManager with iOS (part3)

### LINKED ISSUE #1324: Align SystemCapabilityManager with iOS
### Bug Report
The `SystemCapabilityManager` in Java Suite should be updated to do the following:
- The manager shouldn't send `GetSystemCapability` if the app is already subscribed to the capability (since all the updated values will be received via notifications).
- The manager shouldn't send `GetSystemCapability` requests  if hmilevel is `NONE`

`SystemCapabilityManager` is iOS already does that so Java Suite one should be updated to align.

* SDL Android Version: 4.10.1

bilal-alsharifi:
Fixes #1324 

### Risk

### Testing Plan

#### Unit Tests
Added new unit tests to test getting capabilities on different HMI levels and confirm that `GetSystemCapability` request doesn't get sent out on HMILevel `NONE`

#### Core Tests
Did the following test on both Sync 3.2 and Core 6.0:
- Test `getCapablity()` for a not cached capability on HMI `NONE` and confirm `onError()` is called
- Test `getCapablity()` for a not cached capability on HMI `FULL` and confirm `GetSystemCapbalbity` request is sent to Core and the capability is received correctly 
- Test using `ScreenManager` and make sure it gets the `DISPLAYS` capability correctly and works well with no errors in `logcat`
- Test streaming video and make sure it gets the `VideoStraming` capability correctly and works well with no errors in `logcat`

### Summary
This PR updates `SystemCapabilityManager` to align with iOS:
- Avoid sending  `GetSystemCapability` if the app is already subscribed to the capability (since all the updated values will be received via notifications).
- Avoid sending `GetSystemCapability` requests if hmilevel is `NONE`

### CLA


bilal-alsharifi:
@joeljfischer The `setupRpcListeners()` method was formatted in a strange way. It uses spaces in some parts and tabs in others. I reformatted it to use tabs. PR is ready.

android/sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java
```diff
@@ -866,6 +866,14 @@ public void getCapability(SystemCapabilityType systemCapabilityType, OnSystemCap
 			proxy.getCapability(systemCapabilityType, scListener);
 		}
 
+		@Override
+		public Object getCapability(SystemCapabilityType systemCapabilityType, OnSystemCapabilityListener scListener, boolean forceUpdate) {
+			if (proxy.getSystemCapabilityManager() != null) {
```
joeygrover, line 871:
```suggestion
			if (proxy != null && proxy.getSystemCapabilityManager() != null) {
```

android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -151,6 +151,9 @@ public void onNotified(RPCNotification notification) {
 				}
 				HMILevel prevHMILevel = hmiLevel;
 				hmiLevel = onHMIStatus.getHmiLevel();
+				if (!hmiLevel.equals(HMILevel.HMI_NONE) && VideoStreamManager.this.parameters == null) {
```
joeygrover, line 154:
```suggestion
				if (!HMILevel.HMI_NONE.equals(hmiLevel) && VideoStreamManager.this.parameters == null) {
```

joeygrover: Avoids an NPE if `hmiLevel` is null but doesn't need an NPE check explicitly.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -78,14 +80,17 @@
...
 
 	public SystemCapabilityManager(ISdl callback) {
 		this.callback = callback;
 		this.LISTENER_LOCK = new Object();
 		this.onSystemCapabilityListeners = new HashMap<>();
 		this.cachedSystemCapabilities = new HashMap<>();
 		this.systemCapabilitiesSubscriptionStatus = new HashMap<>();
+		this.systemCapabilitiesSubscriptionStatus.put(SystemCapabilityType.DISPLAYS, true);
 		this.shouldConvertDeprecatedDisplayCapabilities = true;
+		this.currentHMILevel = HMILevel.HMI_NONE;
```
joeygrover, line 93: Is this what iOS does? IMO our `currentHMILevel` would be none until we are at least registered on the IVI and even then we are inferring that. However, I'm not sure if it makes much of a difference to simply put this as the initial state or not so if it is what iOS does then this comment can be resolved without action.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
```diff
@@ -541,6 +555,15 @@ public boolean removeOnSystemCapabilityListener(final SystemCapabilityType syste
 	 * @param subscribe flag to subscribe to updates of the supplied capability type. True means subscribe; false means cancel subscription; null means don't change current subscription status.
 	 */
 	private void retrieveCapability(final SystemCapabilityType systemCapabilityType, final OnSystemCapabilityListener scListener, final Boolean subscribe) {
+		if (currentHMILevel != null && currentHMILevel.equals(HMILevel.HMI_NONE)) {
+			String message = String.format("Attempted to update type: %s in HMI level NONE, which is not allowed. " +
+					"Please wait until you are in HMI BACKGROUND, LIMITED, or FULL before attempting to update any SystemCapabilityType", systemCapabilityType);
+			DebugTool.logError(message);
```
joeygrover, line 561: Do we need the `message` variable here, or couldn't this simply be inline with the `DebugTool` call?

base/src/main/java/com/smartdevicelink/proxy/interfaces/ISdl.java
```diff
@@ -208,6 +208,7 @@
      * @param systemCapabilityType a system capability type that should be retrieved
      * @return the system capability provided if available, null if not
      */
+    @Deprecated
```
joeygrover, line 211: Please add javadoc as to what method should be used instead of this one because of deprecation.

base/src/main/java/com/smartdevicelink/proxy/interfaces/ISdl.java
```diff
@@ -216,8 +217,18 @@
      * @param scListener listener that will be called when the system capability is retrieved. If already cached, it
      *                   will be called immediately
      */
+    @Deprecated
```
joeygrover, line 220: Please add javadoc as to what method should be used instead of this one because of deprecation.

javaSE/src/main/java/com/smartdevicelink/managers/lifecycle/LifecycleManager.java
```diff
@@ -1145,6 +1145,11 @@ public void getCapability(SystemCapabilityType systemCapabilityType, OnSystemCap
 
         }
 
+        @Override
+        public Object getCapability(SystemCapabilityType systemCapabilityType, OnSystemCapabilityListener scListener, boolean forceUpdate) {
+            return LifecycleManager.this.systemCapabilityManager.getCapability(systemCapabilityType, scListener, forceUpdate);
```
joeygrover, line 1150: There should probably be some null checks here and in the `isCapabilitySupported` methods to make sure that `LifecycleManager.this.systemCapabilityManager` isn't null before simply accessing it.

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
bilal-alsharifi: Yes, iOS does a similar thing: https://github.com/smartdevicelink/sdl_ios/blob/4f650efc70d853d31b14ccc3e64210f192f3f9b7/SmartDeviceLink/SDLSystemCapabilityManager.m#L97

base/src/main/java/com/smartdevicelink/proxy/SystemCapabilityManager.java
bilal-alsharifi: I stored the message in var so I can reuse it when I call `scListener.onError(message)`

base/src/main/java/com/smartdevicelink/proxy/interfaces/ISdl.java
bilal-alsharifi: I will add that to deprecated methods here and in SCM

PR #1328, opened 2020-04-02
Inconsistencies in ScreenManager behavior if displayCapabilities is null & revisions

### LINKED ISSUE #1310: Inconsistently in ScreenManager behavior if displayCapabilities is null

bilal-alsharifi:
Fixes #1310 

### Risk

### Testing Plan

#### Unit Tests
[Describe the unit tests and behaviors added in this PR]

#### Core Tests
Tested using manticore 

### Summary
Fixed issue in TextAndGraphicsManager relating to WindowCapability, 
WindowCapability is never null, when we switched from DisplayCapability to using WindowCapability, 
we were checking for if DisplayCapability is null. Now we need to check if a certain capability within WindowCapability is null. 
There were a couple of spots in TextAndGraphiicsManager that weren't changed.

### CLA


base/src/main/java/com/smartdevicelink/managers/ManagerUtility.java
```diff
@@ -0,0 +1,86 @@
...
+
+        /**
+         * Check to see if WindowCapability has an ImageFieldName of a given name.
+         *
+         * @param windowCapability WindowCapability representing the capabilities of the desired window
+         * @param name ImageFieldName representing a name of a given Image field that would be stored in WindowCapability
+         * @return true if name exist in WindowCapability else false
+         */
+        public static boolean hasImageFieldOfName(WindowCapability windowCapability, ImageFieldName name) {
+            if (windowCapability == null) {
```
joeygrover, line 26:
```suggestion
            if (windowCapability == null || name == null) {
```

base/src/main/java/com/smartdevicelink/managers/ManagerUtility.java
```diff
@@ -0,0 +1,86 @@
...
+                return false;
+            }
+            if (windowCapability.getImageFields() != null) {
+                for (ImageField field : windowCapability.getImageFields()) {
+                    if (field != null && field.getName() != null && field.getName().equals(name)) {
```
joeygrover, line 31:
```suggestion
                    if (field != null && name.equals(field.getName())) {
```

joeygrover: If we add the null check to `name` above, we can avoid the null check to `field.getName()`.

base/src/main/java/com/smartdevicelink/managers/ManagerUtility.java
```diff
@@ -0,0 +1,86 @@
...
+
+        /**
+         * Check to see if WindowCapability has a textField of a given name.
+         *
+         * @param windowCapability WindowCapability representing the capabilities of the desired window
+         * @param name TextFieldName representing a name of a given text field that would be stored in WindowCapability
+         * @return true if name exist in WindowCapability else false
+         */
+        public static boolean hasTextFieldOfName(WindowCapability windowCapability, TextFieldName name) {
+            if (windowCapability == null) {
```
joeygrover, line 47:
```suggestion
            if (windowCapability == null || name == null) {
```

base/src/main/java/com/smartdevicelink/managers/ManagerUtility.java
```diff
@@ -0,0 +1,86 @@
...
+                return false;
+            }
+            if (windowCapability.getTextFields() != null) {
+                for (TextField field : windowCapability.getTextFields()) {
+                    if (field != null && field.getName() != null && field.getName().equals(name)) {
```
joeygrover, line 52:
```suggestion
                    if (field != null && name.equals(field.getName())) {
```

base/src/main/java/com/smartdevicelink/managers/ManagerUtility.java
```diff
@@ -0,0 +1,86 @@
...
+                for (TextField field : windowCapability.getTextFields()) {
+                    if (field.getName() != null) {
+                        name = field.getName();
+                        if (name == TextFieldName.mainField1 || name == TextFieldName.mainField2 || name == TextFieldName.mainField3 || name == TextFieldName.mainField4) {
+                            int fieldNumber = Integer.parseInt(name.toString().substring(name.toString().length() - 1));
+                            highestFound = Math.max(highestFound, fieldNumber);
+                            if (highestFound == 4) {
+                                break;
+                            }
+                        }
```
joeygrover, line 79:
```suggestion

                        int fieldNumber =0;
                        switch (textField.getName()){
                            case mainField1:
                                fieldNumber = 1;
                                break;
                            case mainField2:
                                fieldNumber = 2;
                                break;
                            case mainField3:
                                fieldNumber = 3;
                                break;
                            case mainField4:
                                fieldNumber = 4;
                                break;
                        }
                        if(fieldNumber > 0){
                            highestFound = Math.max(highestFound, fieldNumber);
                            if (highestFound == 4) {
                                break;
                            }
                        }
```

joeygrover: While the current solution is more adaptable/clever, the reality is that there are only 4 specific cases that we care about and therefore should avoid string manipulation. I suggest not taking this suggestions from github, but making the changes in Android Studio.

base/src/main/java/com/smartdevicelink/managers/screen/BaseTextAndGraphicManager.java
```diff
@@ -409,7 +410,7 @@ Show assembleShowText(Show show){
 			return show;
 		}
 
-		int numberOfLines = getNumberOfLines();
+		int numberOfLines = (defaultMainWindowCapability != null && defaultMainWindowCapability.getTextFields() != null) ? ManagerUtility.WindowCapabilityUtility.getMaxNumberOfMainFieldLines(defaultMainWindowCapability) : 4;
```
joeygrover, line 413: Does it max sense to just alter the utility method call to return 4 by default instead of this long inline if statement? This is not a hard requirement and would be interested to know how iOS handles it.

base/src/main/java/com/smartdevicelink/managers/screen/BaseTextAndGraphicManager.java
```diff
@@ -741,53 +742,64 @@ private boolean sdlArtworkNeedsUpload(SdlArtwork artwork){
...
 
-		if (defaultMainWindowCapability == null){
-			return 4;
-		}
+	/**
+	 * Check to see if template supports the specified image field
+	 * @return true if image field is supported, false if not
+	 */
+	private boolean templateSupportsImageField(ImageFieldName name) {
+		return (defaultMainWindowCapability == null || defaultMainWindowCapability.getImageFields() == null) || ManagerUtility.WindowCapabilityUtility.hasImageFieldOfName(defaultMainWindowCapability, name);
```
joeygrover, line 778: I don't believe this logic is correct. Based on the RPC spec, the `imageField` array is not mandatory. Therefore if it is not included, we should infer that it is not supported, right? @joeljfischer can you confirm?

base/src/main/java/com/smartdevicelink/managers/screen/choiceset/PreloadChoicesOperation.java
```diff
@@ -227,14 +230,44 @@ private CreateInteractionChoiceSet choiceFromCell(ChoiceCell cell){
 	}
 
 	// HELPERS
+	boolean shouldSendChoiceText() {
```
joeygrover, line 233: Is there a reason why this methods are `shouldSend____Text` instead of something more like `is___TextSupported`?

base/src/main/java/com/smartdevicelink/managers/screen/menu/BaseMenuManager.java
```diff
@@ -780,17 +780,7 @@ private void setAddStatus(Integer size, List<Integer> newArray){
...
-			if (imageFields != null && !imageFields.isEmpty()) {
-				for (ImageField field : imageFields) {
-					if (field != null && field.getName() != null && field.getName().equals(ImageFieldName.cmdIcon)) {
-						return true;
-					}
-				}
-			}
-		}
-		return false;
+		return (defaultMainWindowCapability == null || defaultMainWindowCapability.getImageFields() == null) || ManagerUtility.WindowCapabilityUtility.hasImageFieldOfName(defaultMainWindowCapability, ImageFieldName.cmdIcon);
```
joeygrover, line 783: Same question before around if the `imageField` array is null. @joeljfischer for visibility.

base/src/main/java/com/smartdevicelink/managers/ManagerUtility.java
bilal-alsharifi: I agree that the suggested code should be better in performance. I will update the method.

base/src/main/java/com/smartdevicelink/managers/screen/BaseTextAndGraphicManager.java
bilal-alsharifi: We kind of abstracted the logic in the helper methods in `ManagerUtility` to just return the actually supported things. And we kept the implementation for the rule "if null, then all enabled" directly in the manager's implementation. Of course, that is not a hard rule. We can apply the suggestion if you think that makes more sense. I just did it this way for alignment purposes as iOS implementation also does it similarly: 
https://github.com/smartdevicelink/sdl_ios/blob/d4bd0eed98cb1e872da5bcd4a62b04d81a87f815/SmartDeviceLink/SDLTextAndGraphicManager.m#L315:

base/src/main/java/com/smartdevicelink/managers/screen/choiceset/PreloadChoicesOperation.java
bilal-alsharifi: No reason other than following the iOS naming to make them easier to compare and maintain in the future. If you prefer changing the names, I suggest removing these 5 methods and call `templateSupports___Field()` directly.

base/src/main/java/com/smartdevicelink/managers/screen/BaseTextAndGraphicManager.java
joeygrover: If that's how iOS does it then staying aligned is more important right now. We can always address a better method for both in the future.

base/src/main/java/com/smartdevicelink/managers/screen/choiceset/PreloadChoicesOperation.java
joeygrover: If this matches iOS then we can keep these for now. I would like to revisit for both after this release if possible. The naming doesn't make sense in my opinion.

PR #1330, opened 2020-04-02
Remove left over code in hellosdl

RHenigan:

### Risk

### Testing Plan
- Verify app still runs against sdl core, menticore, and tdk

### Summary
Removes an unnecessary check due to adding ManagerShouldUpdateLifecycle


android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -41,9 +41,7 @@ public void onReceive(Context context, Intent intent) {
 			String action = intent.getAction();
```
bilal-alsharifi, line 41: The whole if statement from line 40 to 49 can be removed.  Also, any unused vars and imports (as a result of the ode removal) can be removed as well.

bilal-alsharifi:
It seems that GitHub CI is not passing on this PR.

PR #1360, opened 2020-05-20
Fix #1351

### LINKED ISSUE #1351: Samsung devices not properly managing notification channels
### Bug Report
Since the channel delete code as added back in last release to address some phones not clearing the notification, a new issue has appeared with Samsung devices. These devices are not properly deleting and recreating the notification channel and therefore it throws a fatal exception. While the code itself in the library should not allow this to happen, it seems the OS has other issues.

##### Reproduction Steps
1. Star SDL connection like normal
2. Disconnect
3. Restart SDL connection, should crash

##### Expected Behavior

Notification is displayed and clears properly.

##### Observed Behavior
Notification is not created, RS not started, app experiences fatal crash

```
Fatal Exception: android.app.RemoteServiceException
Bad notification for startForeground: java.lang.RuntimeException: invalid channel for service notification: Notification(channel=sdl_notification_channel pri=0 contentView=null vibrate=null sound=null tick defaults=0x0 flags=0x40 color=0x00000000 vis=PRIVATE semFlags=0x0 semPriority=0 semMissedCount=0)
```

* Assuming Android 9+, likely 10
*Devices: Samsung S10+, Note 10+, S10, S9+, S9,

joeygrover:
Fixes #1351 

### Risk

### Testing Plan
- [x] I have tested this PR against a TDK and verified behavior (if applicable, if not applicable, explain why below).
- [x] I have tested Android

#### Tests
This issue is tricky to test since we have not found anyone who can actually reproduce the issue. Therefore we will have to simulate the issue and test around it. There are essentially two tests we can run: one to ensure the exception handler is properly triggered, and two to make sure the flag is properly saved and retrieved. 

##### Testing the exception handler:

Comment out line ~1514 that is `notificationManager.createNotificationChannel(notificationChannel);` and add a log into the exception handler. This will prevent the channel from being actually registered with the system and then the exception handler should trigger. Simply turn on the bluetooth adapter and connect to the module. The service will start up and the exception will be thrown within a few seconds.

##### Testing the flag is set

Change the previous line to 

```					
if(getBooleanPref(KEY_AVOID_NOTIFICATION_CHANNEL_DELETE,false)){
	notificationManager.createNotificationChannel(notificationChannel);
    Log.d(TAG, "Created channel");
}else{
	Log.d(TAG, "Did not create channel");
}
```

On first pass this will throw the exception. The router service will shutdown. Turn off bluetooth or disconnect from the device. Then turn bluetooth back on and reconnect to the device. This time, since the flag was set, the channel will be created. 

### Summary

- Added an exception handler for the runtime exception caused by some poor Android implementations around the notification channel. Samsung devices don't seem to properly handle the case where we delete the notification channel after the service exits the foreground. Even though we create the channel on every foreground entrance, the OS doesn't honor that and the crashes. Therefore the handler was added and a flag will be set to prevent the channel deletion in the next run of the router service.
- Increased the 

##### Bug Fixes
* App will no longer crash. Dynamic behavior is in place to still allow SDL to function.

### CLA


android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1170,6 +1178,9 @@ protected void deployNextRouterService(){
 	}
 	
 	public void startUpSequence(){
+
+		setRouterServiceExceptionHandler();
```
bilal-alsharifi, line 1182: I think we should call this method first thing in `onCreate()` because currently, the service could enter foreground before adding the exception handler

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1107,7 +1107,9 @@ private boolean initCheck(){
 	@Override
 	public void onCreate() {
 		super.onCreate();
-		//This must be done regardless of if this service shuts down or not
+		//Add this first to avoid the runtime exceptions for the entire lifecycle of the service
+        setRouterServiceExceptionHandler();
```
bilal-alsharifi, line 1111: Spacing seems to be strange here.

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1229,7 +1229,11 @@ public void uncaughtException(Thread t, Throwable e) {
 							&& e.getMessage().contains("invalid channel for service notification")) { //This is the message received in the exception for notification channel issues
 
 						// Set the flag to not delete the notification channel to avoid this exception in the future
-						SdlRouterService.this.setSdlRouterServicePrefs(KEY_AVOID_NOTIFICATION_CHANNEL_DELETE, true);
+						try{
+						    SdlRouterService.this.setSdlRouterServicePrefs(KEY_AVOID_NOTIFICATION_CHANNEL_DELETE, true);
+                        }catch (Exception exception){
```
bilal-alsharifi, line 1234: Spacing seems to be strange here as well.

PR #1381, opened 2020-06-12
Bugfix/issue 1377 http request task

RHenigan:
Fixes #[1377]

### Risk

### Testing Plan
- [X] I have tested Android, and Java SE

### Summary
This PR will:
 - Deprecate AsyncTask within base/android/os
 - Deprecate HttpRequestTask within JavaSE
 - Move HttpRequestTask to sdl_android
With these changes we can safely remove HttpRequestTask from JavaSE and AsyncTask from base/android/os in the next major release

### CLA


javaSE/src/main/java/com/smartdevicelink/util/HttpRequestTask.java
```diff
@@ -0,0 +1,23 @@
+package com.smartdevicelink.util;
+
```
bilal-alsharifi, line 2: Why do the methods don't have any implementation? I agree that the class should be deprecated. However, it should be an exact copy of the android one for now. Otherwise, it will be a breaking change because the implementation was exposed to the developers who are using JavaSE but now it is not,

RHenigan: Methods implemented

PR #1386, opened 2020-06-17
Bugfix/issue 1385

### LINKED ISSUE #1385: AndroidTestCase is not available on Android 28 and AndroidJUnit4 should be used instead
### Bug Report
`android.test.AndroidTestCase` is no longer available on Android 28. So tests need to be rewritten to work using annotations:

```
@RunWith(AndroidJUnit4.class)
public class ExampleInstrumentedTest {
    @Test
    public void test1() {
    }
}
```

RHenigan:
Fixes #1385

### Risk

### Testing Plan

#### Summary
AndroidTestCase is not available on Android 28 and AndroidJUnit4 should be used instead.
All classes that extended AndroidTestCase2 have been refactored to run with AndroidJUnit4.
BaseRPCTest is a parent class for many other test classes that also needed to be refactored to use the JUnit framework.
RSVTestCase and SdlConnectionTest, once refactored, ran into failing tests because the order the tests are run in differed from AndroidTestCase to AndroidJUnit4. To resolve this issue I have annotated these two classes with `@FixMethodOrder(MethodSorters.JVM)` to ensure these tests are run in the right order.

### CLA


android/sdl_android/src/main/java/com/smartdevicelink/managers/video/VideoStreamManager.java
```diff
@@ -146,6 +146,7 @@ public void onServiceError(SdlSession session, SessionType type, String reason)
 	private final OnRPCNotificationListener hmiListener = new OnRPCNotificationListener() {
 		@Override
 		public void onNotified(RPCNotification notification) {
+			Log.e(TAG, "TESTING");
```
bilal-alsharifi, line 149: can we remove this?

RHenigan: Yeah, thanks for catching that

PR #1428, opened 2020-07-30
Fix ClassCastException when calling RPC getter method returning Float 

### LINKED ISSUE #1407: ClassCastException when calling RPC getter method returning Float type
### Bug Report
A class cast exception occurs when an RPC getter method such as ```getPressure()``` inside of ```SingleTireStatus``` is called. This is because of the method used for JSON marshalling inside of the Java suite (code linked below). The JSON hashtable serializer method will store a whole number as an Integer and a whole number with a decimal included as a Double. When an RPC class method such as ```getPressure()``` inside of ```SingleTireStatus``` attempts to retrieve data from its corresponding list a class cast exception will occur because the method has a Float signature, but the data type is a Double.
https://github.com/smartdevicelink/sdl_java_suite/blob/master/base/src/main/java/com/smartdevicelink/marshal/JsonRPCMarshaller.java

Pressure is defined as a Float in the RPC spec
https://github.com/smartdevicelink/rpc_spec/blob/master/MOBILE_API.xml#L1341

The Java suite throws a ClassCastException when calling the getPressure() method
https://github.com/smartdevicelink/sdl_java_suite/blob/master/base/src/main/java/com/smartdevicelink/proxy/rpc/SingleTireStatus.java#L142

One possible solution to the this issue is to update getter methods to first call the ```SdlDataTypeConverter``` to ensure the proper type is returned for any method that returns a Float value. A similar approach is already being used for retrieving the speed parameter in getVehicleData (please see links below).  Its possible updating the central getFloat() methods to first call the SdlDataTypeConverter would be the best solution.

Speed is defined as a float in the RPC spec
https://github.com/smartdevicelink/rpc_spec/blob/master/MOBILE_API.xml#L2068

The Java suite first calls the ```SdlDataTypeConverter``` before returning the value for speed in getVehicleData response to ensure type safety
https://github.com/smartdevicelink/sdl_java_suite/blob/93778656113f0023ec7025281b77e28000a1f5bd/base/src/main/java/com/smartdevicelink/proxy/rpc/GetVehicleDataResponse.java#L120

##### Reproduction Steps
1. Register a test app with SDL core and activate it on the HMI
2. Ensure the HMI is configured to send data for an RPC method that returns a Float data type
3. Take note of the class cast exception

##### Expected Behavior
No class cast exception occurs.

##### Observed Behavior
Class cast exception occurs.

* Android Version: Android 10
* SDL Android Version: 4.11.1
* Testing Against: Sync Head Unit

bilal-alsharifi:
Fixes #1407 

### Risk

### Testing Plan

#### Core Tests
Get `tirePressure` using the following code snippet and make sure the value is retrieved correctly without crashing the app
```
GetVehicleData getVehicleData = new GetVehicleData();
getVehicleData.setTirePressure(true);
getVehicleData.setOnRPCResponseListener(new OnRPCResponseListener() {
    @Override
        GetVehicleDataResponse getVehicleDataResponse = (GetVehicleDataResponse) response;
        Log.i(TAG, "tirePressure: " + getVehicleDataResponse.getTirePressure().getLeftFront().getPressure());
    }
});
sdlManager.sendRPC(getVehicleData);
```

Module tested against: Manticore

### Summary
This PR fixes an issue in `RCPStruct` that causes a class cast exception when the app tries to retreive value using `getFloat()` & `getDouble()` methods

### CLA

### DIFF
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/RPCStructTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/test/proxy/RPCStructTests.java
@@ -155,4 +155,32 @@ public class RPCStructTests extends TestCase {
         testStruct.setValue(invalidKey, TestValues.GENERAL_STRING);
         assertNull(testStruct.getObject(Integer.class, invalidKey));
     }
+
+    public void testGetFloat() {
+        Hashtable <String, Object> map = new Hashtable<>();
+        String key = "test";
+        Double value = 42.00;
+        map.put(key, value);
+        RPCStruct rpcStruct = new RPCStruct(map);
+        try {
+            Float value2 = rpcStruct.getFloat(key);
+            assertTrue(value2 - value < 0.1);
+        } catch (ClassCastException e) {
+            fail(e.getMessage());
+        }
+    }
+
+    public void testGetDouble() {
+        Hashtable <String, Object> map = new Hashtable<>();
+        String key = "test";
+        Integer value = 42;
+        map.put(key, value);
+        RPCStruct rpcStruct = new RPCStruct(map);
+        try {
+            Double value2 = rpcStruct.getDouble(key);
+            assertTrue(value2 - value < 0.1);
+        } catch (ClassCastException e) {
+            fail(e.getMessage());
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


joeljfischer:
If possible, please add a test case that used to fail and now passes.

bilal-alsharifi:
That is a good idea. I added tests for that.

Issue #1306, opened 2020-03-11
[SDL 0286] Java Suite Cleanup


theresalech:
Proposal: [Java Suite Cleanup](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0286-java-suite-cleanup.md)

> The Java Suite (formerly Android) repo has progressed without the need for a major version for over four years. However, it is time to remove old code from the project and create a more streamlined library. This proposal will go over the changes that are needed to clean up the Java Suite library.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/956

Steering Committee Decision:

> The Steering Committee fully agreed to accept this proposal. It was noted that implementation effort and timing could be brought before the Steering Committee when planning future releases.

theresalech:
Implemented in Java Suite 5.0.

PR #1494, opened 2020-09-11
Fix RPCs Mismatches

bilal-alsharifi:
Fixes #1306 

### Risk

### Testing Plan

### Summary
This PR fixes mismatches in some RPCs between the code in java suite and RPC spec.

### CLA


base/src/main/java/com/smartdevicelink/proxy/rpc/HMICapabilities.java
```diff
@@ -58,11 +58,17 @@ public boolean isNavigationAvailable(){
 		 }
 		 return (Boolean)available;
 	 }
-	 
+
+	 @Deprecated
 	 public HMICapabilities setNavigationAvilable( Boolean available) {
```
joeygrover, line 63: Please add javadoc as to what method to use now that this one is deprecated. Since this is a spelling fix to a method name please use the new method inside this method to avoid code redundancy.

joeygrover: This should be done for all method name spelling fixes as well. I won't leave comments on each.

joeygrover:
Please make sure to add javadoc to deprecated methods especially if there are new methods that should be used. Also, if you are deprecating a method to replace it, you should call the new method to reduce code redundancy.

RHenigan (changes requested):

JulianKast (approved):

PR #1552, opened 2020-11-06
Add code to handle legacy PTU (v4.12)

joeygrover:
Fixes part of #1541 

This PR is **ready** for review.

### Risk
This PR makes **no** API changes.

### Testing Plan
- [x] I have verified that I have not introduced new warnings in this PR (or explain why below)
- [x] I have run the unit tests with this PR
- [x] I have tested this PR against Core and verified behavior (if applicable, if not applicable, explain why below).
- [x] I have tested Android

### Summary
- Added code to the RPC converter to handle legacy RPCs that no longer have classes associated with them
- Added logic to process the legacy PTU RPCs as newer, still supported RPC classes

##### Bug Fixes
* Fixes issue where policy table updates wouldn't occur on older systems

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #1558, opened 2020-11-13
Add code to handle legacy PTU (5.0)

### LINKED ISSUE #1541: Supporting devices with SDL 2.0.0
### Bug Report
I'm using an old device with SDL 2.0 and I don't have the ability to upgrade to the latest. Looks like the last versions of android SDL doesn't support old SDL cores. 

##### Reproduction Steps
1. Set value for `MediaTrackTextField` it doesn't work with 4.12.x versions but work with 4.9.0 version
2. Getting permissions from the server can't process it with success with 4.12.x versions (it tries several times) but works with 4.9.0 version, permissions are downloaded but not applied correctly.  

* Android Version: 9
* SDL Android Version: 4.12.1
* Testing Against: [What you tested with to observe this behavior]

joeygrover:
Fixes #1541 partly for 5.0

### Risk

### Testing Plan
- [ ] I have tested Android, Java SE, and Java EE

### Summary
Same changes present as the PR #1552 

This PR should be reviewed and tested at the same time as it will require a Sync 2.0 TDK

##### Bug Fixes
* Fixes issue where policy table updates wouldn't occur on older systems

### CLA

### DIFF
--- a/base/src/main/java/com/smartdevicelink/managers/lifecycle/BaseLifecycleManager.java
+++ b/base/src/main/java/com/smartdevicelink/managers/lifecycle/BaseLifecycleManager.java
@@ -446,7 +446,6 @@ abstract class BaseLifecycleManager {
                             }
                         }
 
-                        DebugTool.logWarning(TAG, "Unable to handle OnSystemRequest");
                         break;
                     case ON_APP_INTERFACE_UNREGISTERED:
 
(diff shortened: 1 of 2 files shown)


base/src/main/java/com/smartdevicelink/managers/lifecycle/BaseLifecycleManager.java
```diff
@@ -387,42 +391,62 @@ public void onReceived(RPCMessage message) {
...
-                                }
-                            };
-                            handleOffBoardTransmissionThread.start();
+                                };
+                                handleOffBoardTransmissionThread.start();
+                                return;
+                            }
                         }
+
+                        DebugTool.logWarning(TAG, "Unable to handle OnSystemRequest");
```
bilal-alsharifi, line 449: This logs a warning when we get a system request with type `LOCK_SCREEN_ICON_URL`

bilal-alsharifi (changes requested):

bilal-alsharifi (approved):

PR #1575, opened 2020-12-03
Add map to track which transports started a service

### LINKED ISSUE #1574: OnServiceError for Nav and PCM is called even if no service from that type is started
### Bug Report
I see these error messages in the log after transport disconnects even if NAV & PCM series are never started:

```
E/BaseEncryptionLifecycleManager: RC_5.0.0: onServiceError, session Type: NAV, reason: Transport disconnected
E/BaseEncryptionLifecycleManager: RC_5.0.0: onServiceError, session Type: PCM, reason: Transport disconnected
```

The reason for the issue is that we call onTransportError() for the aforementioned services without knowing if the service is started or not. This should be tracked so we only call onTransportError when a service is started on these transports.

RHenigan:
Fixes #1574 

### Risk

### Testing Plan

#### Core Tests
Security should be set up to see the errors. So it is better to test with test suite.

modify the testSuite so that you can call `sdlManager.dispose()` I added it to `appInfo` menu item

Test 1)
Connect your testSuite to core and open the application on core.
Now that the SDL service is started you can call `sdlManager.dispose()` and in the logs you should no longer observer the following log messages:
```
E/BaseEncryptionLifecycleManager: RC_5.0.0: onServiceError, session Type: NAV, reason: Transport disconnected
E/BaseEncryptionLifecycleManager: RC_5.0.0: onServiceError, session Type: PCM, reason: Transport disconnected
```

Test 2)
Make sure you set the test suite to type Navigation.
Connect your testSuite to core and open the application on core.
Once the app is connected and open start the video streaming test in the test suite.
While the video is streaming you can call `sdlManager.dispose()` the app may crash as this is not the proper way clean up the managers but you can now verify in the logs that just the following log is printed
```
E/BaseEncryptionLifecycleManager: RC_5.0.0: onServiceError, session Type: NAV, reason: Transport disconnected
```

Test 3)
Make sure you set the test suite to type Navigation.
Connect your testSuite to core and open the application on core.
Once the app is connected and open start the audio streaming test in the test suite.
While the audio is streaming you can call `sdlManager.dispose()` the app may crash as this is not the proper way clean up the managers but you can now verify in the logs that just the following log is printed
```
E/BaseEncryptionLifecycleManager: RC_5.0.0: onServiceError, session Type: PCM, reason: Transport disconnected
```

### Summary
This PR will add a hashmap to the `SdlProtocolBase` which will be used to track whether there is an active service or not on a given SessionType. This will be used to make sure we are appropriatly logging errors, currently we are logging errors for SessionTypes even when there is no active service for that SessionType.

### CLA

### DIFF
--- a/base/src/main/java/com/smartdevicelink/protocol/SdlProtocolBase.java
+++ b/base/src/main/java/com/smartdevicelink/protocol/SdlProtocolBase.java
@@ -103,6 +103,7 @@ public class SdlProtocolBase {
     private final Hashtable<Byte, Object> _messageLocks = new Hashtable<>();
     private final HashMap<SessionType, Long> mtus = new HashMap<>();
     private final HashMap<SessionType, TransportRecord> activeTransports = new HashMap<>();
+    private final HashMap<SessionType, Boolean> serviceOnTransport = new HashMap<>();
     private final Map<TransportType, List<ISecondaryTransportListener>> secondaryTransportListeners = new HashMap<>();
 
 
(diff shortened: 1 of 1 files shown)


RHenigan:
[(3) onServiceError() for NAV & PCM is called when transport disconnects, even if no service from that type is started](https://trello.com/c/HZK6NbAl/1506-3-onserviceerror-for-nav-pcm-is-called-when-transport-disconnects-even-if-no-service-from-that-type-is-started)

JulianKast (approved):

PR #1652, opened 2021-04-06
Implement queries instead of blanket QUERY_ALL_PACKAGES permission

joeygrover:

### Risk

### Testing Plan
- [x] I have tested Android

##### HelloSDL Base Test

1. Install HelloSDL Android application
2. Connect with head unit over bluetooth
3. Observe connection is established and HelloSDL is hosting the router service

##### Multi App Test

1. Create an additional Android app with SDL integration and install alongside HelloSDL
2. Connect with head unit over bluetooth
3. Observe connection is established and only one app is hosting the router service. 

If you remove the queries section from the manifest of either app it should be observed that two router services are started instead, one by each app.

### Summary
Android 30 introduced a new permission `QUERY_ALL_PACKAGES` that was necessary to query the Android device for other apps. While adding hte permission into the manifest was a simple update, a recent announcement mentioned that the permission was no longer going to be available for most apps and it was considered a sensitive permission. This change was going to be enforced very soon. This PR removes the dependence on that permission and adds the necessary queries declarations into the HelloSDL app's manifest. 

##### Bug Fixes
* The `QUERY_ALL_PACKAGES` permission is no longer needed by adding the queries declarations.

### CLA

### DIFF
--- a/android/hello_sdl_android/src/main/AndroidManifest.xml
+++ b/android/hello_sdl_android/src/main/AndroidManifest.xml
@@ -12,6 +12,16 @@
     <!-- Required to use the USB Accessory mode -->
     <uses-feature android:name="android.hardware.usb.accessory" />
 
+    <!-- Required when targeting Android API 30+ -->
+    <queries>
+        <intent>
+            <action android:name="com.smartdevicelink.router.service" />
+        </intent>
+        <intent>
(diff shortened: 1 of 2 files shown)


JulianKast (commented):
Approved

PR #1683, opened 2021-05-07
Improve packet state machine

### LINKED ISSUE #1678: java.lang.NegativeArraySizeException Crash at SdlPsm.java line 241 com.smartdevicelink.transport.SdlPsm.transitionOnInput
### Bug Report
Fatal Crash 
```
Fatal Exception: java.lang.NegativeArraySizeException: -592969471
       at com.smartdevicelink.transport.SdlPsm.transitionOnInput(SdlPsm.java:241)
       at com.smartdevicelink.transport.SdlPsm.handleByte(SdlPsm.java:88)
       at com.smartdevicelink.transport.MultiplexBluetoothTransport$ConnectedThread.run(MultiplexBluetoothTransport.java:801)

```

##### Reproduction Steps
Seeing crash in Firebase Crashlytics dash, unfortunately no reproduction steps to share. 

##### Expected Behavior
No Crash

##### Observed Behavior
Crashing

* Android Version: 46% Android 11, 45% Android 10, 5% Android 7%
* SDL Android Version: 4.12.0
* Testing Against: 
According to crash logs. app was always in background when the crash was seen.

[Paste a link to a PR, gist, or other code that exemplifies this behavior]

joeygrover:
Fixes #1678 

### Risk

### Testing Plan
- [x] I have tested this PR against an IVI system

#### Unit Tests
Additional tests were added to test the SdlPsm.java class and some tests were refactored. These included more corner case issues as provided in #1678. 

#### Core Tests
Basic smoke testing can be done to ensure the packet state machine still works as intended. It is difficult to perform tests that would have actually broken the PSM so we rely on the unit tests for those scenarios.

### Summary
- Surrounded the base operation of the SdlPsm class in a try/catch that will set the state to ERROR state and safely exit. This prevents exceptions from being thrown in developer's applications. A log is written out through the DebugTool
- Added a check for correct service types. This helps with recovery tremendously. 
- Added better dataSize checking per early versions of the protocol. Dynamic MTUs are not included, but adding them might not actually help that much since they could be extremely large anyways.
- Refactored some checks to reduce redundant logic.
 
##### Enhancements
* Added better checks into the SdlPsm class
 
##### Bug Fixes
* Added safeguards to prevent crashes from propagating to developer's apps.

### CLA

### DIFF
--- a/base/src/main/java/com/smartdevicelink/transport/SdlPsm.java
+++ b/base/src/main/java/com/smartdevicelink/transport/SdlPsm.java
@@ -32,15 +32,19 @@
 package com.smartdevicelink.transport;
 
 import com.smartdevicelink.protocol.SdlPacket;
+import com.smartdevicelink.protocol.enums.SessionType;
+import com.smartdevicelink.util.DebugTool;
 
 import static com.smartdevicelink.protocol.SdlProtocol.V1_HEADER_SIZE;
 import static com.smartdevicelink.protocol.SdlProtocol.V1_V2_MTU_SIZE;
+import static com.smartdevicelink.protocol.SdlProtocol.V2_HEADER_SIZE;
+import static com.smartdevicelink.protocol.SdlProtocol.V3_V4_MTU_SIZE;
 
 
 public class SdlPsm {
-    //private static final String TAG = "Sdl PSM";
-    //Each state represents the byte that should be incoming
+    private static final String TAG = "Sdl PSM";
 
+    //Each state represents the byte that should be incoming
     public static final int START_STATE = 0x0;
     public static final int SERVICE_TYPE_STATE = 0x02;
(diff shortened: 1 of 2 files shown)


base/src/main/java/com/smartdevicelink/transport/SdlPsm.java
```diff
@@ -32,15 +32,19 @@
 package com.smartdevicelink.transport;
 
 import com.smartdevicelink.protocol.SdlPacket;
+import com.smartdevicelink.protocol.enums.SessionType;
```
JulianKast, line 35: This added import is unused
```suggestion
```

JulianKast (changes requested):

JulianKast (approved):

PR #1716, opened 2021-07-07
Notify ChoiceSetManager of failed preloads

### LINKED ISSUE #1715: Choices not saved as preloaded if some choices fail
### Bug Report
The ChoiceSetManager currently only saves preloaded choices if all choices succeed in the PreloadChoicesOperation.
If any choices fail then the choices are not saved. We need to still save any successfully uploaded choices to avoid reuploading them in the future.

RHenigan:
Fixes #1715 

### Risk

### Testing Plan

#### Unit Tests
N/A

#### Core Tests
To test this fix I had to update the preloadCells() method in the PreloadChoicesOperation. 

```
private void preloadCells() {
        isRunning = true;
        final List<CreateInteractionChoiceSet> choiceRPCs = new ArrayList<>(cellsToUpload.size());
        for (ChoiceCell cell : cellsToUpload) {
            CreateInteractionChoiceSet csCell = choiceFromCell(cell);
            if (csCell != null) {
                if(choiceRPCs.size() == 0) {
                    csCell.getChoiceSet().get(0).setMenuName(null);
                }
                choiceRPCs.add(csCell);
            }
        }
```

This will set the menuName of the first csCell to null so it is forced to fail. Now you can debug the preloadChoices method of the BaseChoiceSetManager in the PreloadChoicesOperation onComplete listener you should observe that the preloadedChoices and pendingPreloadChoices arrays is uploaded with the correct choice cells.

### Summary
This PR adds a new ChoiceSetCompletionListener to be used by the PreloadChoicesOperation. In the even that choices fail to preload the new listener is used to notify the BaseChoiceSetManager which choice cells failed to preload.Any choice cells that were successfull will be added to the preloadedChoices and removed from the pendingPreloadChoices.

### CLA


JulianKast (commented):
So if some choices fail to upload while others are successful, when they go to present that choiceSet, should the choices that successfully uploaded be presented?

RHenigan:
@JulianKast If you try to present a list that requires preloads and some of those preloads failed then the entire present should fail. In the event that a choice set has "Yes" and "No" but yes fails to preload, we do not want to present just "no".

JulianKast (approved):

base/src/main/java/com/smartdevicelink/managers/ChoiceSetCompletionListener.java
```diff
@@ -0,0 +1,16 @@
+package com.smartdevicelink.managers;
+
+import com.smartdevicelink.managers.screen.choiceset.ChoiceCell;
+
+import java.util.HashSet;
+
+public interface ChoiceSetCompletionListener {
```
JulianKast, line 7: Can you make this a part of BaseChoiceSetManager like AlertSoftButtonClearListener in BaseAlertManager?

JulianKast (changes requested):

JulianKast (approved):

PR #1724, opened 2021-08-04
[SDL 0317] Protocol Security Spec

### LINKED ISSUE #1720: Security queries are not implemented to spec
### Bug Report
With the acceptance of [SDL 0317](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0317-sdl-protocol-security-specification.md), this library's implementation of security queries does not match the spec fully. Several issues exist with the current implementation such as:

1. When a security query is received from Core, its header is entirely ignored and the library assumes that it is a `Send Handshake Data` request. This causes issues if Core were to send another type of query, such as a `Send Internal Error` notification.

```java
        System.arraycopy(msg.getData(), 12, data, 0, iLen);

        byte[] dataToRead = new byte[4096];

        Integer iNumBytes = sdlSecurity.runHandshake(data, dataToRead);
```

2. Hardcoded values are used when sending security queries for fields such as query ID. In addition, some values are not explicitly set at all, such as query type and sequential number. Any predefined values should be properly added as constants.

```java
        protocolMessage.setFunctionID(0x01);
```

should be changed to something like

```java
        securityQuery.setQueryID(QueryID.SEND_HANDSHAKE_DATA);
        securityQuery.setQueryType(QueryType.RESPONSE);
        securityQuery.setSequenceNumber(msg.getSequenceNumber());
```

3. Security queries are constructed using an RPC header, which has a slightly different format than a security query header. Security query headers should be implemented separate from RPC headers.

```java
        ProtocolMessage protocolMessage = new ProtocolMessage();
        ...
        sendMessage(protocolMessage);
```

should be changed to something like

```java
        SecurityQuery securityQuery = new SecurityQuery();
        ...
        sendMessage(securityQuery);
```

* Android Version: N/A
* SDL Android Version: master
* Testing Against: N/A

RHenigan:
Fixes #1720

### Risk

### Testing Plan
- [X] I have tested Android, Java SE, and Java EE

#### Unit Tests
Add unit tests for new enum types

#### Core Tests
Test to verify Encryption still works, try sending an encrypted RPC from the test app.
To verify error logs work correctly update the `ProtocolMessage` to send junk data in `BaseSdlSession.processControlService`
```
protocolMessage.setData(new byte[iNumBytes + 12]);
```
This should trigger the error logs in processControlService here `DebugTool.logError(TAG, "Client internal error: " + receivedHeader.getErrorCode().getName());` when core responds

Core version / branch / commit hash / module tested against: 7.1.1
HMI name / version / branch / commit hash / module tested against: Sdl Hmi

### Summary
This PR updates the processControlService method of BaseSdlSession to correctly ready the Binary Security Header sent by core. Previously we would always call runHandshake regardless of what core sent. Now we want to only call runHandshake if the header recieved by core has an ID of SEND_HANDSHAKE_DATA and a type of NOTIFICATION or REQUEST.

### CLA


base/src/main/java/com/smartdevicelink/protocol/enums/QueryID.java
```diff
@@ -0,0 +1,102 @@
...
+public class QueryID {
+
+    private static final Vector<QueryID> theList = new Vector<>();
+
+    public static Vector<QueryID> getList() {
+        return theList;
+    }
+
+    private static final byte[] sendHandshakeDataByteArray= {(byte) 0x00, (byte) 0x00, (byte) 0x01};
+    private static final byte[] sendInternalErrorByteArray= {(byte) 0x00, (byte) 0x00, (byte) 0x02};
```
JulianKast, line 19:
```suggestion
    private static final byte[] sendHandshakeDataByteArray = {(byte) 0x00, (byte) 0x00, (byte) 0x01};
    private static final byte[] sendInternalErrorByteArray = {(byte) 0x00, (byte) 0x00, (byte) 0x02};
```

base/src/main/java/com/smartdevicelink/protocol/enums/QueryType.java
```diff
@@ -0,0 +1,38 @@
+package com.smartdevicelink.protocol.enums;
+
+import com.smartdevicelink.util.ByteEnumer;
+
+import java.util.Vector;
+
+public class QueryType extends ByteEnumer {
```
JulianKast, line 7: Should we restrict this to scope library?

base/src/main/java/com/smartdevicelink/protocol/enums/QueryID.java
```diff
@@ -0,0 +1,102 @@
+package com.smartdevicelink.protocol.enums;
+
+import com.smartdevicelink.util.BitConverter;
+
+import java.util.Arrays;
+import java.util.Enumeration;
+import java.util.Objects;
+import java.util.Vector;
+
+public class QueryID {
```
JulianKast, line 10: Should we restrict this to scope library?

base/src/main/java/com/smartdevicelink/protocol/enums/QueryErrorCode.java
```diff
@@ -0,0 +1,58 @@
+package com.smartdevicelink.protocol.enums;
+
+import com.smartdevicelink.util.ByteEnumer;
+
+import java.util.Vector;
+
+public class QueryErrorCode extends ByteEnumer {
```
JulianKast, line 7: Should we restrict to scope Library?

base/src/main/java/com/smartdevicelink/session/BaseSdlSession.java
```diff
@@ -187,19 +190,57 @@ public void setSdlSecurity(SdlSecurityBase sec) {
...
+                DebugTool.logError(TAG, "Client internal error: " + receivedHeader.getErrorCode().getName());
+            } else {
+                DebugTool.logError(TAG, "Client internal error: No information provided");
+            }
+            return;
+        }
 
-        if (iNumBytes == null || iNumBytes <= 0)
+        if (receivedHeader.getQueryID() != QueryID.SEND_HANDSHAKE_DATA
+                && (receivedHeader.getQueryType() != QueryType.NOTIFICATION || receivedHeader.getQueryType() != QueryType.REQUEST)) {
```
jacobkeeler, line 219: `receivedHeader.getQueryType() != QueryType.NOTIFICATION || receivedHeader.getQueryType() != QueryType.REQUEST` is always true, because the queryType cannot be both `NOTIFICATION` and `REQUEST` at the same time. One of these expressions will always be `true`.

base/src/main/java/com/smartdevicelink/session/BaseSdlSession.java
```diff
@@ -185,21 +188,69 @@ public void setSdlSecurity(SdlSecurityBase sec) {
...
         byte[] dataToRead = new byte[4096];
 
-        Integer iNumBytes = sdlSecurity.runHandshake(data, dataToRead);
+        Integer iNumBytes = null;
+
+        // If the query is of type `Notification` and the id represents a client internal error, we abort the response message and the encryptionManager will not be in state ready.
+        if (receivedHeader.getQueryID() == SecurityQueryID.SEND_INTERNAL_ERROR
+                && receivedHeader.getQueryType() == SecurityQueryType.NOTIFICATION) {
+            if (receivedHeader.getErrorCode() != null) {
+                DebugTool.logError(TAG, "Security Query module internal error:" + receivedHeader.getErrorCode().getName());
```
JulianKast, line 218:
```suggestion
                DebugTool.logError(TAG, "Security Query module internal error: " + receivedHeader.getErrorCode().getName());
```

JulianKast (changes requested):
Just one space needs to be added in a log statement

JulianKast (approved):

PR #1726, opened 2021-08-18
Update Handler constructors from deprecated versions

### LINKED ISSUE #1696: Handler instances using deprecated constructors
### Bug Report
As of Android SDK 30 the no parameter and callback constructors of the `Handler` class were deprecated. Now a `Looper` instance is required to be passed. The library uses the default,no param constructor in most cases.

See here: https://developer.android.com/reference/android/os/Handler#Handler()

##### Expected Behavior
The library does not use deprecated methods from the Android API.

##### Observed Behavior
The library uses deprecated constroctures for `Handler` instances

* Android Version: SDK 30
* SDL Android Version: 5.1.1

###### Two solutions are possible:

**Update handler constructors**

```java
Handler handler = new Handler();
```

Needs to be replaced with

```java
Handler handler = new Handler(Looper.getMainLooper());
```

**Switch to using TaskMaster**

Replace `Handler` with a queue from task master and create tasks to be used. This solution will not work in the cases where a `Handler` is passed to a Messenger instance, for example the binding between `TransportBroker` and `SdlRouterService
`.

RHenigan:
Fixes #1696

### Risk

### Testing Plan
- [X] I have tested Android

#### Unit Tests
N/A

#### Core Tests
Smoke Test app still connects and videoStreaming still works

### Summary
The Handler constructor with no parameters has now been depreacted. To avoid using deprecated methods this PR will update the Handler constructors to take a Looper as a parameter. In the VirtualDisplayEncoder if we receive a message terminate then we quit the looper, in this case we will use myLooper rather than getMainLooper as we cannot quit the main thread.

### CLA

### DIFF
--- a/android/sdl_android/src/main/java/com/smartdevicelink/transport/MultiplexBluetoothTransport.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/transport/MultiplexBluetoothTransport.java
@@ -348,7 +348,7 @@ public class MultiplexBluetoothTransport extends MultiplexBaseTransport {
     }
 
     private void timerDelayRemoveDialog(final BluetoothSocket sock) {
-        timeOutHandler = new Handler();
+        timeOutHandler = new Handler(Looper.getMainLooper());
         socketRunnable = new Runnable() {
             public void run() {
                 //Log.e(TAG, "BLUETOOTH SOCKET CONNECT TIMEOUT - ATTEMPT TO CLOSE SOCKET");
--- a/android/sdl_android/src/main/java/com/smartdevicelink/transport/MultiplexBluetoothTransport.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/transport/MultiplexBluetoothTransport.java
@@ -348,7 +348,10 @@ public class MultiplexBluetoothTransport extends MultiplexBaseTransport {
     }
 
     private void timerDelayRemoveDialog(final BluetoothSocket sock) {
-        timeOutHandler = new Handler(Looper.getMainLooper());
+        if (Looper.myLooper() == null) {
+            Looper.prepare();
+        }
+        timeOutHandler = new Handler(Looper.myLooper());
         socketRunnable = new Runnable() {
             public void run() {
                 //Log.e(TAG, "BLUETOOTH SOCKET CONNECT TIMEOUT - ATTEMPT TO CLOSE SOCKET");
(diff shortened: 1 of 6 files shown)


JulianKast (commented):
Need to investigate using `Looper.myLooper` vs using `Looper.getMainLooper` especially since the RouterService runs on its own thread.

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -2501,7 +2504,7 @@ protected static LocalRouterService getLocalRouterService(Intent launchIntent, C
      * This method is used to check for the newest version of this class to make sure the latest and greatest is up and running.
      */
     private void startAltTransportTimer() {
-        altTransportTimerHandler = new Handler();
+        altTransportTimerHandler = new Handler(Looper.getMainLooper());
```
JulianKast, line 2507: Should we use myLooper instead of the main looper here?

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -3057,7 +3060,7 @@ public RegisteredApp(String appId, int routerMessagingVersion, Messenger messeng
             this.messenger = messenger;
             this.sessionIds = new Vector<Long>();
             this.queues = new ConcurrentHashMap<>();
-            queueWaitHandler = new Handler();
+            queueWaitHandler = new Handler(Looper.getMainLooper());
```
JulianKast, line 3063: Should we use myLooper instead of the main looper here?

android/sdl_android/src/main/java/com/smartdevicelink/transport/MultiplexBluetoothTransport.java
```diff
@@ -348,7 +348,7 @@ private void connectionLost() {
     }
 
     private void timerDelayRemoveDialog(final BluetoothSocket sock) {
-        timeOutHandler = new Handler();
+        timeOutHandler = new Handler(Looper.getMainLooper());
```
JulianKast, line 351: Should we use myLooper instead of the main looper here?

JulianKast (changes requested):

JulianKast (approved):

PR #1737, opened 2021-09-23
Fix file operations check file system at queue time instead of run time

### LINKED ISSUE #1736: File Manager will upload the same file multiple times
### Bug Report
When trying to load multiple of the same artwork and the file is not uploaded yet, the file will be uploaded multiple times. This is because we create multiple of the same upload operation and they all execute.

We should probably implement the upload / delete operation to check if the file needs to be uploaded at the time it would run instead of the manager figuring out if the operation should be created at all.

##### Reproduction Steps
1. Create 99 SDLArtworks that are exactly the same
2. Attempt to upload all 99 in a loop

##### Expected Behavior
The one artwork should only be uploaded 1 time.

##### Observed Behavior
The one artwork will be uploaded 99 times, slowing down the head unit like crazy.

* Android Version: n/a
* SDL Android Version: 5.1.1
* Testing Against: manticore

JulianKast:
Fixes #1736 

### Risk

### Testing Plan

#### Unit Tests
Unit Test were added to FileManagerTest

#### Core Tests
Tested uploading a list of 20 files that were all the same, verify that file only gets sent to core once.
Tested uploaded a file and tried to delete in multiple times with a list of fileNames.

Core version / branch / commit hash / module tested against: Core 7.1.1
HMI name / version / branch / commit hash / module tested against: genericHMI 0.10.0

### Summary
This PR updates the File Manager to check if a file is already on the system when an operation starts instead of when an operation is queued. This fixes issues if there is a change between when a file upload or delete is queued rather than when it's run.

##### Bug Fixes
* Fixed issues where file operations could fail if file state changes between when the operation is queued and when it runs. For example, attempting to upload 99 of the same file will cause 99 of that file to actually be uploaded if the file is not on the system when those 99 operations are queued. Now the system will only upload the file once.

### CLA


RHenigan (commented):
SdlArtwork has unused imports (Android and java)
SdlFile (Android and java) also should implement cloneable

base/src/main/java/com/smartdevicelink/managers/file/DeleteFileOperation.java
```diff
@@ -66,6 +70,13 @@ private void start() {
         if (getState() == Task.CANCELED) {
             return;
         }
+        if (!mutableRemoteFileNames.contains(fileName) && completionListener != null) {
```
RHenigan, line 73:
```suggestion
        if (!mutableRemoteFileNames.contains(fileName)) {
            if (completionListener != null) {
                String errorMessage = "File to delete is no longer on the head unit, aborting operation";
                // Returning BaseFileManager.SPACE_AVAILABLE_MAX_VALUE for bytesAvaialble as a placeHolder, it will not get updated in BaseFileManager as long as success returned is false.
                completionListener.onComplete(false, BaseFileManager.SPACE_AVAILABLE_MAX_VALUE, mutableRemoteFileNames, errorMessage);
            }
            onFinished();
            return;
        }
```

Do we still want to call onFinished and return if the completionListener is null?

RHenigan (approved):

PR #1776, opened 2021-12-07
Fix soft button object invalid states

### LINKED ISSUE #1774: SoftButtonObject doesn't handle error states correctly
### Bug Report
The SoftButtonObject can be created with states that don't exist or are incorrect. We should check these states on construction to ensure that developers can't send incorrect soft buttons.

1. Creating a `SoftButtonObject` with empty states should throw an exception
2. Creating a `SoftButtonObject` with states where the initial state name is not among the set states.

##### Reproduction Steps
1. Create a soft button object with empty state list
2. Send the soft button object

##### Expected Behavior
The soft buttons should throw an exception

##### Observed Behavior
The soft buttons return a success because no soft buttons are sent in the `Show`

* Android Version: n/a
* SDL Android Version: v5.3.1
* Testing Against: Sync 3

n/a

noah-livio:
Fixes #1774

This PR is ready for review.

### Risk
This PR adds a new public getter function to SoftButtonObject

### Testing Plan
- [x] I have verified that I have not introduced new warnings in this PR
- [x] I have tested this PR against Core and verified behavior
- [x] I have tested Android and Java SE

#### Unit Tests
 - `testConstructSoftButtonObjectWithEmptyStateList()` passes if `getAttemptedToAssignEmptyStateList()` returns true after an empty state list is passed to the main primary constructor and fails otherwise.

- `testAssignEmptyStateListToSoftButtonObject()` passes if an `getAttemptedToAssignEmptyStateList()` returns true after an empty state list is passed to `setStates()` of an instance of `SoftButtonObject` that already has at least one state in its state list. 

##### Tests performed by modifying a local clone of hello_sdl_android

- Constructed an instance of `SoftButtonObject` with a state list containing a single state, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
  - That instance along with all the other instances in the same transaction displayed successfully.
  - The app continued to function normally
- Constructed an instance of `SoftButtonObject` with an empty state list, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
   - The instance printed an error with `DebugTool` stating that the state list is empty
   - That instance and all other other instances within the same transaction failed to display to the HMI
   - The app otherwise continued to function normally
-  Constructed an instance of `SoftButtonObject` with a state list containing a single state, reassigned the state list to an empty list using `setStates()`, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results:
   - The instance printed an error with `DebugTool` stating that the state list is empty
   - The instance did not set overwrite the old state list with the new state list
   - The app otherwise continued to function normally

SDL Core / 8.0.0 / master / 68f082169e0a40fccd9eb0db3c83911c28870f07 / module tested against: SDL Core on Ubuntu 20.04.3 LTS Virtual Machine
Generic HMI / 0.11.0 / master / 47e0ad42f05b27adff61befd864e79c2ab4b8cec / module tested against: Generic HMI on Ubuntu 20.04.3 LTS Virtual Machine

### Summary
This pull request adds an if statement to `SoftButtonObject.setStates()` that checks if the given state list is empty and displays  if it is. This PR also replaces the direct state list assignment in the main constructor with usage of `setStates()` to enforce the check in the constructor without writing redundant code.

##### Breaking Changes
* None

##### Enhancements
* Adds a getter function to check if an empty list was passed to the instance 

##### Bug Fixes
* Fixes #1774

### CLA

### DIFF
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
@@ -455,10 +455,31 @@ public class SoftButtonManagerTests {
     }
 
     /**
-     * Test assigning a nonempty state list to existing SoftButtonObject
+     * Test assigning a state list with the current state to existing SoftButtonObject
      */
     @Test
-    public void testAssignNonEmptyStateListToSoftButtonObject() {
+    public void testAssignStateListWithCurrentStateToSoftButtonObject() {
+        List<SoftButtonState> stateList1 = new ArrayList<>();
+        SoftButtonState softButtonState1 = new SoftButtonState("hello_there", "Hello there", null);
+        stateList1.add(softButtonState1);
+
+        List<SoftButtonState> stateList2 = new ArrayList<>();
+        SoftButtonState softButtonState2 = new SoftButtonState("general_kenobi", "General Kenobi", null);
+        stateList2.add(softButtonState1);
+        stateList2.add(softButtonState2);
+
+        SoftButtonObject softButtonObject = new SoftButtonObject("general_kenobi", stateList1, "hello_there", null);
+
+        softButtonObject.setStates(stateList2);
+
+        assertEquals(stateList2, softButtonObject.getStates());
+    }
+
+    /**
+     * Test assigning a state list without the current state to existing SoftButtonObject
+     */
+    @Test
+    public void testAssignStateListWithoutCurrentStateToSoftButtonObject() {
         List<SoftButtonState> stateList1 = new ArrayList<>();
         SoftButtonState softButtonState1 = new SoftButtonState("hello_there", "Hello there", null);
         stateList1.add(softButtonState1);
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
@@ -351,8 +351,8 @@ public class SoftButtonManagerTests {
         softButtonStateList.add(softButtonState1);
         softButtonStateList2.add(softButtonState1);
         softButtonStateList2.add(softButtonState2);
-        softButtonObject1 = new SoftButtonObject("hi", softButtonStateList, "Hi", null);
-        softButtonObject2 = new SoftButtonObject("hi", softButtonStateList2, "Hi", null);
+        softButtonObject1 = new SoftButtonObject("hi", softButtonStateList, softButtonStateList.get(0).getName(), null);
+        softButtonObject2 = new SoftButtonObject("hi", softButtonStateList2, softButtonStateList2.get(0).getName(), null);
         assertNotEquals(softButtonObject1, softButtonObject2);
 
         // Case 5: SoftButtonStates are not the same, assertFalse
@@ -366,8 +366,8 @@ public class SoftButtonManagerTests {
         assertNotEquals(softButtonObject1, softButtonObject2);
 
         // Case 7: SoftButtonObject currentStateName not same, assertFalse
-        softButtonObject1 = new SoftButtonObject("hi", softButtonStateList, "Hi", null);
-        softButtonObject2 = new SoftButtonObject("hi", softButtonStateList, "Hi2", null);
+        softButtonObject1 = new SoftButtonObject("hi", softButtonStateList2, softButtonStateList2.get(0).getName(), null);
+        softButtonObject2 = new SoftButtonObject("hi", softButtonStateList2, softButtonStateList2.get(1).getName(), null);
         assertNotEquals(softButtonObject1, softButtonObject2);
     }
 
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
@@ -33,7 +33,6 @@ import com.smartdevicelink.proxy.rpc.enums.SystemCapabilityType;
 import com.smartdevicelink.proxy.rpc.listeners.OnRPCNotificationListener;
 import com.smartdevicelink.test.Validator;
 
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -58,14 +57,11 @@ import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import android.database.sqlite.SQLiteCantOpenDatabaseException;
-
 /**
  * This is a unit test class for the SmartDeviceLink library manager class :
  * {@link SoftButtonManager}
  */
 @RunWith(AndroidJUnit4.class)
-//@RunWith(JUnit4.class)
 public class SoftButtonManagerTests {
 
     private SoftButtonManager softButtonManager;
@@ -414,20 +410,20 @@ public class SoftButtonManagerTests {
     @Test
     public void testConstructSoftButtonObjectWithEmptyStateList() {
         List<SoftButtonState> stateList = new ArrayList<>();
-        SoftButtonObject softButtonObject1 = new SoftButtonObject("hello_there", stateList, "general_kenobi", null);
-        assertNull(softButtonObject1.getStates());
+        SoftButtonObject softButtonObject = new SoftButtonObject("hello_there", stateList, "general_kenobi", null);
+        assertNull(softButtonObject.getStates());
     }
 
     /**
      * Test constructing SoftButtonObject with an nonempty state list
      */
     @Test
-    public void testConstructingSoftButtonObjectWithNonEmptyStateList() {
+    public void testConstructSoftButtonObjectWithNonEmptyStateList() {
         List<SoftButtonState> stateList = new ArrayList<>();
         SoftButtonState softButtonState = new SoftButtonState("general_kenobi", "General Kenobi", null);
         stateList.add(softButtonState);
-        SoftButtonObject softButtonObject2 = new SoftButtonObject("hello_there_again", stateList, "general_kenobi", null);
-        assertEquals(stateList, softButtonObject2.getStates());
+        SoftButtonObject softButtonObject = new SoftButtonObject("hello_there", stateList, "general_kenobi", null);
+        assertEquals(stateList, softButtonObject.getStates());
     }
 
     /**
@@ -435,13 +431,15 @@ public class SoftButtonManagerTests {
      */
     @Test
     public void testAssignEmptyStateListToSoftButtonObject() {
-        List<SoftButtonState> stateList = new ArrayList<>();
-        SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
+        List<SoftButtonState> nonEmptyStateList = new ArrayList<>();
+        List<SoftButtonState> emptyStateList = new ArrayList<>();
+        SoftButtonState softButtonState = new SoftButtonState("general_kenobi", "General Kenobi", null);
+        nonEmptyStateList.add(softButtonState);
 
-        SoftButtonObject softButtonObject = new SoftButtonObject("hi", softButtonState, null);
+        SoftButtonObject softButtonObject = new SoftButtonObject("hello_there", nonEmptyStateList, "general_kenobi", null);
 
-        softButtonObject.setStates(stateList);
-        assertNotEquals(stateList, softButtonObject.getStates());
+        softButtonObject.setStates(emptyStateList);
+        assertEquals(nonEmptyStateList, softButtonObject.getStates());
     }
 
     /**
@@ -449,13 +447,18 @@ public class SoftButtonManagerTests {
      */
     @Test
     public void testAssignNonEmptyStateListToSoftButtonObject() {
-        List<SoftButtonState> stateList = new ArrayList<>();
-        SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
+        List<SoftButtonState> stateList1 = new ArrayList<>();
+        SoftButtonState softButtonState1 = new SoftButtonState("hello_there", "Hello there", null);
+        stateList1.add(softButtonState1);
 
-        SoftButtonObject softButtonObject = new SoftButtonObject("hi", softButtonState, null);
+        List<SoftButtonState> stateList2 = new ArrayList<>();
+        SoftButtonState softButtonState2 = new SoftButtonState("general_kenobi", "General Kenobi", null);
+        stateList2.add(softButtonState2);
 
-        stateList.add(softButtonState);
-        softButtonObject.setStates(stateList);
-        assertEquals(stateList, softButtonObject.getStates());
+        SoftButtonObject softButtonObject = new SoftButtonObject("general_kenobi", stateList1, "general_kenobi", null);
+
+        softButtonObject.setStates(stateList2);
+
+        assertEquals(stateList2, softButtonObject.getStates());
     }
 }
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
@@ -455,7 +455,7 @@ public class SoftButtonManagerTests {
         SoftButtonState softButtonState2 = new SoftButtonState("general_kenobi", "General Kenobi", null);
         stateList2.add(softButtonState2);
 
-        SoftButtonObject softButtonObject = new SoftButtonObject("general_kenobi", stateList1, "general_kenobi", null);
+        SoftButtonObject softButtonObject = new SoftButtonObject("general_kenobi", stateList1, "hello_there", null);
 
         softButtonObject.setStates(stateList2);
 
@@ -477,7 +477,7 @@ public class SoftButtonManagerTests {
         SoftButtonState softButtonState3 = new SoftButtonState("general_kenobi", "General Kenobi Again", null);
         stateListDuplicateNames.add(softButtonState3);
 
-        SoftButtonObject softButtonObject = new SoftButtonObject("general_kenobi", stateListUnique, "general_kenobi", null);
+        SoftButtonObject softButtonObject = new SoftButtonObject("general_kenobi", stateListUnique, "hello_there", null);
 
         softButtonObject.setStates(stateListDuplicateNames);
 
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
@@ -427,6 +427,18 @@ public class SoftButtonManagerTests {
     }
 
     /**
+     * Test constructing SoftButtonObject with an invalid initialStateName
+     */
+    @Test
+    public void testConstructSoftButtonObjectWithInvalidInitialStateName() {
+        List<SoftButtonState> stateList = new ArrayList<>();
+        SoftButtonState softButtonState = new SoftButtonState("general_kenobi", "General Kenobi", null);
+        stateList.add(softButtonState);
+        SoftButtonObject softButtonObject = new SoftButtonObject("hello_there", stateList, "hello_there", null);
+        assertNull(softButtonObject.getStates());
+    }
+
+    /**
      * Test assigning an empty state list to existing SoftButtonObject
      */
     @Test
--- a/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
+++ b/android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
@@ -44,6 +44,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+
 import static junit.framework.TestCase.assertEquals;
 import static junit.framework.TestCase.assertFalse;
 import static junit.framework.TestCase.assertNull;
@@ -56,13 +57,12 @@ import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import junit.framework.TestCase;
-
 /**
  * This is a unit test class for the SmartDeviceLink library manager class :
  * {@link SoftButtonManager}
  */
 @RunWith(AndroidJUnit4.class)
+//@RunWith(JUnit4.class)
 public class SoftButtonManagerTests {
 
     private SoftButtonManager softButtonManager;
@@ -411,14 +411,12 @@ public class SoftButtonManagerTests {
     @Test
     public void testConstructSoftButtonObjectWithEmptyStateList() {
         List<SoftButtonState> stateList = new ArrayList<>();
+        // Uncommenting the following lines should make the test fail
+        // SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
+        // stateList.add(softButtonState);
+        SoftButtonObject softButtonObject = new SoftButtonObject("hi", stateList, "Hi", null);
 
-        try {
-            new SoftButtonObject("hi", stateList, "Hi", null);
-            TestCase.fail("IllegalStateException expected");
-        }
-        catch (IllegalStateException ignored) {
-
-        }
+        assertTrue(softButtonObject.getAttemptedToAssignEmptyStateList());
     }
 
     /**
@@ -429,14 +427,9 @@ public class SoftButtonManagerTests {
         List<SoftButtonState> stateList = new ArrayList<>();
         SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
 
-
-        try {
-            SoftButtonObject softButtonObject = new SoftButtonObject("hi", softButtonState, null);
-            softButtonObject.setStates(stateList);
-            TestCase.fail("IllegalStateException expected");
-        }
-        catch (IllegalStateException ignored) {
-
-        }
+        SoftButtonObject softButtonObject = new SoftButtonObject("hi", softButtonState, null);
+        // Commenting the following line should make the test fail
+        softButtonObject.setStates(stateList);
+        assertTrue(softButtonObject.getAttemptedToAssignEmptyStateList());
     }
 }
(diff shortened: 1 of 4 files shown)


android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -402,4 +404,32 @@ public void testSoftButtonStateEquals() {
...
+     */
+    @Test
+    public void testConstructSoftButtonObjectWithEmptyStateList() {
+        List<SoftButtonState> stateList = new ArrayList<>();
+        // Uncommenting the following lines should make the test fail
+        // SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
+        // stateList.add(softButtonState);
+        SoftButtonObject softButtonObject = new SoftButtonObject("hi", stateList, "Hi", null);
+
+        assertTrue(softButtonObject.getAttemptedToAssignEmptyStateList());
```
RHenigan, line 419: Without the boolean flag we will need to change our assertions.
since we will return out of the constructor early most of our fields will still be null so we can use those for our assertions.
`assertNull(softButtonObject.getName());`
and if we make a SoftButtonObject with a non-empty state list we can assert the fields are set to what was passed in the constructor
`assertEquals("hi", softButtonObject.getName());`

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -402,4 +404,32 @@ public void testSoftButtonStateEquals() {
...
+     * Test assigning an empty state list to existing SoftButtonObject
+     */
+    @Test
+    public void testAssignEmptyStateListToSoftButtonObject() {
+        List<SoftButtonState> stateList = new ArrayList<>();
+        SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
+
+        SoftButtonObject softButtonObject = new SoftButtonObject("hi", softButtonState, null);
+        // Commenting the following line should make the test fail
+        softButtonObject.setStates(stateList);
```
RHenigan, line 432: setStates should be returning early if it is passed an empty list, this means the states on the object will not be updated. to test this we can make an assertion before setting the state that the list is what is expected from the constructor.
`assertEquals(Collections.singletonList(softButtonState), softButtonObject.getStates());`
and then after trying to set the states to an empty list we can make the same assertion as setStates should return early and not update the field
`assertEquals(Collections.singletonList(softButtonState), softButtonObject.getStates());`

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
```diff
@@ -395,4 +404,13 @@ public SoftButtonObject clone() {
         }
         return null;
     }
+
+    /**
+     * Used to make unit testing easier by removing the need to read debug logs programmatically
+     *
+     * @return True if the last list passed to setStates() was empty and false otherwise
+     */
+    public boolean getAttemptedToAssignEmptyStateList() {
```
RHenigan, line 413: This will not be needed if we remove our boolean flag.

RHenigan (changes requested):

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
```diff
@@ -264,6 +266,13 @@ public void setName(@NonNull String name) {
      * @param states a list of the object's soft button states. <strong>states should be unique for every SoftButtonObject. A SoftButtonState instance cannot be reused for multiple SoftButtonObjects.</strong>
      */
     public void setStates(@NonNull List<SoftButtonState> states) {
+        // If the list of states is empty, throw an error with DebugTool and return
+        attemptedToAssignEmptyStateList = states.isEmpty();
+        if (attemptedToAssignEmptyStateList) {
+            DebugTool.logError(TAG,"The state list is empty");
+            return;
+        }
```
RHenigan, line 274: Since we are doing it the constructor i think it also makes sense to check the list does not have duplicate state names here as well.
```
if (hasTwoStatesOfSameName(states)) {
    //Log error
    //return
}
```

noah-livio: That sounds reasonable to me, I just wasn't sure I should increase the scope of this PR

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
noah-livio: In the constructor based test, I think going with something like `assertNull(softButtonObject.getStates());` and `assertEquals(stateList, softButtonObject2.getStates());` for the error case and normal case, respectively, will be more resilient to future changes since it directly tests the actual state list. I hope that is an okay alternative to your suggestion.

RHenigan: Wouldn't this test case fail when sending an empty list though?
In the constructor if the list is empty we should try to return early and none of the fields on that object will be set so those fields would be null?
If this is the case the assertion would fail because you are comparing an empty list to null.
`assertEquals(stateList, softButtonObject2.getStates())` stateList being an empty list and getStates() returning null.

You can write a second test case to test the event that a non-empty list is sent in the constructor. in this case yes we would want to verify getStates is returning the same list that was sent in the constructor

RHenigan: Sorry, I misread your previous comment. Yes I think that is ok
If the list is empty we will return early and should verify that the states remain null.
If the list is populated then we should verify the object if created with the correct states list.

noah-livio: It's all good, I'm just happy I'm going in the right direction in regard to them, I probably could have written my first comment more clearly to prevent misreading

noah-livio:
Fixes #1774

This PR is ready for review.

### Risk
None

### Testing Plan
- [x] I have verified that I have not introduced new warnings in this PR
- [x] I have tested this PR against Core and verified behavior
- [x] I have tested Android and Java SE

#### Unit Tests
- `testConstructSoftButtonObjectWithEmptyStateList()` passes if `SoftButtonObject.getStates()` returns null when the constructor is given an empty state list
- `testConstructingSoftButtonObjectWithNonEmptyStateList()` passes if `SoftButtonObject.getStates()` returns the nonempty state list passed to the constructor
- `testAssignNonEmptyStateListToSoftButtonObject()` passes if `SoftButtonObject.getStates()` returns a state list not equal to the empty state list passed to `SoftButtonObject.setStates()`
- `testAssignEmptyStateListToSoftButtonObject()` passes if `SoftButtonObject.getStates()` returns a state list equal to the nonempty state list passed to `SoftButtonObject.setStates()`


#### Core Tests

##### Tests performed by modifying a local clone of hello_sdl_android

- Constructed an instance of `SoftButtonObject` with a state list containing a single state, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
  - That instance along with all the other instances in the same transaction displayed successfully.
  - The app continued to function normally
- Constructed an instance of `SoftButtonObject` with an empty state list, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
   - The instance printed an error with `DebugTool` stating that the state list is empty
   - That instance and all other other instances within the same transaction failed to display to the HMI
   - The app otherwise continued to function normally
-  Constructed an instance of `SoftButtonObject` with a state list containing a single state, reassigned the state list to an empty list using `setStates()`, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results:
   - The instance printed an error with `DebugTool` stating that the state list is empty
   - The instance did not overwrite the old state list with the new state list
   - The app otherwise continued to function normally

SDL Core / 8.0.0 / master / 68f082169e0a40fccd9eb0db3c83911c28870f07 / module tested against: SDL Core on Ubuntu 20.04.3 LTS Virtual Machine
Generic HMI / 0.11.0 / master / 47e0ad42f05b27adff61befd864e79c2ab4b8cec / module tested against: Generic HMI on Ubuntu 20.04.3 LTS Virtual Machine

### Summary
This pull request adds an if statement to `SoftButtonObject.setStates()` along with an identical if statement in the main constructor that throws an error and returns if the state list passed is empty

### Changelog
##### Breaking Changes
* None

##### Enhancements
* None

##### Bug Fixes
* Fixes #1774

### CLA

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -33,6 +33,7 @@
 import com.smartdevicelink.proxy.rpc.listeners.OnRPCNotificationListener;
 import com.smartdevicelink.test.Validator;
 
+import org.junit.Assert;
```
RHenigan, line 36: We can remove this unused import

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -56,11 +58,14 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import android.database.sqlite.SQLiteCantOpenDatabaseException;
```
RHenigan, line 61: we can remove this. unused import

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -56,11 +58,14 @@
...
+
 /**
  * This is a unit test class for the SmartDeviceLink library manager class :
  * {@link SoftButtonManager}
  */
 @RunWith(AndroidJUnit4.class)
+//@RunWith(JUnit4.class)
```
RHenigan, line 68: We can remove this comment as well

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -402,4 +407,55 @@ public void testSoftButtonStateEquals() {
...
+     */
+    @Test
+    public void testAssignEmptyStateListToSoftButtonObject() {
+        List<SoftButtonState> stateList = new ArrayList<>();
+        SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
+
+        SoftButtonObject softButtonObject = new SoftButtonObject("hi", softButtonState, null);
+
+        softButtonObject.setStates(stateList);
+        assertNotEquals(stateList, softButtonObject.getStates());
```
RHenigan, line 444: Rather than assert that these are not equal we could assert that the state on the Object is left unchanged after calling setStates.
`assertEquals(softButtonState, softButtonObject.getStates)`
You may want to use the constructor that takes a list for states rather than a single state. or you could make a list out of `softButtonState` before you make the assertion

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -402,4 +407,55 @@ public void testSoftButtonStateEquals() {
...
+    @Test
+    public void testAssignNonEmptyStateListToSoftButtonObject() {
+        List<SoftButtonState> stateList = new ArrayList<>();
+        SoftButtonState softButtonState = new SoftButtonState("object1-state1", "o1s1", null);
+
+        SoftButtonObject softButtonObject = new SoftButtonObject("hi", softButtonState, null);
+
+        stateList.add(softButtonState);
+        softButtonObject.setStates(stateList);
+        assertEquals(stateList, softButtonObject.getStates());
```
RHenigan, line 459: For this test it may be beneficial to have stateList contain a state that is different than what was passed in the constructor. this way we are verifying the states are in fact being updated.

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
```diff
@@ -73,17 +73,22 @@
...
             DebugTool.logError(TAG, "Two states have the same name in states list for soft button object");
             return;
         }
 
         this.name = name;
-        this.states = states;
         this.currentStateName = initialStateName;
         this.buttonId = SOFT_BUTTON_ID_NOT_SET_VALUE;
         this.onEventListener = onEventListener;
+        this.states = states;
```
RHenigan, line 91: Can we put this back to the same line it was previously on? this will help to reduce noise in the diff for changes that aren't really changes

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
```diff
@@ -395,4 +406,5 @@ public SoftButtonObject clone() {
         }
         return null;
     }
+
```
RHenigan, line 409: just cleaning up whitespace

android/sdl_android/src/androidTest/java/com/smartdevicelink/managers/screen/SoftButtonManagerTests.java
```diff
@@ -44,6 +45,7 @@
 import java.util.Collections;
 import java.util.List;
 
+
```
RHenigan, line 48: just cleaning up white space.

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
RHenigan: After looking at the iOS implementation the iOS team will also need to make changes. With the teams input I think it makes sense to add this check to setStates

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
noah-livio: If you mean where it was when I was using setStates() in the constructor, I don't think I can without making the diff noisier, because now the check is happening there and the check obviously has to happen first to prevent assignment in the error case. This change is just reverting the states assignment back to how it was in develop before I forked. Maybe I could squash commits? But I think that might be more trouble than it's worth

joeygrover: Just move `this.states = states;` back under `this.name= name;`

noah-livio: Oh okay, will do

noah-livio:
Fixes #1774

This PR is ready for review.

### Risk
None

### Testing Plan
- [x] I have verified that I have not introduced new warnings in this PR
- [x] I have tested this PR against Core and verified behavior
- [x] I have tested Android and Java SE

#### Unit Tests
- `testConstructSoftButtonObjectWithEmptyStateList()` passes if `SoftButtonObject.getStates()` returns null when the constructor is given an empty state list
- `testConstructingSoftButtonObjectWithNonEmptyStateList()` passes if `SoftButtonObject.getStates()` returns the nonempty state list given to the constructor
- `testAssignNonEmptyStateListToSoftButtonObject()` passes if `SoftButtonObject.getStates()` returns the nonempty state list given to the constructor after `SoftButtonObject.setStates()` is given an empty state list
- `testAssignEmptyStateListToSoftButtonObject()` passes if `SoftButtonObject.getStates()` returns a state list equal to the nonempty state list passed to `SoftButtonObject.setStates()`


#### Core Tests

##### Tests performed by modifying a local clone of hello_sdl_android

- Constructed an instance of `SoftButtonObject` with a state list containing a single state, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
  - That instance along with all the other instances in the same transaction displayed successfully.
  - The app continued to function normally
- Constructed an instance of `SoftButtonObject` with an empty state list, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
   - The instance printed an error with `DebugTool` saying that the state list is empty
   - That instance and all other other instances within the same transaction failed to display to the HMI
   - The app otherwise continued to function normally
-  Constructed an instance of `SoftButtonObject` with a state list containing a single state, reassigned the state list to an empty list using `setStates()`, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results:
   - The instance printed an error with `DebugTool` saying that the state list is empty
   - The instance did not overwrite the old state list with the new state list
   - The app otherwise continued to function normally
-  Constructed an instance of `SoftButtonObject` with a state list containing a single state, reassigned the state list to a state list with states with duplicate names using `setStates()`, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results:
   - The instance printed an error with `DebugTool` saying "Two states have the same name in states list for soft button object"
   - The instance did not overwrite the old state list with the new state list
   - The app otherwise continued to function normally

SDL Core / 8.0.0 / master / 68f082169e0a40fccd9eb0db3c83911c28870f07 / module tested against: SDL Core on Ubuntu 20.04.3 LTS Virtual Machine
Generic HMI / 0.11.0 / master / 47e0ad42f05b27adff61befd864e79c2ab4b8cec / module tested against: Generic HMI on Ubuntu 20.04.3 LTS Virtual Machine

### Summary
This pull request adds an if statement to `SoftButtonObject.setStates()` along with an identical if statement in the main constructor that throws an error and returns if the state list passed is empty. This pull request also adds a similar check to `SoftButtonObject.setStates()` for state lists containing states with duplicate names.

### Changelog
##### Breaking Changes
* None

##### Enhancements
* None

##### Bug Fixes
* Fixes #1774

### CLA

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
```diff
@@ -73,6 +71,11 @@
      */
     public SoftButtonObject(@NonNull String name, @NonNull List<SoftButtonState> states, @NonNull String initialStateName, OnEventListener onEventListener) {
 
+        // If the list of states is empty, throw an error with DebugTool and return
+        if (states.isEmpty()) {
+            DebugTool.logError(TAG,"The state list is empty");
+            return;
+        }
```
joeljfischer, line 78: Alignment note:

On iOS I did this

```objc
BOOL hasStateWithInitialName = NO;
for (SDLSoftButtonState *state in states) {
    if ([state.name isEqualToString:initialStateName]) {
        hasStateWithInitialName = YES;
    }
}
NSAssert(![self sdl_hasTwoStatesOfSameName:states], @"A SoftButtonObject must have states with different names.");
NSAssert(hasStateWithInitialName, @"A SoftButtonObject must have a state with initialStateName.");
if ([self sdl_hasTwoStatesOfSameName:states]) { return nil; }
if (!hasStateWithInitialName) { return nil; }
```

The advantage this has is only throwing an exception in DEBUG builds (I don't know that is available on Java Suite). Second, this tests not only that there _are_ states, but also that the initial state name is available. The for loop won't run if there are no states, so the boolean will just stay false. That check at least should be a part of this PR as well.

Last, I also made a change to `SoftButtonState` on iOS to check if both text and image are `null`, which is an invalid state configuration. I think that should be a part of this PR as well.

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
joeljfischer: See my above comment

joeljfischer (changes requested):

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
noah-livio: Understood, I'll work on implementing this

noah-livio:
I have added the `initialStateName` check in the constructor, added `AssertionError`s that only throw when built in debug mode, and adjusted existing tests to avoid failing for the `initialStateName` check. I also added a debug only `AssertionError` to `SoftButtonState` in the check for null text and image.

RHenigan:
Older unit tests fail as a result of the new hasStateWithInitialName check.
Some tests are using empty lists and null or mismatched initia state names which are the states we are trying to avoid.

ScreenManagerTests.testAssigningIdsToSoftButtonObjects
SoftButtonManagerTests.testSoftButtonObjectEquals

RHenigan (changes requested):

RHenigan (changes requested):
`testAssignNonEmptyStateListToSoftButtonObject` is now failing, this test would need to set a list with a state that is equivalent to the currentState. We should also test to verify the negative is true as well (confirm setting a list without the currentState fails to update)

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
```diff
@@ -264,6 +275,26 @@ public void setName(@NonNull String name) {
...
+
+        if (repeatedStateNames) {
+            DebugTool.logError(TAG, "A SoftButtonObject must have states with different names.");
+            return;
+        }
+
+        boolean hasStateWithCurrentName = false;
+        for (SoftButtonState state : states) {
+            if(state.getName().equals(currentStateName)) {
+                hasStateWithCurrentNameName = true;
```
RHenigan, line 289:
```suggestion
                hasStateWithCurrentName = true;
```

noah-livio: Good catch, and that'll be easy to fix

RHenigan: Also to avoid confusion for the developer we will still want to set the states in the case the currentState is not present in the list, but we will still want to log the error.

We should test and confirm that the SoftButtonManager can handle this error case (trying to upload a SoftButton where the currentState is not in the list)

noah-livio:
Fixes #1774

This PR is ready for review.

### Risk
None

### Testing Plan
- [x] I have verified that I have not introduced new warnings in this PR
- [x] I have tested this PR against Core and verified behavior
- [x] I have tested Android and Java SE

#### Unit Tests
- `testConstructSoftButtonObjectWithEmptyStateList()` passes if `SoftButtonObject.getStates()` returns null when the constructor is given an empty state list
- `testConstructingSoftButtonObjectWithNonEmptyStateList()` passes if `SoftButtonObject.getStates()` returns the nonempty state list given to the constructor
- `testConstructSoftButtonObjectWithInvalidInitialStateName()` passes if `SoftButtonObject.getStates()` returns null when the constructor is given a state list that does not contain `initialStateName`
- `testAssignStateListWithCurrentStateToSoftButtonObject()` passes if `SoftButtonObject.getStates()` returns the new state list given to `SoftButtonObject.setStates()`
- `testAssignStateListWithoutCurrentStateToSoftButtonObject()` passes if `SoftButtonObject.getStates()` returns the new state list given to `SoftButtonObject.setStates()`
- `testAssignEmptyStateListToSoftButtonObject()` passes if `SoftButtonObject.getStates()` returns a state list equal to the nonempty state list passed to `SoftButtonObject.setStates()`

#### Core Tests

##### Tests performed by modifying a local clone of hello_sdl_android

- Constructed an instance of `SoftButtonObject` with a state list containing a single state, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
  - That instance along with all the other instances in the same transaction displayed successfully.
  - The app continued to function normally
- Constructed an instance of `SoftButtonObject` with an empty state list, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results: 
   - The instance printed an error with `DebugTool` saying that the state list is empty
   - That instance and all other other instances within the same transaction failed to display to the HMI
   - The app otherwise continued to function normally
-  Constructed an instance of `SoftButtonObject` with a state list containing a single state, reassigned the state list to an empty list using `setStates()`, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results:
   - The instance printed an error with `DebugTool` saying that the state list is empty
   - The instance did not overwrite the old state list with the new state list
   - The app otherwise continued to function normally
-  Constructed an instance of `SoftButtonObject` with a state list containing a single state, reassigned the state list to a state list with states with duplicate names using `setStates()`, added it to a list of `SoftButtonObject`s, and attempted to display them to the Generic HMI using `ScreenManager`. Results:
   - The instance printed an error with `DebugTool` saying "Two states have the same name in states list for soft button object"
   - The instance did not overwrite the old state list with the new state list
   - The app otherwise continued to function normally

SDL Core / 8.0.0 / master / 68f082169e0a40fccd9eb0db3c83911c28870f07 / module tested against: SDL Core on Ubuntu 20.04.3 LTS Virtual Machine
Generic HMI / 0.11.0 / master / 47e0ad42f05b27adff61befd864e79c2ab4b8cec / module tested against: Generic HMI on Ubuntu 20.04.3 LTS Virtual Machine

### Summary
This pull request adds an if statement to `SoftButtonObject.setStates()` along with an identical if statement in the main constructor that throws an error and returns if the state list passed is empty. This pull request also adds a similar check to `SoftButtonObject.setStates()` for state lists containing states with duplicate names.

Additionally `AssertionError`s are now thrown in these error checks when built in debug mode.

### Changelog
##### Breaking Changes
* None

##### Enhancements
* None

##### Bug Fixes
* Fixes #1774

### CLA

RHenigan (approved):

RHenigan:
Awaiting iOS Alignment before merging

base/src/main/java/com/smartdevicelink/managers/screen/SoftButtonObject.java
```diff
@@ -264,6 +275,30 @@ public void setName(@NonNull String name) {
...
+
+        boolean hasStateWithCurrentName = false;
+        for (SoftButtonState state : states) {
+            if(state.getName().equals(currentStateName)) {
+                hasStateWithCurrentName = true;
+                break;
+            }
+        }
+        if (!hasStateWithCurrentName) {
+            DebugTool.logError(TAG, "A SoftButtonObject should have a state with currentStateName.");
```
joeljfischer, line 299:
```suggestion
            DebugTool.logError(TAG, "A SoftButtonObject setting states must contain a state with currentStateName.");
```

Can you also include the name of `currentStateName`?

joeljfischer (changes requested):
This also seems to be missing the changes to `SoftButtonState` to ensure a state never has neither text nor an image.

noah-livio:
> This also seems to be missing the changes to `SoftButtonState` to ensure a state never has neither text nor an image.

That check is already handled in the constructor of `SoftButtonState`

joeljfischer (approved):

RHenigan (approved):

PR #1789, opened 2022-02-07
Prevent NPE in SdlDeviceListener

### LINKED ISSUE #1780: NPE crash in SdlDeviceListener
### Bug Report
Crash in `SdlDeviceListener.notifyConnection()` on line 262. Attempting to call `sdlListener.bluetoothTransport.stop()` when `bluetoothTransport` value is null.

##### Reproduction Steps
Unfortunately unable to reproduce at the moment. This crash came from our Crashlytics logs and has only been observed after moving from 5.2 to 5.3.

* Android Version: Seeing this in Android 9, 10, 11, and 12
* SDL Android Version: 5.3
* Testing Against: No tests as this came from Crashlytics logs.

None currently available.

##### Crash Log
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.smartdevicelink.transport.MultiplexBluetoothTransport.stop()' on a null object reference
       at com.smartdevicelink.transport.utl.SdlDeviceListener$TransportHandler.notifyConnection(SdlDeviceListener.java:262)
       at com.smartdevicelink.transport.utl.SdlDeviceListener$TransportHandler.onPacketRead(SdlDeviceListener.java:223)
       at com.smartdevicelink.transport.utl.SdlDeviceListener$TransportHandler.handleMessage(SdlDeviceListener.java:183)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:237)
       at android.app.ActivityThread.main(ActivityThread.java:8167)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)
```

RHenigan:
Fixes #1780

This PR is **[not ready]** for review.

### Risk

### Testing Plan
- [X] I have tested Android, Java SE, and Java EE

#### Unit Tests
N/A

#### Core Tests
[List of tests performed against Core and behaviors verified]

Core version / branch / commit hash / module tested against: [INSERT]
HMI name / version / branch / commit hash / module tested against: [INSERT]

### Summary
In `SdlDeviceListener` the `bluetoothTransport` and `bluetoothHandler` are only initialized if the device has not connected previously. In the event that the device has connected before the `bluetoothTransport` is not initialized because we get the relevant data from the shared preferences. If this is the case `bluetoothTransport` will not be initialized and will be null. If this is the case the `vehicleInfo` will be retrieved from the shared preferences instead and the logic to retrieve the vehicleData from the startServiceACK is not needed.

in `SdlDeviceListener` the `onPacketRead` method will be called when the StartServiceACK is received. This will later call `notifyConnection` which will try to stop the `bluetoothTransport` if the RFCOMM channel should not remain open. If this is the case after stoping the `bluetoothTransport` it will then be set to null. When this happens we should also set the `bluetoothHandler` to null as we no longer want its `handleMessages` method to be called. This will prevent the case where `notifyConnection` is called while the `bluetoothTransport` is null

In `onPacketRead` we are already performing a null check before writing to the `bluetoohTransport`. Since it is possible that `bluetoothTransport` is null when `notifyConnection` is called we should also check for null before calling `bluetoothTransport.stop()`.

### CLA

### DIFF
--- a/android/sdl_android/src/main/java/com/smartdevicelink/transport/utl/SdlDeviceListener.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/transport/utl/SdlDeviceListener.java
@@ -262,6 +262,7 @@ public class SdlDeviceListener {
                 if (sdlListener.bluetoothTransport != null) {
                     sdlListener.bluetoothTransport.stop();
                     sdlListener.bluetoothTransport = null;
+                    sdlListener.bluetoothHandler = null;
                 }
                 sdlListener.timeoutHandler.removeCallbacks(sdlListener.timeoutRunner);
             }
--- a/android/sdl_android/src/main/java/com/smartdevicelink/transport/utl/SdlDeviceListener.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/transport/utl/SdlDeviceListener.java
@@ -259,8 +259,10 @@ public class SdlDeviceListener {
             AndroidTools.saveVehicleType(sdlListener.contextWeakReference.get(), vehicleType, sdlListener.connectedDevice.getAddress());
             boolean keepConnectionOpen = sdlListener.callback.onTransportConnected(sdlListener.contextWeakReference.get(), sdlListener.connectedDevice);
             if (!keepConnectionOpen) {
-                sdlListener.bluetoothTransport.stop();
-                sdlListener.bluetoothTransport = null;
+                if (sdlListener.bluetoothTransport != null) {
+                    sdlListener.bluetoothTransport.stop();
+                    sdlListener.bluetoothTransport = null;
+                }
                 sdlListener.timeoutHandler.removeCallbacks(sdlListener.timeoutRunner);
             }
         }


noah-livio:
Fixes #1780

### Risk

### Testing Plan
- [] I have run the unit tests with this PR - N/A
- [X] I have tested Android, Java SE, and Java EE

#### Unit Tests
N/A

#### Core Tests
Using a modified version of the SDL library without Robert's changes, I was able to force the NPE by setting `bluetoothTransport` to null before `notifyConnection()` is called.

After including Robert's changes, the NPE no longer occurs and the crash is prevented.

Core version / branch / commit hash / module tested against: SDL Core v8.0.0 master 68f082169e0a40fccd9eb0db3c83911c28870f07
HMI name / version / branch / commit hash / module tested against: Generic HMI v0.11.0 master 47e0ad42f05b27adff61befd864e79c2ab4b8cec


### Summary
This PR surrounds `sdlListener.bluetoothTransport.stop()` in `SdlDeviceListener.notifyConnection()` with a null check and sets `sdlListener.bluetoothHandler` to null in the last line of the null checked block. 

I have verified Robert's assessment regarding possible causes of `bluetoothTransport`  entering a null state and have also verified his changes are sufficient to prevent the reported NPE.

### CLA

JulianKast (approved):

PR #1840, opened 2023-02-02
Hot Fix for  #1839. Adding fallback logic for starting router service

### LINKED ISSUE #1839: Nearby Devices permission on Android 12 and Above
### Bug Summary

This issue is related to #1794. The app connection and registration fails when multiple apps are installed and a single app does not have permission for nearby devices on Android 12 and above. The scenario in the reproduction steps can be replicated by using a combination of iHeartRadio from the production Play Store. and HelloSDL built from the latest master branch in the java suite.

### Reproduction Steps

1. Install iHeartRadio from the production Play Store on an Android 12 or above device.
2. Build and install HelloSDL on the same device from the master branch of the java suite github repository.
3. Provide Nearby device permission to the HelloSDL app in the app settings or in the app itself when prompted.
4. Connect the device over bluetooth to a Sync TDK.
5. Take note that no apps are registered on the HMI.
6. On the Android device, go to the App Permissions for iHeartRadio and provide permission for Nearby devices to iHeartRadio.
7. Disconnect and reconnect bluetooth on the Android device to cycle the bluetooth connection.
8. Take note that both iHeartRadio and HelloSDL are registered on the HMI.
9. Go back into the iHeartRadio settings and remove permission for Nearby devices. 
10. Take note that both HelloSDL and iHeartRadio are removed from the HMI

Expected Behavior:
Nearby device permission settings for one app do not impact the settings for another app. In the case above, the Nearby device permission from iHearthRadio should not impact the registration of HelloSDL

Observed Behavior:
The Nearby devices app permission settings for a single app does impact other apps (in the case above HelloSDL cannot register due to iHeartRadio missing permission for Nearby devices).

### Which projects have you seen this bug on?

Android

### Android Version(s)

12

### Android Device(s)

~~Essential~~ Nokia 8.3

### sdl_java_suite Version

5.6.0

### Testing Environment(s)

```markdown
Ford Sync TDK
```

### Relevant log output

_No response_

joeygrover:
Fixes #1839

### Risk

### Testing Plan
- [x] I have tested Android, ~Java SE, and Java EE~

#### Unit Tests
None added for this PR as the changes are not unit testable.

#### Core Tests

Tested with a Gen3 Ford TDK
Cycled through testing different permission allocations and who would start and ultimately host the router service.
image

### Summary
Essentially added a new logical assumption that all apps need to be targeting Android 12 at this point and therefore require the user granted Nearby Devices permission to receive Bluetooth connection intents. Therefore, if an older version of the library is installed that does not have the newer permission check logic prior to start the RS, if that app does not have the correct permission we can assume they will not be receiving the BT ACL connection intent and we can ignore the old legacy logic that it would have executed. 

We also added a fallback call to start the newest router service in the case there was an older SDL enabled app on the device but their integration was incorrect. 

##### Hot Fix
* Adds better recoverability to the RS starting process when legacy SDL apps are installed on the device.

### Tasks Remaining:
- [ ] Update the versions before officially releasing as a hot fix

### CLA


mrapitis:
>Update the versions before officially releasing as a hot fix

Does this include an increment of the sdl_router_service_version_value value? I noticed the RS version of 17 in the graphic, but still 16 in the sdl.xml file (I can manually adjust to 17 for testing if needed).

joeygrover:
> > Update the versions before officially releasing as a hot fix
> 
> Does this include an increment of the sdl_router_service_version_value value? I noticed the RS version of 17 in the graphic, but still 16 in the sdl.xml file (I can manually adjust to 17 for testing if needed).

RS version will remain 16. 

So yes, 17 was for testing only and was manually adjusted. Referenced comment is to update the version in the gradle file before merging as a hotfix.

mrapitis (approved):

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlBroadcastReceiver.java
```diff
@@ -333,6 +341,16 @@ public void onComplete(Vector<ComponentName> routerServices) {
                                 } else {
                                     DebugTool.logInfo(TAG, "Not starting device listener, bluetooth device is null and other SDL apps installed.");
                                 }
+                            } else if(isPreAndroid12RSOnDevice){
```
JulianKast, line 344:
```suggestion
                            } else if(isPreAndroid12RSOnDevice) {
```

JulianKast (changes requested):
One formatting error, otherwise looks good, Do you want to update the version as well for the release?

PR #1860, opened 2023-08-11
Android 14 foreground service type required

### LINKED ISSUE #1859: Android 14 foreground service type required
### Description

Android 14 requires us to set a foreground service type of `connectedDevice`. 
With that, we must have the `BLUETOOTH_CONNECT` permission or call `UsbManager.requestPermission()` to allow our services to enter the foreground.

JulianKast:
Fixes #1859 

### Risk

### Testing Plan
- [x] I have tested Android

#### Unit Tests
Unit tests were run on the integration branch. CI check failed due to needing the app to target API 34.

#### Core Tests
Note: Before you install a second and third app, unplug the phone from the computer and switch the build variant before deploying the app to the phone otherwise Android Studio will kill the app.

Apps A, B and C.
Test 1:
Install App A
Accept BT permission
Install App B
Deny BT permission
Connect to TDK via AOA
Select App A to receive intent
Allow App B to connect when prompted
Observe: App B pass hosting the RS to app A and both apps connect.

Test 2:
Install App A
Deny BT permission
Install App B
Deny BT permission
Connect to TDK via AOA
Select App A to receive intent
Allow App B to connect when prompted
Observe App B passed hosting the RS to app A and both apps connect.

Test 3:
Install App A
Deny BT permission
Install App B
Deny BT permission
Install App C 
Deny BT permission
Connect to TDK via AOA
Select App A to receive intent
Allow App C to connect when prompted
Allow App B to connect if you're prompted.(not 100%)
Observe App C passed hosting the RS to App B who passed to App A and Apps A and C are connected. App B may not connect.

Core version / branch / commit hash / module tested against: Sync 3
HMI name / version / branch / commit hash / module tested against: Sync 3

### Summary
This pr adds the foreground service type of connectedDevice to our library. This is a requirement for Android 14.
With this addition, we are required to have either the Bluetooth Connect permission or call UsbManager.requestPermission() to have our service enter the foreground. if the connection is AOA and an app selected to host the RouterService does not have either permission, we will try to pass hosting the RouterService off to another app.

I also add the ability to be able to request the usb permission in SdlReceiver.onSdlEnabled in the event that 

##### Bug Fixes
* Fixed the library to allow for it to target Android 14 and be able to enter the foreground

### CLA

### DIFF
--- a/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
+++ b/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
@@ -102,7 +102,10 @@ public class SdlReceiver extends SdlBroadcastReceiver {
      */
     private void requestUsbAccessory(Context context) {
         UsbManager manager = (UsbManager) context.getSystemService(Context.USB_SERVICE);
-        if (manager.getAccessoryList() == null) {
+        UsbAccessory[] accessoryList = manager.getAccessoryList();
+        if (accessoryList == null || accessoryList.length == 0) {
+            startSdlServiceIntent = null;
+            pendingIntentToStartService = null;
             return;
         }
         PendingIntent mPermissionIntent = PendingIntent.getBroadcast(context, 0, new Intent(ACTION_USB_PERMISSION), PendingIntent.FLAG_IMMUTABLE);
--- a/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
+++ b/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
@@ -15,7 +15,6 @@ import com.smartdevicelink.transport.TransportConstants;
 import com.smartdevicelink.util.AndroidTools;
 import com.smartdevicelink.util.DebugTool;
 
-
 public class SdlReceiver extends SdlBroadcastReceiver {
     private static final String TAG = "SdlBroadcastReceiver";
 
--- a/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
+++ b/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
@@ -35,7 +35,7 @@ public class SdlReceiver extends SdlBroadcastReceiver {
             PendingIntent pendingIntent = (PendingIntent) intent.getParcelableExtra(TransportConstants.PENDING_INTENT_EXTRA);
             if (pendingIntent != null) {
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
-                    if (!AndroidTools.ServicePermissionUtil.hasForegroundServiceTypePermission(context)) {
+                    if (!AndroidTools.hasForegroundServiceTypePermission(context)) {
                         requestUsbAccessory(context);
                         storedIntent = intent;
                         storedContext = context;
@@ -82,7 +82,7 @@ public class SdlReceiver extends SdlBroadcastReceiver {
         public void onReceive(Context context, Intent intent) {
             String action = intent.getAction();
             if (ACTION_USB_PERMISSION.equals(action) && storedContext != null && storedIntent != null && pendingIntent != null) {
-                if (AndroidTools.ServicePermissionUtil.hasForegroundServiceTypePermission(storedContext)) {
+                if (AndroidTools.hasForegroundServiceTypePermission(storedContext)) {
                     try {
                         pendingIntent.send(storedContext, 0, storedIntent);
                     } catch (PendingIntent.CanceledException e) {
--- a/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
+++ b/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
@@ -44,8 +44,7 @@ public class SdlReceiver extends SdlBroadcastReceiver {
                 }
                 try {
                     pendingIntent.send(context, 0, intent);
-                    context.unregisterReceiver(this);
-                } catch (Exception e) {
+                } catch (PendingIntent.CanceledException e) {
                     e.printStackTrace();
                 }
             }
@@ -83,7 +82,8 @@ public class SdlReceiver extends SdlBroadcastReceiver {
                 if (AndroidTools.hasForegroundServiceTypePermission(context)) {
                     try {
                         pendingIntentToStartService.send(context, 0, startSdlServiceIntent);
-                    } catch (PendingIntent.CanceledException e) {
+                        context.unregisterReceiver(this);
+                    } catch (Exception e) {
                         e.printStackTrace();
                     }
                 }
--- a/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
+++ b/android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
@@ -1,17 +1,27 @@
 package com.sdl.hellosdlandroid;
 
 import android.app.PendingIntent;
+import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
+import android.content.IntentFilter;
+import android.hardware.usb.UsbAccessory;
+import android.hardware.usb.UsbManager;
 import android.os.Build;
 
 import com.smartdevicelink.transport.SdlBroadcastReceiver;
 import com.smartdevicelink.transport.SdlRouterService;
 import com.smartdevicelink.transport.TransportConstants;
+import com.smartdevicelink.util.AndroidTools;
 import com.smartdevicelink.util.DebugTool;
 
 public class SdlReceiver extends SdlBroadcastReceiver {
     private static final String TAG = "SdlBroadcastReceiver";
+
+    private static final String ACTION_USB_PERMISSION = "com.android.example.USB_PERMISSION";
+    private PendingIntent pendingIntent;
+    private Context storedContext;
+    private Intent storedIntent;
 
     @Override
     public void onSdlEnabled(Context context, Intent intent) {
@@ -22,8 +32,18 @@ public class SdlReceiver extends SdlBroadcastReceiver {
         // We will check the intent for a pendingIntent parcelable extra
         // This pendingIntent allows us to start the SdlService from the context of the active router service which is in the foreground
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            storedIntent = intent;
             PendingIntent pendingIntent = (PendingIntent) intent.getParcelableExtra(TransportConstants.PENDING_INTENT_EXTRA);
             if (pendingIntent != null) {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+                    if (!AndroidTools.ServicePermissionUtil.hasForegroundServiceTypePermission(context)) {
+                        requestUsbAccessory(context);
+                        storedContext = context;
+                        this.pendingIntent = pendingIntent;
+                        DebugTool.logInfo(TAG, "Permission missing for ForegroundServiceType connected device." + context);
+                        return;
+                    }
+                }
                 try {
                     pendingIntent.send(context, 0, intent);
                 } catch (PendingIntent.CanceledException e) {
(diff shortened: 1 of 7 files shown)


JulianKast:
Unit test for Android currently fail due to needing to target API 34 which is in a different PR.

joeygrover (changes requested):
Looks like the other PRs need to be finished and merged into develop, then develop back into this branch to test more properly. During the first review it seems like everything is working as it should, but will need to retest and more robustly post the rest of the fixes being merged.

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -1,18 +1,28 @@
...
 import com.smartdevicelink.transport.TransportConstants;
+import com.smartdevicelink.util.AndroidTools;
 import com.smartdevicelink.util.DebugTool;
 
 public class SdlReceiver extends SdlBroadcastReceiver {
     private static final String TAG = "SdlBroadcastReceiver";
 
+    private static final String ACTION_USB_PERMISSION = "com.android.example.USB_PERMISSION";
+    private PendingIntent pendingIntent;
+    private Context storedContext;
```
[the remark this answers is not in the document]
JulianKast: Yes it appears that we can use the context from `mUsbReceiver`

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -56,4 +76,43 @@ public void onReceive(Context context, Intent intent) {
...
+
+        public void onReceive(Context context, Intent intent) {
+            String action = intent.getAction();
+            if (ACTION_USB_PERMISSION.equals(action) && context != null && cachedIntent != null && pendingIntent != null) {
+                if (AndroidTools.hasForegroundServiceTypePermission(context)) {
+                    try {
+                        pendingIntent.send(context, 0, cachedIntent);
+                    } catch (PendingIntent.CanceledException e) {
+                        e.printStackTrace();
+                    }
```
joeygrover, line 90: There's currently no call to unregister this receiver so it will potentially leak. Changed the caught exception to a generic one as well in case something happens during unregistering.

```suggestion
                    try {
                        pendingIntent.send(context, 0, cachedIntent);
                        context.unregisterReceiver(this);
                    } catch (Exception e) {
                        e.printStackTrace();
                    }
```

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -56,4 +76,43 @@ public void onReceive(Context context, Intent intent) {
...
+     * we can request the USB Accessory permission to satisfy the requirements for
+     * FOREGROUND_SERVICE_CONNECTED_DEVICE and can start our service and allow
+     * it to enter the foreground. FOREGROUND_SERVICE_CONNECTED_DEVICE is a requirement
+     * in Android 14
+     */
+    private void requestUsbAccessory(Context context) {
+        UsbManager manager = (UsbManager) context.getSystemService(Context.USB_SERVICE);
+        if (manager.getAccessoryList() == null) {
+            return;
+        }
```
joeygrover, line 108: First a reference to the list can be created here since it is used a few times in the logic below. Then in the case where there is no available USB device we should clear out the intents that we stored.

```suggestion
        UsbAccessory[] accessoryList = manager.getAccessoryList();
        if (accessoryList == null || accessoryList.length == 0) {
            cachedIntent = null;
            pendingIntent = null;
            return;
        }
```

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -1,18 +1,29 @@
...
 import com.smartdevicelink.util.DebugTool;
 
+import java.lang.ref.WeakReference;
+
 public class SdlReceiver extends SdlBroadcastReceiver {
     private static final String TAG = "SdlBroadcastReceiver";
 
+    private static final String ACTION_USB_PERMISSION = "com.android.example.USB_PERMISSION";
+    private PendingIntent pendingIntent;
+    private Intent cachedIntent;
```
joeygrover, line 25: These should probably be a little more specific in terms of what they are without having to read into the code more to find out what they do. These are verbose names though so maybe you can think of something better. 

```suggestion
    private PendingIntent pendingIntentToStartService;
    private Intent startSdlServiceIntent;
```

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -56,4 +76,43 @@ public void onReceive(Context context, Intent intent) {
     public String getSdlServiceName() {
         return SdlService.class.getSimpleName();
     }
+
+    private final BroadcastReceiver mUsbReceiver = new BroadcastReceiver() {
```
joeygrover, line 80: Name should be more specific, and avoid the `m` member naming convention if possible.
```suggestion
    private final BroadcastReceiver usbPermissionReceiver = new BroadcastReceiver() {
```

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -56,4 +76,43 @@ public void onReceive(Context context, Intent intent) {
...
+    private void requestUsbAccessory(Context context) {
+        UsbManager manager = (UsbManager) context.getSystemService(Context.USB_SERVICE);
+        if (manager.getAccessoryList() == null) {
+            return;
+        }
+        PendingIntent mPermissionIntent = PendingIntent.getBroadcast(context, 0, new Intent(ACTION_USB_PERMISSION), PendingIntent.FLAG_IMMUTABLE);
+        IntentFilter filter = new IntentFilter(ACTION_USB_PERMISSION);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            context.registerReceiver(mUsbReceiver, filter, Context.RECEIVER_EXPORTED);
+        }
```
joeygrover, line 113: The new method that was added to the utility class can be used here. 

```suggestion
       AndroidTools.registerReceiver(context, mUsbReceiver, filter, Context.RECEIVER_EXPORTED);
```

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -56,4 +76,43 @@ public void onReceive(Context context, Intent intent) {
...
+        for (final UsbAccessory usbAccessory : manager.getAccessoryList()) {
```
joeygrover, line 114: Best to avoid calling a getter method for each iteration through the for loop. See previous comment for where the reference should be created. 

```suggestion
        for (final UsbAccessory usbAccessory : accessoryList) {
```

android/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
```diff
@@ -417,4 +421,54 @@ public static void registerReceiver(Context context, BroadcastReceiver receiver,
...
+     * @param applicationContext Context used to check permission
+     * @param permission String representing permission that is being checked.
+     * @return true if app has permission.
+     */
+    public static boolean checkPermission(Context applicationContext, String permission) {
+        if (applicationContext == null) {
+            return false;
+        }
+        return PackageManager.PERMISSION_GRANTED == ContextCompat.checkSelfPermission(applicationContext, permission);
+    }
```
joeygrover, line 459: Unless this needs to specifically be the context for the application it should be renamed to just `context`. If it needs to be an application context then any method that calls this one, eg `hasForegroundServiceTypePermission`, would need to make that a requirement. 

```suggestion
    public static boolean checkPermission(Context context, String permission) {
        if (context == null) {
            return false;
        }
        return PackageManager.PERMISSION_GRANTED == ContextCompat.checkSelfPermission(context, permission);
    }
```

android/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
```diff
@@ -417,4 +421,54 @@ public static void registerReceiver(Context context, BroadcastReceiver receiver,
...
+            if (manager.hasPermission(usbAccessory)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Helper method used to check permission passed in.
+     * @param applicationContext Context used to check permission
```
joeygrover, line 450:
```suggestion
     * @param context Context used to check permission
```

android/sdl_android/src/main/java/com/smartdevicelink/util/AndroidTools.java
```diff
@@ -417,4 +421,54 @@ public static void registerReceiver(Context context, BroadcastReceiver receiver,
...
+        }
+        return PackageManager.PERMISSION_GRANTED == ContextCompat.checkSelfPermission(applicationContext, permission);
+    }
+
+    /**
+     * A helper method used for Android 14 or greater to check if app has necessary permissions
+     * to enter the foreground.
+     * @param context context used to check permissions.
+     * @return true if app has permission to enter foreground or if Android version < 14
+     */
```
joeygrover, line 466:
```suggestion
     * A helper method used for Android 14 or greater to check if app has necessary permissions
     * to have a service enter the foreground.
     * @param context context used to check permissions.
     * @return true if app has permission to have a service enter foreground or if Android version < 14
     */
```

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -1,18 +1,29 @@
...
 import android.os.Build;
 
 import com.smartdevicelink.transport.SdlBroadcastReceiver;
 import com.smartdevicelink.transport.SdlRouterService;
 import com.smartdevicelink.transport.TransportConstants;
+import com.smartdevicelink.util.AndroidTools;
 import com.smartdevicelink.util.DebugTool;
 
+import java.lang.ref.WeakReference;
+
```
joeygrover, line 19: Unused import.

```suggestion
```

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -862,6 +862,14 @@ public void handleMessage(Message msg) {
                     ParcelFileDescriptor parcelFileDescriptor = (ParcelFileDescriptor) msg.obj;
```
joeygrover, line 862: Not related to this line, but this class. Since this new code is using the alt transport timer I think it's important that it gets updated a bit. I think we should just use the main looper instead of trying to call a different one. So the method should be updated as follows:

```java
    private void startAltTransportTimer() {
        if (altTransportTimerHandler != null && altTransportTimerRunnable != null) {
            altTransportTimerHandler.removeCallbacks(altTransportTimerRunnable);
        }

        altTransportTimerHandler = new Handler(Looper.getMainLooper());
        altTransportTimerRunnable = new Runnable() {
            public void run() {
                altTransportTimerHandler = null;
                altTransportTimerRunnable = null;
                if (!shouldServiceRemainOpen(null)) {
                    closeSelf();
                }
            }
        };
        altTransportTimerHandler.postDelayed(altTransportTimerRunnable, ALT_TRANSPORT_TIMEOUT_RUNNABLE);
    }
```

joeygrover (changes requested):

joeygrover (changes requested):
Looks like one last missed change request from the previous review

android/hello_sdl_android/src/main/java/com/sdl/hellosdlandroid/SdlReceiver.java
```diff
@@ -56,4 +75,46 @@ public void onReceive(Context context, Intent intent) {
...
+        public void onReceive(Context context, Intent intent) {
+            String action = intent.getAction();
+            if (ACTION_USB_PERMISSION.equals(action) && context != null && startSdlServiceIntent != null && pendingIntentToStartService != null) {
+                if (AndroidTools.hasForegroundServiceTypePermission(context)) {
+                    try {
+                        pendingIntentToStartService.send(context, 0, startSdlServiceIntent);
+                    } catch (PendingIntent.CanceledException e) {
+                        e.printStackTrace();
+                    }
+                }
```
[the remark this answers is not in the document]
JulianKast: ah, looks like I made the this change in the wrong spot

PR #1862, opened 2023-08-15
Android 14 add package to PendingIntents in RouterService

### LINKED ISSUE #1861: Android 14 Pending Intent must have a package
### Description

In Android 14 we are now required to add a package to pending intents.

 ["If an app creates a mutable pending intent with an intent that doesn't specify a component or package, the system now throws an exception."](https://developer.android.com/about/versions/14/behavior-changes-14#safer-intentsurl)

JulianKast:
Fixes #1861 

### Risk

### Testing Plan
- [x] I have tested Android

#### Unit Tests
n/a

#### Core Tests
Note: Before you install a second and thrid app, unplug the phone from the computer and switch the build variant before deploying app to the phone otherwise Android Studio will kill the app. 
Or you can open each app up on the phone after deploying all three apps.

Test 1
Install 3 apps to the phone
Connect via Bluetooth to TDK
Observe all apps connect.

Test 2
Repeat test 1 but with release build versions of the apps.

Core version / branch / commit hash / module tested against: Sync 3
HMI name / version / branch / commit hash / module tested against: Sync 3

### Summary
This PR adds a package to pending intents in the RouterService to prevent an exception from being thrown in the RouterService.

##### Bug Fixes
* Add Package to pending intent to prevent an exception from being thrown due to API 34 restrictions. 

### CLA

### DIFF
--- a/android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
+++ b/android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
@@ -158,9 +158,7 @@ public class SdlRouterService extends Service {
     public static final String SDL_NOTIFICATION_FAQS_PAGE = "https://smartdevicelink.com/en/guides/android/frequently-asked-questions/sdl-notifications/";
 
     public static final String REGISTER_WITH_ROUTER_ACTION = "com.sdl.android.register";
-
-    public static final String SDL_PACKAGE = "com.smartdevicelink";
-
+    
     /**
      * Message types sent from the BluetoothReadService Handler
      */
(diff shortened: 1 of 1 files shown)


android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -1858,6 +1860,7 @@ public void onTransportConnected(final TransportRecord record) {
             //the developer can use this pendingIntent to start their SdlService from the context of
             //the active RouterService
             Intent pending = new Intent();
+            pending.setPackage(SDL_PACKAGE);
```
joeygrover, line 1863: I think it is better to use the actual package of the currently running service than one that is generic one.

```suggestion
            pending.setPackage(getPackageName());
```

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -2981,6 +2984,7 @@ private Intent createPingIntent() {
...
```
joeygrover, line 2987: See previous comment.

```suggestion
            pending.setPackage(getPackageName());
```

android/sdl_android/src/main/java/com/smartdevicelink/transport/SdlRouterService.java
```diff
@@ -159,6 +159,8 @@ public class SdlRouterService extends Service {
 
     public static final String REGISTER_WITH_ROUTER_ACTION = "com.sdl.android.register";
 
+    public static final String SDL_PACKAGE = "com.smartdevicelink";
+
```
joeygrover, line 163: See previous comment. 

```suggestion
```

joeygrover (changes requested):

joeygrover (approved):

